PR #5, opened 2014-07-03
[IMP] account_statement_base_completion: Overridable method for get completion rules.

pedrobaeza:


account_statement_base_completion/statement.py
```diff
@@ -139,6 +139,10 @@ def _get_functions(self, cr, uid, context=None):
             ('get_from_label_and_partner_field', 'From line label (based on partner field)'),
             ('get_from_label_and_partner_name', 'From line label (based on partner name)')]
 
+    def __get_functions(self, cr, uid, context=None):
+        """Internal method. Do not override."""
```
bwrsandman, line 143: Is this how we officially do private functions in Odoo?

bwrsandman:
:+1:

account_statement_base_completion/statement.py
pedrobaeza, line 143: Well, this is a descriptive warning, but you know that python makes a "semi-protection" of these methods (prepending the class name to the method).

PR #10, opened 2014-07-13
[IMP] Add statement line reference to search

RoelAdriaans:
Add the reference to the search for matching statements. When importing CAMT (Dutch banking files) the description of the statement line is filled in the reference field. The name field only has the index (eg 001), not the full text.

### DIFF
--- a/account_statement_completion_label/statement.py
+++ b/account_statement_completion_label/statement.py
@@ -92,7 +92,7 @@ class AccountStatementCompletionRule(orm.Model):
                          ON
                             st_l.statement_id = s.id
                     WHERE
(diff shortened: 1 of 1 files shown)


pedrobaeza:
LGTM

:+1:

sbidoul:
I have doubts about this one. Indeed the name field in a statement line is the OBI (Originator to Beneficiary Information) and the ref field should contain the line reference.

Don't you think the CAMT parser should be adapted instead?

max3903:
:+1:

foutoucour:
:+1:

PR #12, opened 2014-07-15
[imp] auto reconcile: commit every 10 lines

lepistone:


account_advanced_reconcile/base_advanced_reconciliation.py
```diff
@@ -269,4 +282,7 @@ def _rec_auto_lines_advanced(self, cr, uid, rec, credit_lines, debit_lines, cont
             elif reconciled:
                 partial_reconciled_ids += reconcile_group_ids
 
+            if COMMIT_EVERY and group_count % COMMIT_EVERY == COMMIT_EVERY - 1:
```
guewen: I wonder if:

```
(group_count + 1) % COMMIT_EVERY == 0:
```

would not be more simple to grasp.

guewen:
Yes! :cool: 
Just a diff comment on the modulo.

lepistone:
@guewen refactored as you suggest

guewen:
:+1:

lmignon:
Hi,

Thank you for the contrib...

The code looks good but I wonder what problem you are trying to solve? Why make a commit every 10 entries and not every 20 or??

Regards,

lmi

lepistone:
Hi @lmignon,
an account with 50K unreconciled lines. This avoids reconciling a single transaction running all day that will eventually lock the table and cause serialisation errors. In case of problems, all work is not lost.

I didn't feel like 5 or 10 or 50 made a big difference so it is a constant. It could be made a configuration parameter later.

Another option I'll consider only if really necessary is to use the @guewen 's [connector](https://github.com/OCA/connector) like I did in the module [account_move_batch_validate](https://github.com/OCA/account-financial-tools/tree/7.0/account_move_batch_validate).

sebastienbeau:
:+1: 
I totaly approve the change. I will avoid. to run and run again and again the reconciliation starting from the start.
Good job

lmignon:
Hi Leonardo,

I understand the problem, but I'm not comfortable with the solution since it breaks the main transaction.   This kind of code also breaks tests based on TransactionCase.  It would be safer to add a new column (your Model inherit from easy.reconcile.options....)  to enable or not this feature and to provide the batch size. 

Regars,

lmi

lepistone:
I see your point @lmignon.

Since this has just been merged, I'll make later another PR to add the configuration parameter, disabled by default.

thanks

lmignon:
too bad .. It would be safe to have a delay between the last approval and the merge to allow time to answer

bwrsandman:
Would you like me to revert the merge?

lepistone:
@bwrsandman since i'm preparing another PR now to handle @lmignon's remarks, I suggest we do not.

Thanks!

lmignon:
@bwrsandman no, since a new PR in is preparation :smiley:

guewen:
Having a batch size on the profiles seems overkill to me. Would you configure it differently per profile?
It seems to me that a global option would be sufficient.

A remark on the usage of a new transaction: this strategy is used in the official addons as well, for instance in the procurements scheduler. This is the proper way to commit chunks of work on a long-running process. I don't deny the benefits to be able to switch off the usage of a new cursor though, as it would allow to run tests in the TransactionCase.

lepistone:
+1 @guewen: i'll make that a global setting, I don't see any reason (only many possible mistakes) in configuring that per profile.

lmignon:
+1 you're right for the batch size, my main concern is to provide a way to switch off the usage of a new cursor.

PR #13, opened 2014-07-16
7.0 Add meta files

bwrsandman:


bwrsandman:
Failing due to pep8 and absence of account_default_draft_move for account_statement_cancel_line

bwrsandman:
Failing due to dependency: `ofxparse`
I don't have time to investigate, but if someone would like to let me know where and how to get it, either post it here of do an MP to my branch.

pedrobaeza:
Hi, Sandy, you can install it with:

`sudo pip install ofxparse`

yvaucher:
TypeError: argument must be 9-item sequence, not datetime.datetime

This can be fixed in another PR

Otherwise :+1:

PR #15, opened 2014-07-16
allow to configure reconcile_commit_every

lepistone:
this allows to configure if we want to commit the automatic reconcile every
N lines. It is left to zero by default to avoid possible problems with
tests that rely on a transaction.

### DIFF
--- a/account_advanced_reconcile/__init__.py
+++ b/account_advanced_reconcile/__init__.py
@@ -1,7 +1,8 @@
 # -*- coding: utf-8 -*-
 ##############################################################################
 #
-#    Author: Guewen Baconnier, Leonardo Pistone
+#    Author: Guewen Baconnier
+#    Contributor: Leonardo Pistone
 #    Copyright 2012-2014 Camptocamp SA
 #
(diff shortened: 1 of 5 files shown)


account_advanced_reconcile/__init__.py
```diff
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 ##############################################################################
 #
-#    Author: Guewen Baconnier
-#    Copyright 2012 Camptocamp SA
+#    Author: Guewen Baconnier, Leonardo Pistone
```
pedrobaeza: I think you must put as contributor, not author.

lepistone:
This continues the discussion started on https://github.com/OCA/bank-statement-reconcile/pull/12 with @guewen @lmignon @sebastienbeau @bwrsandman.

Thanks!

lepistone:
@pedrobaeza I changed according to your remark. Thanks!

guewen:
:+1:

lmignon:
:+1:  
Thank you for this improvement to your initial PR!

bwrsandman:
:+1:

PR #20, opened 2014-07-23
[MIG] <account_statement_ext> migrated

oihane:


bwrsandman:
Hi, can you re-open your PR against 8.0, right now it's against master which relates to 9.0, and should be removed eventually.

PR #25, opened 2014-08-01
[FIX] travis: update travis file to match the one in https://github.com/...

lmignon:
...OCA/maintainer-quality-tools/blob/master/sample_files/.travis.yml

Tests already fails but everything is now properly installed and configured.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,17 +6,20 @@ python:
 virtualenv:
   system_site_packages: true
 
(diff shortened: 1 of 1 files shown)


bwrsandman:
:+1:

pedrobaeza:
:+1: 

We have to fix afterwards PEP8 and an error in one module.

Regards.

PR #29, opened 2014-08-04
[IMP] account_statement_base_import: Fill balance start when parser doesn't filled it

pedrobaeza:
Get starting balance from journal balance if parser doesn't fill this data, simulating the manual flow

### DIFF
--- a/__unported__/account_statement_base_import/statement.py
+++ b/__unported__/account_statement_base_import/statement.py
@@ -127,12 +127,21 @@ class AccountStatementProfil(Model):
         values = statement_line_obj._add_missing_default_values(cr, uid, values, context)
         return values
 
-    def prepare_statement_vals(self, cr, uid, profile_id, result_row_list, parser, context):
+    def prepare_statement_vals(self, cr, uid, profile_id, result_row_list,
+                               parser, context=None):
         """
         Hook to build the values of the statement from the parser and
         the profile.
         """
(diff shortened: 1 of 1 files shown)


guewen:
:+1:

gurneyalex:
:+1:

__unported__/account_statement_base_import/statement.py
```diff
@@ -127,13 +127,22 @@ def prepare_statement_lines_vals(
...
-    def prepare_statement_vals(self, cr, uid, profile_id, result_row_list, parser, context):
+    def prepare_statement_vals(self, cr, uid, profile_id, result_row_list,
+                               parser, context=None):
         """
         Hook to build the values of the statement from the parser and
         the profile.
         """
         vals = {'profile_id': profile_id}
         vals.update(parser.get_st_vals())
+        if not vals.get('balance_start'):
```
sbidoul, line 138: Is it better to use if vals.get('balance_start') is None? To distinguish cases were the parser returns a 0 balance from cases were the parser does not return balance_start.

sbidoul:
Oh, I notice this one is already merged in 7.0 and propagated to 8.0. So this PR should be closed?

pedrobaeza:
Yeah, I also didn't notice that. BTW, I prefer this method because this balance is filled by default with 0.0, so the method `if ... is None` is not valid.

PR #34, opened 2014-08-07
[FIX] port account_statement_ext to V8.0

Cedric-Pigeon:
Migration of the module to V8. 

---

The main modifications are due to the deletion of the field type of statement line. It was used to defined which account to use.

Now odoo bases its decision on the amount. If amount is positive, a receivable account is proposed else a payable account is returned.


pedrobaeza:
Well, I see this a huge problem, because you can have refunds that change normal sign. I think we must implement this field in this module, and continue using it regardless of Odoo decission.

Cedric-Pigeon:
You are right but we discussed internally about this topic with Stéphane Bidoul and Laurent Mignon.
Our feeling is that we don't need to re-introduce this field and neither to fix the account.

The new reconciliation process of Odoo V8 is quiet more efficient and include algorithms which are more able to suggest the right accounts.
The account_id on statement line is from now only present for POS module which must always give the account to use. This is why the field is now optional. 

What do you think ?

sbidoul:
This one should be set "work in progress" as per this thread:
https://lists.launchpad.net/banking-addons-drivers/msg00096.html

JehoG:
Is there some evolution on this?

pedrobaeza:
This has been deprecated due to new bank import framework, so it doesn't apply and can be closed.

JehoG:
And the functionality of this are somewhere in V8? 
Are they in the new bank import framework or they are not there anymore and should be recoded?

pedrobaeza:
Most of the functionality is in the new framework, and some missing features are being handled here: https://github.com/OCA/bank-statement-import/pull/4

yvaucher:
@pedrobaeza If it's deprecated, can you open an issue or a PR to remove `account_statement_ext` from this repository

pedrobaeza:
Done

PR #35, opened 2014-08-11
Corrects issue with customer refunds and partner lookup for imported lines.

mdietrichc2c:
If you try to auto-complete a customer refund, the partner retrieve will be the partner_id, not the commercial_partner_id like in customer invoices. This fix is to homogenize the behaviour between invoices and refunds.

### DIFF
--- a/account_statement_transactionid_completion/statement.py
+++ b/account_statement_transactionid_completion/statement.py
@@ -110,16 +110,7 @@ class AccountStatementCompletionRule(Model):
         elif len(invoice_id) == 1:
             invoice = invoice_obj.browse(cr, uid, invoice_id[0],
                                          context=context)
-            # if account_report_company is installed, the field to be
-            # used as partner is the commercial partner; otherwise,
-            # the basic partner must be used.
-            # 
-            # This is used to keep consistent behavior from the module
-            # account_statement_base_completion.
-            if hasattr(invoice, 'commercial_partner_id'):
-                res['partner_id'] = invoice.commercial_partner_id.id
-            else:
-                res['partner_id'] = invoice.partner_id.id
+            res['partner_id'] = invoice.commercial_partner_id.id
             # we want the move to have the same ref than the found
             # invoice's move, thus it will be easier to link them for the
             # accountants
--- a/account_statement_transactionid_completion/statement.py
+++ b/account_statement_transactionid_completion/statement.py
@@ -110,9 +110,12 @@ class AccountStatementCompletionRule(Model):
         elif len(invoice_id) == 1:
             invoice = invoice_obj.browse(cr, uid, invoice_id[0],
                                          context=context)
-            # FIXME use only commercial_partner_id of invoice in 7.1
-            # this is for backward compatibility in 7.0 before
-            # the refactoring of res.partner
+            # if account_report_company is installed, the field to be
+            # used as partner is the commercial partner; otherwise,
+            # the basic partner must be used.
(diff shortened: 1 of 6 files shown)


account_statement_transactionid_completion/statement.py
```diff
@@ -110,7 +110,13 @@ def get_from_transaction_id_and_invoice(self, cr, uid, st_line,
...
             invoice = invoice_obj.browse(cr, uid, invoice_id[0],
                                          context=context)
-            res['partner_id'] = invoice.partner_id.id
+            # FIXME use only commercial_partner_id of invoice in 7.1
+            # this is for backward compatibility in 7.0 before
+            # the refactoring of res.partner
+            if hasattr(invoice, 'commercial_partner_id'):
+                res['partner_id'] = invoice.commercial_partner_id.id
+            else:
+                res['partner_id'] = invoice.partner_id.id
```
pedrobaeza: I think there is no such 7.1 version, so you can simply put
`res['partner_id'] = invoice.commercial_partner_id.id`

sbidoul: In 7.0 invoice.commercial_partner_id comes from the module account_report_company.

This module is auto_install, but is optional.

mdietrichc2c:
Thanks for the feedback.

Sorry, I wanted to write an explanation: this part is derived from the module account_statement_base_completion, which used this comment and this code to use the commercial partner in the invoice, instead of the partner, if it is present. The goal here is to have the same behavior as the "base" one.

I'll re-work the comment, but I'll leave the "fallback" to the partner_id since account_report_company is optional in 7.0.

Regards,
Matthieu Dietrich

pedrobaeza:
We can make two things:
- Put _account_report_company_ as a dependency and simplify expression
- Or leave it as you said.

I'm not sure which one I prefer. Maybe the first one, because it assures consistent information.

Regards.

mdietrichc2c:
If we have to add it as a dependency, I would prefer to put it in account_statement_base_completion; the same behavior is present, and the module account_statement_transactionid_completion depends on this one.

Although, to be honest, I would prefer not to add a new dependency.

Regards,
Matthieu Dietrich

pedrobaeza:
Yeah, but as you say, in each inherited module you will have to take in account this patch. This dependency is very light, and included on core, so I think it doesn't hurt.

Regards.

mdietrichc2c:
Okay, I added the dependency on _account_statement_base_completion_, and removed the condition both in this module and _account_statement_transactionid_completion_.

Regards.

sbidoul:
:+1: (code review)

pedrobaeza:
Thanks for the changes.

:+1:

gurneyalex:
@mdietrichc2c it would be soooooo nice to have an automated test for this fix... :neutral_face:

mdietrichc2c:
@gurneyalex I'll do a quick test soon.

guewen:
Hey, coverage decreased :hushed:

Perhaps you should create a test to cover your test :grin:

Seriously, the .coveragerc [0] file miss to omit _/test/_ (I wonder if there is the same issue in the other projects)

[0] https://github.com/mdietrichc2c/bank-statement-reconcile/blob/a71ee406173c9bb84138717d3b28f57c88310e92/.coveragerc

guewen:
I'm wrong because it is yaml files. So there is something weird with coveralls.

nbessi:
:+1:

PR #45, opened 2014-09-26
fix unreachable code flagged by pylint

gurneyalex:

PR #46, opened 2014-09-26
fix unreachable code flagged by pylint

gurneyalex:


gurneyalex:
cherry pick of #45

lmignon:
+1:  Just a question. Why a cherry pick and not a forward port. I've the feeling it's safer to always  merge the 7.0 branch to 8.0 (IMO).

gurneyalex:
@lmignon I would tend to agree on a general basis. But I was concerned the porting of modules in the 8.0 branch to the new API would cause lots of conflicts, and I did not want to spend the time fixing these when forward porting a trivial patch. 

You're welcome to prove me wrong, though... :smile_cat:  and my git-fu is not that high, so maybe I'm missing some magical command / switch which does everything and also prepares coffee as a side effect...

vrenaville:
:+1: : thanks for the fix ;)

lmignon:
@gurneyalex This is a recurring question in my mind. What is the best way to manage the replication of changes from one branch to the other, making sure we do not lose some? If every change in 7.0 is followed by a merge in 8.0, you only have to mange the conflicts resulting of your changes and we are sure that every commit in 7.0 were applied in 8.0. 

My question is not related to this change, but would it not be appropriate to define a methodology, a process to ensure that everything done in 7.0 is ported to 8.0? 

Once again :+1: and thanks for the fix :smile:

gurneyalex:
@lmignon would you like to start a discussion on the community mailing list about this? This PR is hardly the place for it :-)

pedrobaeza:
:+1:

PR #47, opened 2014-10-09
Migration of base_transaction_id

guewen:
Port the module to v8.
Use the new API where it has been used in the base model (account.invoice)
and the old API where the base model is still using it.


pedrobaeza:
Good work! :+1:

yvaucher:
:+1:

PR #48, opened 2014-10-09
Port statement voucher killer patch

vrenaville:
In the module voucher killer ,
 when it import payment line into the bank statement it takes the maturity date of the payment line,
to define the bank statement line date, that is wrong it must take the date of the payment line or if not present the date of the payment order.

See bug : https://bugs.launchpad.net/banking-addons/+bug/1301781

### DIFF
--- a/statement_voucher_killer/test/correct_date_in_move_line.yml
+++ b/statement_voucher_killer/test/correct_date_in_move_line.yml
@@ -10,20 +10,22 @@
 -
   I create a statement line for a SO
 -
-  !record {model: account.bank.statement.line, id: statement_line_so}:
+  !record {model: account.bank.statement.line, id: statement_line_date}:
     name: Test autocompletion based on Sale Order Number
     statement_id: statement_test_date1
     ref: DATE1
+    account_id : account.a_sale
     date: '2014-12-20'
     amount: 1000.0
 -
   I confirm the bank statement
 -
   !python {model: account.bank.statement}: |
-    result = self.button_confirm_bank(cr, uid, [ref("statement_test_sale1")])
+    result = self.button_confirm_bank(cr, uid, [ref("statement_test_date1")])
 -
-  Now I can check that all is nice and shiny, line 1. I expect the Sale Order
-  Number to be recognised.
+  Now I can check that that the date in move line is equal to line date
 -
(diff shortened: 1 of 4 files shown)


statement_voucher_killer/test/correct_date_in_move_line.yml
```diff
@@ -0,0 +1,31 @@
...
+    balance_end_real : 1000.0
+-
+  I create a statement line for a SO
+-
+  !record {model: account.bank.statement.line, id: statement_line_date}:
+    name: Test autocompletion based on Sale Order Number
+    statement_id: statement_test_date1
+    ref: DATE1
+    account_id : account.a_sale
+    date: '2014-12-20'
```
guewen: The date should not be fixed because it will fail next year (no period). You should use:

```
date: !eval time.strftime('%Y-12-20')
```

statement_voucher_killer/test/correct_date_in_move_line.yml
```diff
@@ -0,0 +1,31 @@
...
+-
+  !python {model: account.bank.statement}: |
+    result = self.button_confirm_bank(cr, uid, [ref("statement_test_date1")])
+-
+  Now I can check that that the date in move line is equal to line date
+-
+  !python {model: account.move.line}: |
+    line_ids = self.search(cr, uid, [('statement_id', '=', ref('statement_test_date1'))],limit=1)
+    line = self.browse(cr, uid, line_ids)[0]
+    assert line.date == '2014-12-20', "Date (2014-12-20) not equal to = %s" % line.date
```
guewen: Same comment as above.

sbidoul:
I tried to run the test on a fresh database and it fails because it tries to use account.statement.profile which is not in the dependencies of statement_voucher_killer.

To make the test work you would need to add a dependency on bank_statement_ext which would be a bit heavy, in situation where we just need the voucher killer feature.

lepistone:
thanks :+1:

sbidoul:
Hi @vrenaville, thanks a lot for adapting the tests.

Sorry to bother you, but while trying to understand the change with the debugger, I noticed your test does not exercise the 'populate' wizard that are modified by this PR. Maybe I miss something?

vrenaville:
You are right ;) I have improve the test to use the wizard instead ;)

gurneyalex:
:+1:

PR #42, opened 2014-09-05
Port voucher killer modules to 8.0

sbidoul:
invoicing_voucher_killers migrates trivially, which I confirmed with a couple of manual tests.

statement_voucher_killer is not necessary anymore since bank statements do not use vouchers in 8.0.

PR #50, opened 2014-10-09
Migration of statement_voucher_killer to Odoo 8

guewen:
Please review this PR before as it is a pre-requisite (migration of `base_transaction_id`):
- [ ] https://github.com/OCA/bank-statement-reconcile/pull/47


sbidoul:
Hi @guewen,

I had submitted #42 on this same topic a few weeks ago.

A couple of questions. 

Why are the commits on base_transaction_id in this PR (statement voucher killer does not depend on it)?

About statement_voucher_killer the only thing remaining in the module are the date fixes in the 'Populate' wizards, since the 8.0 bank statement do not use vouchers anymore. So IMO these changes should either go into the core as PR, or to OCB, or into a module with another name than statement_voucher_killer.

What do you think?

guewen:
I think that I should be tired and how lucky I have holidays next weeks.

I needed to port `account_payment_transaction_id` so I migrated this one without brain it seems.
I wasn't aware that there is a date fix specifically in this module but indeed it should be moved to the core (even if this module was to keep).

The reason why `account_payment_transaction_id` depends on it are special `_prepare...` methods that do not exists in the core, but I will try to dig around or propose a patch in the core.

Thanks for sharing your brain when I don't use mine! :wink:

PR #57, opened 2014-10-14
Modify "Import invoices" to take statement's date

mdietrichc2c:
This PR is to correct an issue when the statement voucher killer is installed;
the date of the imported line is not the one in the statement (as in Odoo), 
but the current date.


pedrobaeza:
LGTM :+1:

gurneyalex:
have you checked that the overloading of _prepare_statement_line_vals in account_payment_transaction_id are unaffected by this change?

mdietrichc2c:
@gurneyalex good point, I didn't think about dependencies on voucher_killer; I'll revert the method signature.

gurneyalex:
:+1:

mdietrichc2c:
Actually, https://github.com/OCA/bank-statement-reconcile/pull/48 has been merged and corrects the issue. So, this one can be closed without merge.

PR #60, opened 2014-10-29
[FIX] If balance_start is set by parser and equals 0.0, don't override it

ecino:
In account_statement_base_import, there is a problem if a parser wants to set the balance_start to zero, the value won't be retained. I propose to fix that in this small patch.

### DIFF
--- a/account_statement_base_import/statement.py
+++ b/account_statement_base_import/statement.py
@@ -135,7 +135,7 @@ class AccountStatementProfil(orm.Model):
         """
         vals = {'profile_id': profile_id}
         vals.update(parser.get_st_vals())
-        if not vals.get('balance_start'):
+        if 'balance_start' not in vals.keys():
             # Get starting balance from journal balance if parser doesn't
             # fill this data, simulating the manual flow
             statement_obj = self.pool['account.bank.statement']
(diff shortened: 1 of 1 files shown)


account_statement_base_import/statement.py
```diff
@@ -135,7 +135,7 @@ def prepare_statement_vals(self, cr, uid, profile_id, result_row_list,
         """
         vals = {'profile_id': profile_id}
         vals.update(parser.get_st_vals())
-        if not vals.get('balance_start'):
+        if 'balance_start' not in vals.keys():
```
guewen: `if 'balance_start' not in vals:` suffice and is a better idiom that using `keys()`, it will look by default in the keys (using the dictionary hashing vs linear search when using `keys()`).

ecino: You’re right, I corrected it.

guewen:
Thanks for your contrib!

:+1:

sbidoul:
LGTM :+1:

pedrobaeza:
The problem is that parser always return a value in that key, although it can be None: https://github.com/OCA/bank-statement-reconcile/blob/7.0/account_statement_base_import/parser/parser.py#L124

That's why I put `get` sentence instead of the other thing. For now :-1:

lmignon:
@pedrobaeza so the fix should be 

``` python
if vals.get('balance_start') is None:
```

pedrobaeza:
Yeah, I think so. :smile:

lmignon:
@pedrobaeza :smile: 
@ecino you know what to do

ecino:
@lmignon Thanks for the proposition, I corrected the fix.

pedrobaeza:
:+1:

PR #64, opened 2014-11-19
Use the correct accounts for write-off

mdietrichc2c:
A small but important fix: without this, the profit account is used for debit write-off


rdeheele:
it seems coherent :+1:

pedrobaeza:
Thanks for these refinements (waiting for Travis)

lepistone:
thanks :+1:

guewen:
Travis fails for other reasons.

:+1:

PR #68, opened 2014-11-26
7.0 fix write off

vrenaville:
The writeoff comparaison is in the wrong way


lepistone:
:-1: it seems inconsistent with the name of the method: it is called `below_writeoff_limit` to check something like `self.something <= limit`. Maybe the logic can be flipped somewhere else.

vrenaville:
@lepistone it's seem that I have mis understanding the limit. so I close this PR

guewen:
The test is correct and from what I read in the code, the logic is correct too (if amount is below the limit, it will do a partial reconcile which is what we want).

PR #70, opened 2014-11-27
Multi currency exchange writeoff

vrenaville:
Add the possibility to define account for exchange writeoff in order to correct this bug:

https://github.com/OCA/bank-statement-reconcile/issues/69

### DIFF
--- a/account_easy_reconcile/base_reconciliation.py
+++ b/account_easy_reconcile/base_reconciliation.py
@@ -226,7 +226,7 @@ class EasyReconcileBase(orm.AbstractModel):
                 writeoff_account_id = rec.income_exchange_account_id.id
             else:
                 writeoff_account_id = rec.expense_exchange_account_id.id
-            period_id = self.pool.get('account.period').find(
+            period_id = self.pool['account.period'].find(
                 cr, uid, dt=date, context=context)[0]
             if rec.analytic_account_id:
                 rec_ctx['analytic_id'] = rec.analytic_account_id.id
--- a/account_easy_reconcile/base_reconciliation.py
+++ b/account_easy_reconcile/base_reconciliation.py
@@ -222,18 +222,14 @@ class EasyReconcileBase(orm.AbstractModel):
             # lines to reconcile
             # it will do a full reconcile instead of a partial reconcile
             # and make a write-off for exchange
-            if rec.account_id.currency_id:
-                if sum_credit > sum_debit:
-                    writeoff_account_id = rec.income_exchange_account_id.id
-                else:
-                    writeoff_account_id = rec.expense_exchange_account_id.id
-                period_id = self.pool.get('account.period').find(
-                    cr, uid, dt=date, context=context)[0]
-                if rec.analytic_account_id:
-                    rec_ctx['analytic_id'] = rec.analytic_account_id.id
+            if sum_credit > sum_debit:
+                writeoff_account_id = rec.income_exchange_account_id.id
             else:
-                writeoff_account_id = False
-                period_id = False
+                writeoff_account_id = rec.expense_exchange_account_id.id
+            period_id = self.pool.get('account.period').find(
+                cr, uid, dt=date, context=context)[0]
+            if rec.analytic_account_id:
+                rec_ctx['analytic_id'] = rec.analytic_account_id.id
             ml_obj.reconcile_partial(
                 cr, uid,
                 line_ids,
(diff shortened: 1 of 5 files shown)


account_easy_reconcile/base_reconciliation.py
```diff
@@ -215,10 +215,32 @@ def _reconcile_lines(self, cr, uid, rec, lines, allow_partial=False,
...
                         return True, False
 
+            # We need to give a writeoff_acc_id
+            # in case we have a multi currency lines
+            # to reconcile.
+            # If amount in currency is equal between
+            # lines to reconcile
+            # it will do a full reconcile instead of a partial reconcile
+            # and make a write-off for exchange
+            if rec.account_id.currency_id:
```
pedrobaeza: If rec.account_id.currency_id is set, does it mean that this currency is different from company currency? If not, you should check that condition.

vrenaville:
Hello,

Thanks @pedrobaeza  for your review,

I have completely remove the test has I have a test for the currency with required in the view.
The test for company currency is done is reconcile function of module account so I do not add extra test in my function, I let the reconcile function make the test.

account_easy_reconcile/easy_reconcile.xml
```diff
@@ -129,6 +129,9 @@ The lines should have the same amount (with the write-off) and the same referenc
                 <field name="write_off"/>
                 <field name="account_lost_id" attrs="{'required':[('write_off','>',0)]}"/>
                 <field name="account_profit_id" attrs="{'required':[('write_off','>',0)]}"/>
+                <field name="account_has_currency" invisible="1"/>
+                <field name="income_exchange_account_id" attrs="{'required':[('account_has_currency','=',True)],'invisible':[('account_has_currency','=',False)]}"/> 
+                <field name="expense_exchange_account_id" attrs="{'required':[('account_has_currency','=',True)],'invisible':[('account_has_currency','=',False)]}"/>
```
gurneyalex: regarding the visibility of these fields, isn't there a group handling this globally which should be also added to the criteria?

account_easy_reconcile/easy_reconcile.xml
```diff
@@ -147,6 +150,9 @@ The lines should have the same amount (with the write-off) and the same referenc
                 <field name="write_off"/>
                 <field name="account_lost_id" attrs="{'required':[('write_off','>',0)]}"/>
                 <field name="account_profit_id" attrs="{'required':[('write_off','>',0)]}"/>
+                <field name="account_has_currency" invisible="1"/>
+                <field name="income_exchange_account_id" attrs="{'required':[('account_has_currency','=',True)]}"/> 
+                <field name="expense_exchange_account_id" attrs="{'required':[('account_has_currency','=',True)]}"/>
```
gurneyalex: same question about visibility of these fields.

fclementic2c:
functional test Ok :+1:

account_easy_reconcile/base_reconciliation.py
```diff
@@ -215,10 +215,28 @@ def _reconcile_lines(self, cr, uid, rec, lines, allow_partial=False,
...
+            # to reconcile.
+            # If amount in currency is equal between
+            # lines to reconcile
+            # it will do a full reconcile instead of a partial reconcile
+            # and make a write-off for exchange
+            if sum_credit > sum_debit:
+                writeoff_account_id = rec.income_exchange_account_id.id
+            else:
+                writeoff_account_id = rec.expense_exchange_account_id.id
+            period_id = self.pool.get('account.period').find(
```
gurneyalex: self.pool['account.period']

gurneyalex:
automated test would be nice. 

otherwise globally :+1:

vrenaville:
@gurneyalex I haven't time currently to write test, but I have a the group and remove the currency test

vrenaville:
@gurneyalex I will plan to add test at the end of the month, it can be really nice to have test on this critical module.

Vincent

lepistone:
I would be ready to merge that, but a gave a look to the last failures in travis. The last two build have the same failure, seemingly unrelated to this PR, but on odoo only (ocb is OK), while the commit before was on ocb only (odoo is ok). That could be a flickering situation (that I'd love to fix) or the effect of some change in odoo that was caught on ocb only the day after. 

I am restarting yesterday's build and let's see what happens.

lepistone:
Yesterday's build is now green. I just restarted today's build as well.

I'm merging if it passes, but I'd love to find out what makes those builds flicker.

PR #72, opened 2014-12-08
7.0 fix account statement one move

florian-dacosta:
The method def create_move_transfer_lines was never called. The bug was introduced here probably by mistake : 
https://github.com/OCA/bank-statement-reconcile/commit/d0a06fd532f4fc689f6952e6035e48f6a5be49e9

So no transfer lines are created, which is one of the main feature of this module.

### DIFF
--- a/account_statement_one_move/statement.py
+++ b/account_statement_one_move/statement.py
@@ -184,6 +184,7 @@ class AccountBankStatement(orm.Model):
 
     def button_confirm_bank(self, cr, uid, ids, context=None):
         st_line_obj = self.pool['account.bank.statement.line']
+        move_obj = self.pool['account.move']
         if context is None:
             context = {}
         for st in self.browse(cr, uid, ids, context=context):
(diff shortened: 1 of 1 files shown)


sbidoul:
Hi @florian-dacosta,

Is the first commit of this PR on purpose?

florian-dacosta:
@sbidoul not really... Is it a problem for the merge?
If it is, we can close this PR and I will make a new one without the first commit

sbidoul:
@florian-dacosta not really an issue but having a clean history is always nice.

No need to close the PR. A git rebase -i 7.0 followed by a git push -f should do the trick.

Otherwise LGTM.

florian-dacosta:
Ok, it's clean now.

pedrobaeza:
:+1: when Travis green

florian-dacosta:
Travis is green now :)
thanks for the reviews

sebastienbeau:
The fix is ok for me :+1:

PR #73, opened 2014-12-08
7.0 sale order completion

florian-dacosta:
Add a module which allow to link a bank statement line with sale order ids.


pedrobaeza:
Is this module still valid?

pedrobaeza:
No answer, so I close. If there's interest, please reopen.

PR #74, opened 2014-12-10
add possibility to create a dialect for the csv file

florian-dacosta:
Add the possibility to create its own Dialect in the custom parser.
It can be useful especially because `sniffer.sniff(sample_data, delimiters=',;\t')` is not always able to find the right dialect.

### DIFF
--- a/account_statement_base_import/parser/parser.py
+++ b/account_statement_base_import/parser/parser.py
@@ -29,11 +29,15 @@ def UnicodeDictReader(utf8_data, **kwargs):
     pos = utf8_data.tell()
     sample_data = utf8_data.read(2048)
     utf8_data.seek(pos)
-    dialect = sniffer.sniff(sample_data, delimiters=',;\t')
+    if not kwargs.get('dialect'):
+        dialect = sniffer.sniff(sample_data, delimiters=',;\t')
+        del kwargs['dialect']
+    else:
+        dialect = kwargs.pop('dialect')
     csv_reader = csv.DictReader(utf8_data, dialect=dialect, **kwargs)
     for row in csv_reader:
-        yield dict([(key, unicode(value, 'utf-8')) for key, value in
-                    row.iteritems()])
+        yield dict([(unicode(key, 'utf-8'), unicode(value, 'utf-8'))
+                    for key, value in row.iteritems() if key])
 
 
 class BankStatementImportParser(object):
--- a/account_statement_base_import/parser/parser.py
+++ b/account_statement_base_import/parser/parser.py
@@ -36,8 +36,9 @@ def UnicodeDictReader(utf8_data, **kwargs):
         dialect = kwargs.pop('dialect')
     csv_reader = csv.DictReader(utf8_data, dialect=dialect, **kwargs)
     for row in csv_reader:
-        yield dict([(unicode(key, 'utf-8'), unicode(value, 'utf-8'))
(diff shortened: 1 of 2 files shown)


pedrobaeza:
This is not needed, because you can see that `**kwargs` is passed to parser constructor, so you can include already any csv parser argument, that it's going to work.

florian-dacosta:
I am not sure what you mean.
Indeed I do not need to pass the dialect to the `_init__` of the FileParser.

But I still need to pass it to the `UnicodeDictReader`  (line 116) right?

pedrobaeza:
Seeing the code deeply, you're right. What I did in a similar case is to ovewrite some functions to mimic what you're putting now, so it's useful.

:+1:

account_statement_base_import/parser/parser.py
```diff
@@ -29,11 +29,15 @@ def UnicodeDictReader(utf8_data, **kwargs):
...
+        dialect = sniffer.sniff(sample_data, delimiters=',;\t')
+        del kwargs['dialect']
+    else:
+        dialect = kwargs.pop('dialect')
     csv_reader = csv.DictReader(utf8_data, dialect=dialect, **kwargs)
     for row in csv_reader:
-        yield dict([(key, unicode(value, 'utf-8')) for key, value in
-                    row.iteritems()])
+        yield dict([(unicode(key, 'utf-8'), unicode(value, 'utf-8'))
+                    for key, value in row.iteritems() if key])
```
guewen: Why did you put `if key` in there?
I am not sure we want to remove a column if the column header is empty.

pedrobaeza: Well, I see that having empty columns (the same as duplicated columns), are going to be troubled anyway, because dicts don't add duplicated keys, but you're right, there's no sense for removing the first empy column from the dict. Maybe we should put a warning?

florian-dacosta: Well, that is because indeed, it can be troubled for some case.
For example I have a file with unwanted empty column which don't return a string as value. As a consequence, it raises an unicode error.

If this solution is not acceptable, what about something like that : 
`if isinstance(value, str) and isinstance(key, str)`

guewen: And why not `(unicode(key or '', 'utf-8')`?

florian-dacosta: Well, this is a good idea, I will do it if checking the type is not acceptable because it still won't work for my case. But maybe it is too specific to solve this in this generic module.

I thought it would be great to test the type as if it is not empty but filled with something of a wrong type, it will raise an error.
In my case, the value returned is ['']. But I guess we can imagine a lot of files with bad type values without beeing empty.

guewen: The key is really `['']`? How is your csv file?

florian-dacosta: it is the value, the key is None.

guewen: Oh, ok. Then I don't think it is the role of the reader to check / filter that.

florian-dacosta: Ok, I'll change the PR for `(unicode(key or '', 'utf-8')` then

florian-dacosta:
Here, I did what we discussed, does it look good now?

guewen:
Thanks :+1:

PR #76, opened 2014-12-16
check statement_id exist before adding it to the list we return

florian-dacosta:
In case of multi statements like CFONB files, it can happen we do not want to import some specific accounts.
In this case, we just return None in the method def statement_import.
We do not want to add None to the list of ids we return.


guewen:
:+1:

pedrobaeza:
:+1:

yvaucher:
:+1:

PR #79, opened 2015-01-07
Add history line + error message if reconciliation fails

mdietrichc2c:
This PR is to add an history line + a mail thread to easy reconciliation tasks. The reason is that the current cron can "loop" on a task since it takes the one with the oldest history, and no history was added in case of error.

I also added the error message in order to get basic logging on it.

### DIFF
--- a/account_easy_reconcile/easy_reconcile.py
+++ b/account_easy_reconcile/easy_reconcile.py
@@ -25,6 +25,9 @@ from openerp.tools import DEFAULT_SERVER_DATETIME_FORMAT
 from openerp.tools.translate import _
 from openerp import pooler
 
+import logging
+_logger = logging.getLogger(__name__)
+
 
 class EasyReconcileOptions(orm.AbstractModel):
     """Options of a reconciliation profile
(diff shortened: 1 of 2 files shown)


fclementic2c:
:+1:

mdietrichc2c:
The Runbot failed on another module.

sbidoul:
One test fails due to missing 2014 period. Can you fix it?

pedrobaeza:
:+1:

lepistone:
It is indeed another module that fails. We will address it separately.

lepistone:
Thinking about @sbidoul's comment, maybe he is right: the history line is for users and should help to understand if there is a functional problem. On the other hand, we are suppressing the stack trace that would be logged without this try/except. If it is that way, probably yes, it would be good to `logger.exception` the stacktrace.

Another thing: could it be that a DB error leaves a broken cursor so that we don't manage to write our history line?

lepistone:
PS I understand this is a somewhat urgent bugfix: ping me for an approval+merge once it's ready.

lepistone:
For the red build I opened #80.

mdietrichc2c:
@sbidoul @lepistone thanks for the feedback, I added the stack trace in the latest commit.

sbidoul:
:+1:

lepistone:
Thanks @mdietrichc2c.

As for the other point I raised, it is probably enough to log the trace in that case. To be on the safe side, I'd put the `_logger.exception` line right after `except`.

Done that, I'll merge it. Thanks!

mdietrichc2c:
@lepistone done

lepistone:
ok, only the failure we know about. :+1:

PR #83, opened 2015-01-13
Account partner statement import

mdietrichc2c:
This is a new module for importing bank statements. The main difference is there are 2 new columns, "account" and "partner", to add specific partners/accounts per line.


yvaucher:
@mdietrichc2c No reason to do a merge here. Please do a `git rebase` instead

yvaucher:
account_statement_account_partner_import/data/statement.csv is an exemple file ?

It would be great to have a test based on it.

mdietrichc2c:
@yvaucher rebase is done. Re: unit test, there is already one in the base_import.

sbidoul:
This module since to be just 2 additional columns in the generic csv parser?

Would it be possible to implement that in base_import directly?

mdietrichc2c:
@sbidoul my goal was to do something similar to other "extension" of base_import, like account_statement_ofx_import or account_statement_transactionid_import.

sbidoul:
OFX is a different format. Since the csv parser is already in base import and what you are doing here is supporting additional columns it maybe simpler to put that in base import so we have only one place to look at to find the csv parser with all the features.

That's not blocking for me here but in 8.0 I'd really prefer to have one module for csv import of bank statement.

account_statement_account_partner_import/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from . import parser
+from . import statement
```
ehdem, line 22: Better to do:
from . import (
    model_1,
    model_2,
)

yvaucher, line 22: I disagree, current solution is 2 lines only and is fine

max3903:
I agree with @sbidoul. Should be done in base_import directly.

pedrobaeza:
Is this going to be continued?

guewen:
@pedrobaeza is there something to change?

guewen:
@pedrobaeza I mean, this is a PR for 7.0. In 8.0 this can be done in `base_import` and anyway in 9.0 this is a different story as we don't import anymore bank statements.

pedrobaeza:
OK, so do you consider this can be merged or close?

pedrobaeza:
As there's no answer, I close. Feel free to reopen in case there's interest.

Issue #43, opened 2014-09-25
Automatic reconcile delete and replace some partial reconcile done by user manually


vrenaville:
Automatic reconcile delete and replace some partial reconcile done by user manually

To reproduce:
- Create a customer invoice for 10000 € to Customer X (customer account: 4110000) - Add a reference : ABC1234 and validate the invoice to book it into ledger
- Pay partialy that invoice (8000€) using the bank statement...but fill in a different reference (XYZ789) - Confirm the bank statement - the payment entry will be posted.
- Reconcile manually the invoice and the partial payment
- Create a new bank statement to record a 2nd partial payment for that invoice (1500€) but this time the reference is the same than the one use on the invoice entry (ABC1234) .
- Lauch the automatic reconcile on the receivable account

The bug is the following:
- The system will create a partial reconcile between the invoice and 2nd payment without including the 1st payment done manualy reconciled (because this one do not have the same reference).
- the initial partial reconcile is broken because linked to only 1 account move line (the other one being now reconciled with the 2nd payment)

fclementic2c:
+1
This is a big problem we face today with the advanced reconcile module. 
Users expect the system not altering reconciliation manually created !

pedrobaeza:
Do you have a PR?

laetitia-gangloff:
Hello,

I have the following use case :
- an invoice payment
- an invoice fee
- a bank statement

When bank statement is not loaded, the invoice payment and invoice fee are partially reconciled automatically.
Then the bank statement comes. I want the reconciliation take account of the preceding partially reconciled move.

With the proposed fix, it is not the case. Is it possible to propose an option that allow or not (default is not) to rewrite partial reconciled item ?

PR #44, opened 2014-09-25
[FIX] Prevent remove partial reconcile, when you use easy reconcile

vrenaville:
Fixes #43

PR #86, opened 2015-01-22
add partial_reconcile option to allow to remove partial reconcile

laetitia-gangloff:
Hello,

following my comment on #43 I propose this commit with a new option : 'partial_reconcile' to allow to remove partial reconcile.

### DIFF
--- a/account_easy_reconcile/base_reconciliation.py
+++ b/account_easy_reconcile/base_reconciliation.py
@@ -79,17 +79,20 @@ class EasyReconcileBase(orm.AbstractModel):
         return "SELECT %s" % ', '.join(self._base_columns(rec))
 
     def _from(self, rec, *args, **kwargs):
-        return ("FROM account_move_line "
-                "LEFT OUTER JOIN account_move_reconcile ON "
-                "(account_move_line.reconcile_partial_id "
-                "= account_move_reconcile.id)"
-                )
+        res = "FROM account_move_line "
+        if not rec.partial_reconcile:
+            res += ("LEFT OUTER JOIN account_move_reconcile ON "
+                    "(account_move_line.reconcile_partial_id "
+                    "= account_move_reconcile.id)")
+        return res
 
     def _where(self, rec, *args, **kwargs):
         where = ("WHERE account_move_line.account_id = %s "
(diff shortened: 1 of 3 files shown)


guewen:
@vrenaville @fclementic2c any opinion?

Technically ok, but the help on the field should be improved, because it concerns only `manual` partial reconcile, the other ones are already removed.

laetitia-gangloff:
@guewen I think it concerns all partial reconcile (at moment of the PR).
In my use case we use only automatic reconciliation, and partial reconciliation was not unreconciled, and this is the reason of this PR.

guewen:
Isn't the following line in #44 there for that reason:

```
 "AND COALESCE(account_move_reconcile.type,'') <> 'manual' "
```

Anyway, automatic partial reconciliation must always be unreconciled, if it is not the case, it is a bug.
I see an interest in this option only if it concerns the manual partials, when the automatic are always retried.

laetitia-gangloff:
For partial reconcile, the type is always set to 'manual' :
https://github.com/OCA/bank-statement-reconcile/blob/7.0/account_easy_reconcile/base_reconciliation.py#L237

guewen:
> For partial reconcile, the type is always set to 'manual' :
> https://github.com/OCA/bank-statement-reconcile/blob/7.0/account_easy_reconcile/base_reconciliation.py#L237

I see now! I think that's what is wrong. I think they should be 'auto', so partial reconciles created manually are kept and the ones created by the auto-reconcile are unreconciled. What do you think?

vrenaville:
@laetitia-gangloff I think that we need to preserve "real" manual partial reconcile.
I think that we need to correct this bug before:  https://github.com/OCA/bank-statement-reconcile/blob/7.0/account_easy_reconcile/base_reconciliation.py#L237 
all reconcile full or partial done automatically need to be set to "auto".

laetitia-gangloff:
I will change this 'manual' to 'auto'. I think, it will correct my initial problem. 
Then, I think, this PR, can be closed.

vrenaville:
@laetitia-gangloff Thanks for your response, your PR on the change from manual to auto will be appreciated ;)

fclementic2c:
:+1:

PR #87, opened 2015-01-23
Regression fix for #79

ecino:
- Fix error logs that tried to access inexistant attribute.


guewen:
ping @mdietrichc2c

mdietrichc2c:
LGTM (it's true that Exception doesn't have a 'value' attribute).
:+1:

guewen:
:+1:

yvaucher:
:+1:

guewen:
Oups, it was not correct. An Exception has only 1 item in `args`.

``` python
>>> Exception('test')[1]
IndexError: tuple index out of range
```

`orm.except_orm` has 2 items (the title and the message).
Maybe the simplest is just to show all the args, maybe concatenating them with a join.

yvaucher:
Done in #89

PR #88, opened 2015-01-23
Revert "Regression fix for #79"

guewen:
Reverts OCA/bank-statement-reconcile#87


yvaucher:
Otherwise we can go further with #89

PR #92, opened 2015-02-04
Errors args are not always string.

ecino:
The fix proposed last time is still not working as error objects can have non-string arguments.


pedrobaeza:
:+1:

laetitia-gangloff:
:+1:  code only
but it seems there is some comparison problem...

PR #97, opened 2015-02-23
7.0 get first write off date available

florian-dacosta:
When we search for the write off date, during the reconciliation process, we check if the corresponding period is still open. If not, we search for the nearest opened period so the reconciliation can be successful.

### DIFF
--- a/account_easy_reconcile/base_reconciliation.py
+++ b/account_easy_reconcile/base_reconciliation.py
@@ -126,10 +126,42 @@ class EasyReconcileBase(orm.AbstractModel):
         writeoff_amount = round(debit - credit, precision)
         return bool(writeoff_limit >= abs(writeoff_amount)), debit, credit
 
+    def _check_period_state(self, cr, uid, date, context=None):
+        period_id = self.pool['account.period'].find(
+                cr, uid, dt=date, context=context)[0]
+        cr.execute("""
+            SELECT state 
+            FROM account_period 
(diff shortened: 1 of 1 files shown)


account_easy_reconcile/base_reconciliation.py
```diff
@@ -126,9 +126,40 @@ def _below_writeoff_limit(self, cr, uid, rec, lines,
         writeoff_amount = round(debit - credit, precision)
         return bool(writeoff_limit >= abs(writeoff_amount)), debit, credit
 
+    def _is_period_close(self, cr, uid, date, context=None):
+        period_obj = self.pool['account.period']
+        period_id = period_obj.find(cr, uid, dt=date, context=context)[0]
+        period = period_obj.browse(cr, uid, period_id, context=context)
+        if period.state == 'done':
```
pedrobaeza: `return period.state == 'done'`

guewen:
I am not sure that's something that's desirable in any cases. @fclementic2c your opinion?

fclementic2c:
No, you cannot take such assumption florian but I agree the error message when period is closed can be unclear for users.
Today, you can already choose the write off date (today, most recent move lines...) so I see 2 possibilities : either we force the write-of date to 'today' al the time and hide the field or we simply add a message to the user to explain the set-up he has to change and why this happenned. 
I would prefer the second option.

PR #99, opened 2015-03-03
fix travis (flake8)

gurneyalex:


pedrobaeza:
PEP8 is correct, so I merge.

PR #101, opened 2015-03-16
Put the field transaction_id in the group 'sale_pay'

guewen:
Add the field inside the group instead of after a field.  In another module, I
add a button right after 'payment_term' and it must be right after it.  Adding
the field in the group prevent it to be placed between the payment_term field
and the button.


sebastienbeau:
It's ok for me :+1:

pedrobaeza:
:+1:

yvaucher:
:+1:

PR #102, opened 2015-03-23
Set type auto on partial automatic reconciliation

laetitia-gangloff:
Following discussion on #86
I set the value of type to 'auto' for partial reconciliation.


guewen:
:+1: 
Thanks!

vrenaville:
:+1: 
Thanks

mdietrichc2c:
:+1:
Thanks!

PR #103, opened 2015-03-25
protect import

gurneyalex:


pedrobaeza:
You can simply initialize to None, because manifest already declared external dependency.

sbidoul:
Note that in this specific case of statement parsers, the absence of the ofx library does not prevent Odoo to start: it simply displays an error in the log file. So this specific PR is probably not necessary.

sbidoul:
Oops. Forget my previous comment. I thought we were on the 8.0 branch.

pedrobaeza (approved):
This old one is correct and this shouldn't harm the Travis status. @gurneyalex, if you rebase, we can see.

gurneyalex:
@pedrobaeza rebased.

pedrobaeza:
As branch is red, we can simply merge this old one.

PR #104, opened 2015-03-31
Overloaded method onchange_type had an extra argument without default value

ghost:
Added fefault value for argument profile_id, and cleaned a PEP8 issue.

### DIFF
--- a/account_statement_ext/statement.py
+++ b/account_statement_ext/statement.py
@@ -728,7 +728,7 @@ class AccountBankStatementLine(orm.Model):
         line_type = obj_stat.get_type_for_counterpart(
             cr, uid, 0.0, partner_id=partner_id)
         res_type = self.onchange_type(
-            cr, uid, ids, partner_id, line_type, profile_id, context=context)
+            cr, uid, ids, partner_id, line_type, profile_id=profile_id, context=context)
         if res_type['value'] and res_type['value'].get('account_id', False):
             return {'value': {'type': line_type,
                               'account_id': res_type['value']['account_id'],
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I think it's better to just make profile_id a keyword argument, instead of having a new method.

ghost:
<Odoo 7 addon>account/account_bank_statement.py:529
The function signature is def onchange_type(self, cr, uid, line_id, partner_id, type, context=None):

The osv considers this to be the valid signature for the function when using a view onchange property. Because of that, the additional argument is consider to be the context (as it doesn't accepts keyword arguments in a view's XML), instead of the new argument.

In this case it is better to have a new method, for those views that can offer the extra argument, and delegate the call to this new method when a call to the old method is done.

pedrobaeza:
I still think a keyword argument works better. You can call it the old or the new way, and treat inside testing for None for profile_id value.

ghost:
During our test, Odoo consider the extra argument to be the context, and view's XML don't accept arguments by keyword.

pedrobaeza:
It can be if you call it correctly with keyword argument:

```
def onchange_type(self, cr, uid, line_id, partner_id, type, profile_id=None, context=None):
    res = super(X, self).onchange_type(cr, uid, line_id, partner_id, type, context=context)
```

ghost:
I repeat, the problem is in the onchange property of a form view XML.

pedrobaeza:
But you're overwriting that onchange property in all places...

I still don't understand the problem, but I let others to bring clarity to this, because maybe I'm wrong.

pedrobaeza:
As there's no answer in this, I think we can close this.

PR #110, opened 2015-06-12
[ADD] add module 'account_easy_reconcile' : v8 migration

damdam-s:

PR #112, opened 2015-06-15
[MIG] 8.0 portage of module account_advanced_reconcile

yvaucher:
Depends on:
- [x] #110


sbidoul:
Hi @yvaucher,

We have a requirement to automate e-commerce reconciliation where we have
- all e-commerce payments move lines waiting on a transit account
- grouped payments by the payment processor (ie one line on the bank statement for many e-commerce transactions)
- for each grouped payment, we receive a file from the payment processor listing the corresponding ecommerce transaction ids 

In 7.0 we were doing this quite successfully with advanced_reconcile.

So my questions are:
- what's the status of this PR according to you?
- do you know of an alternative approach for the above use case? we'd like to create some kind of generic module to handle that

cc/ @laetitia-gangloff @lmignon

yvaucher:
@sbidoul I haven't touched this since and the project for which I gave an occasional hand will be launched near 1st January. I have no idea if this was fully tested or not. Though, it shouldn't have changed much since 7.0.  And sorry I don't know of an alternative.

@damdam-s @fclementic2c can you provide more info on this?

laetitia-gangloff:
@yvaucher is it possible to rebase your PR ?
I would make some tests.

yvaucher:
@laetitia-gangloff I have rebased it and @damdam-s gave me some fixup

laetitia-gangloff:
Thank you :)

JordiBForgeFlow:
Hi @laetitia-gangloff did you make any progress on adding unit tests? We want to use the advance reconcile to reconcile goods received not invoiced account.

laetitia-gangloff:
@jbeficent no unit test, only some little functional test, and I it is always a nice module. But we will take another way for our use cases, and I am not going further :(

jgrandguillaume:
Hi,

@sbidoul We also used it in version 7.0 for the use case you mentioned above. In version 8.0. we never had the need for that, so I've no feedback.

On version 9.0, Sebastien Beau is suggesting something else, would be good to have your opinion here @jbeficent @sbidoul :

https://github.com/OCA/bank-statement-reconcile/issues/129

Then, the code here seems good, no test, I'm :+1: 

Regards,

Joël

vrenaville:
:+1:

JordiBForgeFlow:
We are using this module in 8.0 to reconcile the Goods Received Not Invoiced GL account in the context of real time inventory valuation (see http://www.eficent.com/blog_en/perpetual-inventory-management-with-odoo-reconcile-your-goods-received-not-invoiced-grni-account/). So we consider that this tool is essential for us.

So far, it works good for us.
:+1:

yvaucher:
Thus feel free to merge it . Thanks for the reviews.

yvaucher:
Can this be merged ? I think it is ready since a while. I don't want to self merge it.

JordiBForgeFlow:
@yvaucher Thanks for the great module! Sorry, the PR was left unattended.

PR #116, opened 2015-07-07
[IMP] Changed the name(OBI) and ref(Reference) fields from statement …

sadamo:
…line to be dynamically imported from the ofx file.

This change was made because the .ofx files from Brazil where diferents. 
This way, the module becomes more flexible.

### DIFF
--- a/account_statement_ofx_import/parser/ofx_parser.py
+++ b/account_statement_ofx_import/parser/ofx_parser.py
@@ -61,8 +61,10 @@ class OfxParser(BankStatementImportParser):
             res.append({
                 'date': transaction.date,
                 'amount': transaction.amount,
-                'ref': getattr(transaction, self.profile.ofx_parser_ref),
-                'label': getattr(transaction, self.profile.ofx_parser_label),
+                'ref': getattr(transaction, self.profile.ofx_parser_ref,
+                               transaction.type),
+                'label': getattr(transaction, self.profile.ofx_parser_label,
+                                 transaction.payee)
             })
         self.result_row_list = res
         return True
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Interesting change, but you should provide a default value for the fields that guarantees that existing profiles don't break and newcomers to this module don't lose in the technical part of the OFX file.

account_statement_ofx_import/statement.py
```diff
@@ -19,12 +19,23 @@
...
 class AccountStatementProfil(orm.Model):
     _inherit = "account.statement.profile"
 
+    _columns = {
+        'ofx_parser_label': fields.char('OFX parser Label field',
+                                        required=True),
+        'ofx_parser_ref': fields.char('OFX parser Ref field', required=True),
+    }
+
+    _defaults = {
```
sadamo, line 34: @pedrobaeza
The original values for the fields, are setted as defaults here.

account_statement_ofx_import/parser/ofx_parser.py
```diff
@@ -61,8 +61,8 @@ def _parse(self, *args, **kwargs):
             res.append({
                 'date': transaction.date,
                 'amount': transaction.amount,
-                'ref': transaction.type,
-                'label': transaction.payee,
+                'ref': getattr(transaction, self.profile.ofx_parser_ref),
```
pedrobaeza: You should protect this in case of the provided label is incorrect.

pedrobaeza:
Ah, sorry, I suppose I was expecting new API and then didn't see the old defaults section.

pedrobaeza:
I have added the account-financial-tools repo to runbot to get a green result in next commit.

pedrobaeza:
This is still in red. @sadamo, are you still interested in this change? If so, please rebase and check Travis status.

pedrobaeza:
Closing as no answer.

PR #117, opened 2015-07-22
[ADD] account_reconcile_different_partners

hbrunn:
# Reconcile move lines from different partners

While it is in general not a good idea to reconcile move lines from different partners, there can be situations where this is not avoidable. This module allows you to do that.
# Configuration

By default, only finance managers are allowed to reconcile different partners' move lines. If you need other people to do that, add them to the group _Technical/Reconcile different partners' move lines_.
# Known issues / Roadmap
- enforcing the same partner for reconciliation avoids mistakes, so assign this group only to people who know what they are doing
- it would be nice to have warnings when selecting different partners


fclementic2c:
:-1: 

Sorry but I have a big problem with that.
This feature is not compatible with the webkit financial reports (you will have differences between partner ledger report and partner balance).

hbrunn:
so much worse for that module I'd say. You are aware that different partners in opening periods are allowed by odoo standard?

sbidoul:
@brunn I agree with @fclementic2c here. When reconciliations between different partners must occur, it can and must be done on accounts that are not of type payable/receivable.

Regarding the opening periods, the self reconciliation that occurs there is just a technical side-effect of the way opening entries work in Odoo, and IMO must not be extended to other situations.

hbrunn:
@sbidoul "[...] must be done on accounts that are not of type payable/receivable" is this a local juristical issue or a technical one?

sbidoul:
@hbrunn ~~sorry, I should have written "_can_ be done on accounts that are not of type payable/receivable".~~

If you mix partners in reconciliations on receivable/payable accounts, you could get an Open Invoices report that shows nothing for such partner (because all reconciled) while the Partner Balance report would show non zero amounts for such partners. Not desirable IMO.

pedrobaeza:
@hbrunn is this still needed? There are question without answers, so I tend this to close it.

PR #119, opened 2015-08-04
[ADD] account_reconcile_payment_order

hbrunn:
# Reconcile payment orders

Payment orders that show up as as one big transaction can be difficult for the accounting to handle if a transfer acount is used. In this case, we need to reconcile this transaction with possibly hunderds of move lines, which can be a bit tiresome. This module tries to recognize transactions deriving from payment orders and propose the unreconciled move lines from this payment order.
# Usage

It should just work. What the module does is to search for a payment order in state 'sent' that has the same amount as the statement line, and the same bank account than the statement.

For further information, please visit:
- https://www.odoo.com/forum/help-1
# Known issues / Roadmap
- it would be good to check references too, but at least the bank in use here changes some characters, so this doesn't seem to be a general solution.


hbrunn:
I also considered to make this a PR against account_banking_payment_transfer directly, because you probably always want this if you use batch transfers. What do you think?

pedrobaeza:
Name should be _account_bank_statement_reconcile_batch_payment_ to accommodate to conventions and to generalize for other operations in batch that also generates only one statement entry.

account_reconcile_payment_order/__openerp__.py
```diff
@@ -0,0 +1,33 @@
...
+##############################################################################
+{
+    "name": "Reconcile payment orders",
+    "version": "1.0",
+    "author": "Therp BV,Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "category": "Accounting & Finance",
+    "summary": "Automatically reconcile all lines from sent payment orders",
+    "depends": [
+        'account_banking_payment_transfer',
```
pedrobaeza, line 28: Why do you need this dependency?

hbrunn, line 28: Because of https://github.com/OCA/bank-statement-reconcile/pull/119/files#diff-78d4c467126eaf80cce3194e883d8fbaR59 cf https://github.com/OCA/bank-payment/blob/8.0/account_banking_payment_transfer/model/account_payment.py#L101

pedrobaeza, line 28: Thanks for the clarification. Then, definitively the module name should be _account_bank_statement_reconcile_payment_order_. We can't remove the bank_statement part from the name because we have several kind of reconciliations, and this one is for bank statements.

pedrobaeza:
It's too much difficult to include a test for this module?

pedrobaeza:
I have tried on runbot and it's working perfect

:+1: and with tests, better yet

hbrunn:
tests should be done. blame the author for laziness for not having them included in the firsrt place

account_reconcile_payment_order/README.rst
```diff
@@ -0,0 +1,51 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+Reconcile payment orders
+========================
+
+Payment orders that show up as as one big transaction can be difficult for the accounting to handle if a transfer acount is used. In this case, we need to reconcile this transaction with possibly hunderds of move lines, which can be a bit tiresome. This module tries to recognize transactions deriving from payment orders and propose the unreconciled move lines from this payment order.
```
pedrobaeza, line 6: s/hunderds/hundreds

pedrobaeza:
Please add `Installation` section mentioning that this module requires _account_payment_transfer_ module in OCA/bank-payment

pedrobaeza:
Hi, @hbrunn, I have tried it with a real case, and found a problem that we should address: this only works when there's a transfer account defined in the payment mode. Although it can be desirable to move balances to an intermediate account, not always all accountants make this operation, and not only this, system allows you not to do it and let the original move lines unreconciled. So we should handle also this case on the module.

I have made https://github.com/hbrunn/bank-statement-reconcile/pull/1 for taking this into account. Please check it.

account_reconcile_payment_order/README.rst
```diff
@@ -0,0 +1,51 @@
...
+
+Usage
+=====
+
+It should just work. What the module does is to search for a payment order in state 'sent' that has the same amount as the statement line, and the same bank account than the statement.
```
pedrobaeza, line 11: Please include new "Try me on runbot" button

account_reconcile_payment_order/__openerp__.py
```diff
@@ -0,0 +1,33 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    "name": "Reconcile payment orders",
+    "version": "1.0",
```
pedrobaeza, line 22: Please use new version numbering scheme: 8.0.1.0.0

pedrobaeza:
@hbrunn, can you tell me something about this? May I superseed your PR with my changes?

hbrunn:
@pedrobaeza sorry, I'm mostly without a network connection currently. Go ahead and do as you see fit

pedrobaeza:
Superseeded by https://github.com/OCA/bank-statement-reconcile/pull/124

PR #120, opened 2015-08-06
New reconciliation rule, searching within the name

mdietrichc2c:
New module, redefining the basic "Advanced. Partner and Ref." rule in `account_advanced_reconcile` to search the credit's reference in the debit's name or reference, instead of having the same value.


pedrobaeza:
Is this still relevant or should we close it?

PR #123, opened 2015-09-09
[ADD] account_reconcile_prepare_account

hbrunn:
# 
# Prepare accounts before reconciliation

This module supoprts a workflow for bank statements where one person with few
permissions already assigns accounts (analytic and non-analytic) to bank
transactions, so that another person only needs to review the assignment,
select the correct moves (filtered by the assigned partner and account), and
then confirm the reconciliation.

Those preassigned accounts are also used as default values for newly created
move lines during reconciliation.
# Usage

Select accounts on the bank statement form, and offered moves for
reconciliation will be filtered accordingly.


moylop260:
Travis red. Is a know issue?

hbrunn:
@moylop260 thanks, I forgot to push the last commit

moylop260:
:+1:

max3903:
:+1:

PR #124, opened 2015-10-06
[8.0] account_reconcile_payment_order

pedrobaeza:
Superseed https://github.com/OCA/bank-statement-reconcile/pull/119
# 
# Reconcile payment orders

Payment orders that show up as as one big transaction can be difficult for the
accounting to handle if a transfer acount is used. In this case, we need to
reconcile this transaction with possibly hundreds of move lines, which can be a
bit tiresome. This module tries to recognize transactions deriving from payment
orders and propose the unreconciled move lines from this payment order.
# Usage

It should just work. What the module does is to search for a payment order in
state 'sent' that has the same amount as the statement line, and the same bank
account than the statement.
# Known issues / Roadmap
- it would be good to check references too, but at least the bank in use here
  changes some characters, so this doesn't seem to be a general solution.


sergio-teruel:
:+1:

account_reconcile_payment_order/README.rst
```diff
@@ -0,0 +1,67 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+========================
+Reconcile payment orders
+========================
+
+Payment orders that show up as as one big transaction can be difficult for the
```
antespi: s/as as/as/

account_reconcile_payment_order/README.rst
```diff
@@ -0,0 +1,67 @@
...
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/bank-statement-reconcile/issues/new?body=module:%20account_reconcile_payment_order%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+Credits
+=======
+
+Contributors
+------------
+
+* Holger Brunn <hbrunn@therp.nl>
```
antespi: @pedrobaeza You should be included as contributor, isn't it?

antespi:
:+1: Tested in runbot, minor remarks

account_reconcile_payment_order/README.rst
```diff
@@ -0,0 +1,67 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+========================
+Reconcile payment orders
+========================
+
+Payment orders that show up as as one big transaction can be difficult for the
+accounting to handle if a transfer acount is used. In this case, we need to
```
rafaelbn: s/acount/account

rafaelbn:
:+1:

account_reconcile_payment_order/README.rst
pedrobaeza: Yeah, I forgot it!

pedrobaeza:
As I'm not the original author and it has been tested/expanded by me and checked by others, I merge.

PR #126, opened 2015-10-16
Restore 8.0 were it was replaced wrongly

yvaucher:
This has been removed automatically but we want 8.0 here not 9.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,10 +4,10 @@
 Odoo modules for statements tasks (completion, reconciliation)
 ==============================================================
 
-***Important notice: since version 9.0, the import feature has moved here: https://github.com/OCA/bank-statement-import***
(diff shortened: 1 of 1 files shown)


pedrobaeza:
The migration script makes a blindly replace, so that's why it has this value. You can replace 8.0 by 8 without losing meaning, and thus, we avoid future replacements. for version 10 and so on.

yvaucher:
@pedrobaeza I disagree to replace dumbly all 8.0 by 8. By the way it won't change for version 10.0 as we will be looking on 9.0. And by the way 7 is named 7.0 below

pedrobaeza:
OK, you're right about that appreciation that on v10 migration this won't change. OK for me.

StefanRijnhart:
Trivial, no? :+1:

yvaucher:
@StefanRijnhart Yup just did a PR to make a trace to warn it might have to be fixed in other repo

PR #133, opened 2016-02-03
[FIX] the line's bank account is the other party's bank account, not ours

hbrunn:
We can't really do this check to my knowledge, as a bank statement only knows the journal it belongs to, but the journal doesn't necessarily know for which bank account it's meant.


lfreeke:
:+1:

pedrobaeza:
:+1: 

Fast-tracking

PR #134, opened 2016-03-10
[9.0] add german translations, from *.pot if present

rschnapka:
havn't seen a Transifex-repo, so I chose to add in source.


pedrobaeza:
This should be done on Weblate, so closing.

PR #137, opened 2016-04-28
[9.0] account move base import

mdietrichc2c:
This PR is to migrate multiple modules as one:
- `account_statement_base_import`
- `account_statement_base_completion`
- `account_statement_commission`
- parts of `account_statement_ext` (profile attributes + mass insertion/update of lines)

All are now grouped under `account_statement_base_import`

The main changes are:
- instead of creating / auto-completing a bank statement, with a defined profile, Journal Entries are now used (since the goal is to import a lot of lines with commission amount), and instead of the profile, the journal is now used.
- two booleans were added, in order to use a journal for import, completion or both.

Other migrated modules are `account_statement_bankaccount_completion`, `account_statement_so_completion`, and `account_statement_transactionid_import` (which also contains `account_statement_transactionid_completion`).

Left to do:
- rename the module to something more "on point", like `credit_statement_import` ? (if you have suggestions, they are welcome)


account_statement_base_import/models/account_move.py
```diff
@@ -0,0 +1,423 @@
...
+            self.env.cr.executemany(sql, tuple(move_store))
+        except psycopg2.Error as sql_err:
+            self.env.cr.rollback()
+            raise ValidationError(_("ORM bypass error"),
+                                  sql_err.pgerror)
+
+    def _update_line(self, vals):
+        """ Do raw update into database because ORM is awfully slow
+            when cheking security.
+        TODO / WARM: sparse fields are skipped by the method. IOW, if your
```
guewen: isn't this comment outdated? I guess there is no sparse fields any more?

pedrobaeza:
Why not use the Odoo core statement import option?

mdietrichc2c:
@pedrobaeza the main issue with the core statement import is that reconciliation must take place when the statement is imported ; in our 7.0 cases with the old `account_statement_base_import`, we often imported files containing payments for many invoices (> 1000), so doing this synchronously would cripple the system.

The goal is to use the journal entry instead, which allows to import the file quickly, then auto-complete the partner depending on completion rules, then use the `account_mass_reconcile` module to process the reconciliations asynchronously.

pedrobaeza:
That's not true. There's only a matching for the partner part, that you can even inhibit if you want via a custom module.

I don't agree to add another import framework that confuses developers on what to use, so :-1:

account_statement_bankaccount_completion/tests/test_bankaccount_completion.py
```diff
@@ -20,55 +20,52 @@
...
     def setUp(self):
-        super(bankaccount_completion, self).setUp()
+        super(BankAccountCompletion, self).setUp()
+        tools.convert_file(self.cr, 'account',
+                           get_module_resource('account', 'test',
+                                               'account_minimal_test.xml'),
+                           {}, 'init', False, 'test')
+        self.account_move_obj = self.env["account.move"]
+        self.account_move_line_obj = \
+            self.env["account.move.line"]
```
yvaucher: As you call it every time with a context, you might want to prepare this context here

As you use multiple time the same `with_context`, you can do something like:

```
self.move_line_obj = self.env['account.move.line].with_context(...)
```

account_statement_base_import/models/account_journal.py
```diff
@@ -0,0 +1,280 @@
...
+        comodel_name='account.account',
+        string='Commission account')
+
+    import_type = fields.Selection(
+        __get_import_type_selection,
+        string='Type of import',
+        default='generic_csvxls_so',
+        required=True,
+        help="Choose here the method by which you want to import bank"
+        "statement for this profile.")
```
yvaucher: Missing space after bank or before statement

account_statement_base_import/models/account_move.py
```diff
@@ -0,0 +1,419 @@
...
+    def __get_functions(self):
+        """ Call method which can be inherited """
+        return self._get_functions()
+
+    sequence = fields.Integer(
+        string='Sequence',
+        help="Lower means parsed first.")
+    name = fields.Char(
+        string='Name',
+        size=128)
```
yvaucher: size limit can be removed.

account_statement_base_import/models/account_move.py
```diff
@@ -0,0 +1,419 @@
...
+            number_field = 'number'
+        else:
+            raise ValidationError(
+                _('Invalid invoice type for completion: %') % inv_type)
+
+        invoices = inv_obj.search([(number_field, '=', line.name.strip()),
+                                   ('type', 'in', type_domain)])
+        if invoices:
+            if len(invoices) == 1:
+                invoice = invoices[0]
```
yvaucher: you can directly return `invoices` as you know it is a recordset with one element

Or directly return `invoices[0]` for more clarity

account_statement_base_import/models/account_move.py
```diff
@@ -0,0 +1,419 @@
...
+    _inherit = "account.move.line"
+    _order = "already_completed desc, date asc"
+
+    already_completed = fields.Boolean(
+        string="Auto-Completed",
+        default=False,
+        help="When this checkbox is ticked, the auto-completion "
+        "process/button will ignore this line.")
+
+    def _get_line_values_from_rules(self, line, rules):
```
yvaucher: Can't we make this method an api.multi ? And thus use `self` as a recordset to replace `line`.

account_statement_base_import/models/partner.py
```diff
@@ -0,0 +1,21 @@
...
+
+class ResPartner(models.Model):
+    """Add a bank label on the partner so that we can use it to match
+    this partner when we found this in a statement line.
+    """
+    _inherit = 'res.partner'
+
+    bank_statement_label = fields.Char(
+        string='Bank Statement Label',
+        size=100,
```
yvaucher: Any reason to keep the size here ?

account_statement_base_import/models/account_move.py
guewen: By default a method is a `api.multi`, the decorator only serves for the compatibility with the old api.
I agree that `line` should just be `self` and adding the decorator is anyway better for explicitness.

account_statement_base_import/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_account_bank_st_cmpl_user,account.move.completion.rule,model_account_move_completion_rule,account.group_account_user,1,0,0,0
+access_account_bank_st_cmpl_manager,account.move.completion.rule,model_account_move_completion_rule,account.group_account_manager,1,1,1,1
```
yvaucher: Would be nice to have 2 different names both are named `account.move.completion.rule` Please at least add `user` and `manager` respectively.

account_statement_base_import/tests/__init__.py
```diff
@@ -20,8 +20,10 @@
 #
 #
 
+from . import test_base_completion
 from . import test_base_import
 
 checks = [
```
yvaucher: This can be removed. All file starting with `test_` will be loaded as tests.

account_statement_base_import/views/account_move_view.xml
```diff
@@ -0,0 +1,67 @@
...
+            </field>
+            <button name="button_cancel" position="after">
+                <button name="button_auto_completion"
+                        string="Auto Completion"
+                        type="object"
+                        class="oe_highlight"
+                        groups="account.group_account_invoice"
+                        attrs="{'invisible': ['|', ('used_for_completion','=',False), ('state','not in', ['draft'])]}"/>
+            </button>
+            <xpath expr="/form/sheet/notebook/page/field[@name='line_ids']/tree/field[@name='credit']" position="after">
```
yvaucher: In such case I prefer to use more relative paths.
Here you could have:
`//field[@name='line_ids']/tree/field[@name='credit']`
Thus if someone move the field outside notebook it will still work.

account_statement_base_import/wizard/import_statement.py
```diff
@@ -23,98 +23,82 @@
...
+        comodel_name='account.journal',
+        string='Import configuration parameter',
+        required=True)
+    input_statement = fields.Binary(
+        string='Statement file',
+        required=True)
+    partner_id = fields.Many2one(
+        comodel_name='res.partner',
+        string='Credit institute partner')
+    file_name = fields.Char('File Name', size=128)
```
yvaucher: size to remove

yvaucher: String is optional as `file_name` field name will fill string value as `File Name`

account_statement_base_import/wizard/import_statement.py
```diff
@@ -23,98 +23,82 @@
...
-        action_obj = self.pool.get('ir.actions.act_window')
-        action_id = model_obj.get_object_reference(
-            cr, uid, 'account', 'action_bank_statement_tree')[1]
-        res = action_obj.read(cr, uid, action_id)
-        res['domain'] = res['domain'][:-1] + ",('id', 'in', %s)]" % sid
-        return res
+        moves = self.env['account.move']
+        for importer in self:
+            journal = importer.journal_id
+            ftype = self._check_extension(importer.file_name)
```
yvaucher: Could be `importer._check_extension()` by making it a `api.multi` method

account_statement_base_import/wizard/import_statement_view.xml
```diff
@@ -1,41 +1,36 @@
...
+                    <field name="journal_id" on_change="onchange_journal_id(journal_id)" domain="[('used_for_import', '=', True)]"/>
+                    <field name="input_statement" filename="file_name" colspan="2"/>
+                    <field name="file_name" colspan="2" invisible="1"/>
+                    <separator string="Import Parameters Summary" colspan="4"/>
+                    <field name="partner_id" readonly="1"/>
+                    <field name="receivable_account_id" readonly="1"/>
+                    <field name="commission_account_id" readonly="1"/>
+                </group>
+                <separator string="" colspan="4"/>
+                <group colspan="4" col="6">
```
yvaucher: To move in `<footer>`

account_statement_base_import/wizard/import_statement_view.xml
```diff
@@ -1,41 +1,36 @@
...
+                    <button  icon="gtk-cancel" special="cancel" string="Cancel"/>
+                    <button  icon="gtk-ok" name="import_statement" string="Import statement" type="object"/>
```
yvaucher: Change button order Cancel on right

yvaucher: Add `class="oe_highlight"`

account_statement_so_completion/models/account_move.py
```diff
@@ -60,27 +58,14 @@ def get_from_ref_and_so(self, cr, uid, st_line, context=None):
...
-                    master_account_id=st_line['master_account_id'],
-                    partner_id=res.get('partner_id', False),
-                    line_type='customer',
-                    amount=st_line['amount'] if st_line['amount'] else 0.0,
-                    context=context)
-                res.update(st_vals)
+        so_obj = self.env['sale.order']
+        orders = so_obj.search([('name', '=', line.name)])
+        if len(orders) > 1:
+            raise ErrorTooManyPartner(
```
yvaucher: This exception seems wrong. Here we have too many sale orders not partners. Maybe we simply want a `self.ensure_one()` in such case.

yvaucher: `ensure_one` might be too technical to help the user though.

yvaucher:
Great work!

Only few comments otherwise LGTM

mdietrichc2c:
@yvaucher thanks for the review, I corrected those points.

jgrandguillaume:
Hi,

@sebastienbeau Here is the "rework" of yours by @mdietrichc2c : can you please have a look and give your input ?

@pedrobaeza May be some misunderstanding: the goal here is to import move line, not statement line any more. With the Odoo approach, not possible to import thousands of statement lines because when you open the reconcile JavaScript view, it breaks. So it was discussed here : https://github.com/OCA/bank-statement-reconcile/issues/129  that we'll make it through move lines directly. May be the module must be renamed (account_move_completion and so on) and may be it should move to account-financial-tools I don't know. May be it is good to keep it here cause people were used to look here for such a purpose. Can you please re-consider your :-1: ?

On my side, the code is clean, tests have been ported and functionally speaking it is working like a charm, so I'm :+1: 

Regards,

Joël

pedrobaeza:
@jgrandguillaume, I have used the JS reconciliation widget with thousands of statement lines without problems, as the match is done 20 by 20 lines, so that's not really a problem.

The discussion you mention is about another object for remittances. I would agree to add these modules for not totally standing against, but renaming the modules like you mention to `account_move...` instead of `account_statement...`

fclementic2c:
Hello,

Please find my comments : 

*Entries uses the wrong default account (should be the opposite) :  
image
- Error  if in the csv template you add a positive amount by error in the column commission_amount and the total of the commision_amount becomes positive. 
  _(File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3147613-137-43f8bb/openerp/addons/account/models/account_move.py", line 193, in assert_balanced
  raise UserError(_("Cannot create unbalanced journal entry."))_
- No import in multicurrency possible : wanted feature ?
- Menu 'Import bank statement' must be renamed to : Import batch file 
- replace 'Counterpart line' -> '/'
- replace 'Commission line' -> 'Commission' (so its works in french and english)

fclementic2c:
@mdietrichc2c : It is all fixed (except 1 minor point see below) ! great. 
You forgot that point also : 'Menu 'Import bank statement' must be renamed to : Import batch file'. Import bank statement is too confusing with the other tool dedicated for this.

However I see a new issue : 
I cannot find the 2 most important autocompletion rules:
- 'Match from line label (based on SO number)
- 'Match from line label (based on transaction ID number)
  is it in some extra modules ?

Also please add in read me that this tool do not handle multicurrency

thanks

fclementic2c:
autocompletion by SO in a separate module : tests Ok

So after the 2 small outstanding points indicated previously it will be Ok for me:  
-  'Menu 'Import bank statement' must be renamed to : Import batch file'. Import bank statement is too confusing with the other tool dedicated for this.
- add in read me that this tool do not handle multicurrency

thxs

mdietrichc2c:
@fclementic2c menu was renamed + readme mentions that multicurrency is not supported.

yvaucher:
:+1:

yvaucher:
Tests are failing on last commit

account_move_base_import/models/account_journal.py
```diff
@@ -0,0 +1,287 @@
...
+    used_for_import = fields.Boolean(
+        string="Journal used for import")
+
+    commission_account_id = fields.Many2one(
+        comodel_name='account.account',
+        string='Commission account')
+
+    import_type = fields.Selection(
+        [('generic_csvxls_so', 'Generic .csv/.xls based on SO Name')],
+        string='Type of import',
```
florian-dacosta, line 29: Why "SO Name" is the description of the key?
This is just usefull to import a specific file format right? It confused me a little.

account_move_base_import/models/account_journal.py
```diff
@@ -0,0 +1,287 @@
...
+        default='generic_csvxls_so',
+        required=True,
+        help="Choose here the method by which you want to import bank "
+        "statement for this profile.")
```
florian-dacosta: We should not speak about bank statement or profile anymore.
A lot of description are still the old ones. I don't think it should block the PR, but we should fix it. (I can do it once it is merged if needed)

account_move_base_import/models/account_journal.py
```diff
@@ -0,0 +1,287 @@
...
+        move_line_obj = self.env['account.move.line']
+        global_commission_amount = 0
+        total_amount = 0
+        for row in parser.result_row_list:
+            global_commission_amount += float(
+                row.get('commission_amount', '0.0'))
+            total_amount += float(
+                row.get('amount', '0.0'))
+        total_amount += global_commission_amount
+        partner_id = self.partner_id.id
```
florian-dacosta, line 179: I don't get this part.
parser.result_row_list is the file content right? Si the key 'commission_amount' or 'amount' only work for the generic CSV format file right? 
If I make a module with a custom format where the fields names are différent, it won't create any extra move lines.
Also, I guess the computation of commission could be more complicated (May some format do not have a special column for this, but we need to do 'brut amount' - 'net amount' to get it).

So I guess we should have a special method on the parser level to get the commission.
And for the total amount, we should compute it from move_store variable (in _move_import method) and commission amount.

account_move_base_import/models/account_journal.py
```diff
@@ -0,0 +1,287 @@
...
+          statement line.
+        """
+        move_line_obj = self.env['account.move.line']
+        values = parser_vals
+        values['company_id'] = self.company_id.id
+        values['currency_id'] = self.currency_id.id
+        values['company_currency_id'] = self.company_id.currency_id.id
+        values['journal_id'] = self.id
+        values['move_id'] = move.id
+        values['account_id'] = self.receivable_account_id.id
```
florian-dacosta: In the generic parser, there is a part adding account_id in the parser_vals, but the data is lost anyway here no?

account_move_base_import/models/account_journal.py
mdietrichc2c, line 179: Actually, for this part, it's already the wanted goal ; the method is "_write_extra_move_lines", which could not even be commission lines, or the counterpart ; you could specify what to add to the file as you wanted (for example, you could have the file with both the commission line and the counterpart in the file itself) for each type of import.

So I'm for keeping the current method, as it allows for more flexibility.

account_move_base_import/models/account_journal.py
mdietrichc2c: Correct ; I'll only set it if it was not set before.

account_move_base_import/models/account_journal.py
mdietrichc2c: I corrected this one (and tried to remove mentions of profile/statements as much as possible), but some may be left. I do agree that those could be removed at a later date.

account_move_base_import/models/account_journal.py
florian-dacosta, line 179: I still don't get it...
Let's take an example : If I want to implement a new file format. The file contain a column with the commission amount, which is called 'commission_total'.
And the transaction amount is called 'transaction_amount'

How should I do to create the commission extra line and the counterpart extra line?

mdietrichc2c, line 179: It would be a bit long, but here's basically how it works:
- in your module, you would inherit `account.journal`, and add an extra import type (for the example, let's call it "example").
- then, in `_write_extra_move_lines`, you would start with an if-condition on that import type. If the type is not "example", you call `super()._write_extra_move_lines()`, else, you can define your own import.

florian-dacosta, line 179: Ok, in this case, I would add an "if self.import_type == generic_csvxls_so" condition already now, as what is done is specific for generic_csvxls_so type and not the other types. It would be more logical to do the same.
And it could cause problems, if my "example" file I have a column "commission_amount" I do not want to use, it will fail.

Doing that, it would be great to add methods _get_counter_part_values and _get_commission_line_values
So if we need to do the same for another type, it would be less redondant.

Also, I think most of the  types we could implement will need a counterpart creation. I wonder if it would not be better to add the option on the journal.

mdietrichc2c, line 179: I'm not against it, but it should be done at a later date ; right now, this PR is 2 months old, and I'm already over budget development-wise.

account_move_base_import/parser/generic_file_parser.py
```diff
@@ -0,0 +1,82 @@
...
+            partners = partner_obj.search([('name', '=', line['partner'])])
+            if len(partners) == 1:
+                partner_id = partners[0].id
+
+        amount = line.get('amount', 0.0)
+        return {
+            'name': line.get('label', '/'),
+            'date_maturity': line.get('date', datetime.datetime.now().date()),
+            'credit': amount > 0.0 and amount or 0.0,
+            'debit': amount < 0.0 and amount or 0.0,
```
florian-dacosta: I think it should be '-amount' for the debit

mdietrichc2c:
@florian-dacosta thanks for the improvements, and for signaling the issue with debit amounts ; both are fixed in the branch now.

florian-dacosta:
I did some tests and code review, it seems good to me now.
:+1:

fclementic2c:
:+1:

PR #140, opened 2016-05-16
fixed expense account id

levkar:
fixed expense account id (somehow it was income, probably from copy/paste)


levkar:
@jgrandguillaume 
I think this is a serious problem. What do you think?

PR #152, opened 2016-09-12
Don't write company_id on counterpart move line as it is a related on…

yvaucher:
… account and we have no reason to change company on account


florian-dacosta:
I have included this fix here : https://github.com/OCA/bank-statement-reconcile/pull/149
I would be great to review/merge it instead

guewen (approved):

PR #157, opened 2016-10-06
Force import partner on account.move

mdietrichc2c:
Small PR to fix an issue I encountered with the import : since multiple partners are now in the move, this means that editing one move line from an import will change the move's partner to a random partner present.

With this, the partner defined as the bank's partner will remain the move's partner.


florian-dacosta:
:+1:

pedrobaeza (approved):

PR #158, opened 2016-10-10
FIX account_reconcile_payment_order: move lines preparation for bank statement reconciliation

eLBati:
Without this, in multi currency environment, the amount shown in reconciliation widget would be incoherent:
base currency amount and foreign currency symbol would be mixed


pedrobaeza (approved):

pedrobaeza:
Merging this old PR as it's correct

PR #161, opened 2016-10-19
Remove now non-existing "Display partial" button

mdietrichc2c:
Small part I missed when migrating the modules : the "Display partial" button should not exist anymore.


florian-dacosta:
:+1:

pedrobaeza (approved):
We can fast-track this as it's obvious.

PR #168, opened 2017-03-02
[9.0]Fix balance for account_move_base_import

florian-dacosta:
Since the orm is by passed I found 2 problems : 
- computed fields (like field balance) are not computed
- automatic field (create_date...) are empty

So I just add automatic field and I call the method to recompute the balance field.
Still, it does not fix the problem for other fields. I am not sure of what to do about it...
Is there a way to recompute all computed fields of a model? Also, maybe it would slow the import a lot...
Any ideas?

### DIFF
--- a/account_move_base_import/models/account_journal.py
+++ b/account_move_base_import/models/account_journal.py
@@ -7,6 +7,7 @@
 import sys
 import traceback
 import os
+from datetime import datetime
 from openerp import _, api, fields, models
 from ..parser.parser import new_move_parser
 from openerp.exceptions import UserError, ValidationError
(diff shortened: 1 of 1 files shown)


florian-dacosta:
@mdietrichc2c what do you think?

mdietrichc2c:
@florian-dacosta that we cannot recompute all fields on the move line (or otherwise we'd just use the ORM for importing). I'm not against case-by-case recomputes, but for me, the create_date/write_date are not necessary.

florian-dacosta:
Yes, that what I thought about compute field.
But this one (balance) is a really important one since all OCA fiancial report are based on it.

About the create_date/write_date, I do not think it is really time consuming and it is better to have it for consistency, I don't see why we should not add it?

PR #169, opened 2017-03-24
[9.0][BUG][account_mass_reconcile] Fix writeoff limit in case of partial reconciliation

florian-dacosta:
There is an important issue in case of advance reconciliation.
Indeed, for advance reconciliation, the system always make partial reconciliation, but when checking the write_off limit, we just consider the credit/debit and we do not check if a part of the journal entry has already been partially reconciled. 
So it could fail on some cases, I made a test to illustrate : 

write off limit : 0.1
Whe have 1 payment made day 1 : 500
One small invoice made day 1 : 45
=> It gets partially reconciled, so the invoice is paid/reconciled, and the paymant is partially reconciled

We got a bigger invoice made day 2 : 500
Before this PR, the system will find 2 journal entries with amount = 500, it will consider it is below the write off limit. But the payment has already been partially reconciled, so, the residual amount is 455, et this is not taken into account.
The, it will do the reconciliation and create a write off with amount = 45 ! We definitily don't want that.

To fix it, I check the write off limit with amount_residual field, instead of credit/debit fields.
I am not sure it is the best way to fix the bug, any better idea is appreciated!

@mdietrichc2c @fclementic2c @sebastienbeau

### DIFF
--- a/account_mass_reconcile/models/base_reconciliation.py
+++ b/account_mass_reconcile/models/base_reconciliation.py
@@ -52,6 +52,7 @@ class MassReconcileBase(models.AbstractModel):
             'id',
             'debit',
             'credit',
+            'amount_residual',
             'date',
             'ref',
             'name',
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow:
Does this also work well with multicurrency?

guewen (approved):
We use it in production since july on a large implementation. There is a lint error to fix though

PR #175, opened 2017-05-31
[9.0] Use correct order if cron only does one reconcile at a time

mdietrichc2c:
Pretty much a straight-forward PR: `sorted` was used on a set to get the oldest reconcile method in `run_scheduler`, but since it's not assigned, the set remained in the same order and only 1 method was ever used in the cron job.

### DIFF
--- a/account_mass_reconcile/models/mass_reconcile.py
+++ b/account_mass_reconcile/models/mass_reconcile.py
@@ -301,7 +301,7 @@ class AccountMassReconcile(models.Model):
         if run_all:
             reconciles.run_reconcile()
             return True
-        reconciles.sorted(key=_get_date)
-        older = reconciles[0]
+        ordered_reconciles = reconciles.sorted(key=_get_date)
+        older = ordered_reconciles[:1]
         older.run_reconcile()
         return True


account_mass_reconcile/models/mass_reconcile.py
```diff
@@ -301,7 +301,7 @@ def _get_date(reconcile):
         if run_all:
             reconciles.run_reconcile()
             return True
-        reconciles.sorted(key=_get_date)
-        older = reconciles[0]
+        ordered_reconciles = list(reconciles.sorted(key=_get_date))
```
pedrobaeza: Why do you make a list from the result instead using the recordset itself?

mdietrichc2c:
@pedrobaeza I did not know the RecordSet could be sorted ; it's now fixed.

account_mass_reconcile/models/mass_reconcile.py
```diff
@@ -301,7 +301,7 @@ def _get_date(reconcile):
         if run_all:
             reconciles.run_reconcile()
             return True
-        reconciles.sorted(key=_get_date)
-        older = reconciles[0]
+        ordered_reconciles = reconciles.sorted(key=_get_date)
+        older = ordered_reconciles[0]
```
pedrobaeza: You should put `older = order_reconciles[:1]` for avoiding error in case there's no reconciles.

mdietrichc2c:
@pedrobaeza thanks ; fixed as well.

pedrobaeza (approved):

guewen (approved):

PR #177, opened 2017-06-19
[9.0] correcty use temporary cursor in reconcile method

mdietrichc2c:
We had some concurrency issues with mass reconcile. After checking, here are my fixes:
* if a temporary new cursor is created, use it (with the context) for reconciliation (this way, we avoid releasing the lock on the cron),
* remove `ctx['commit_every']` from `_rec_auto_lines_advanced` (since it's defined in `run_reconcile`),
* add a `SELECT FOR UPDATE` to lock the current mass reconcile (useful to avoid processing the same account in 2 concurrent processes)


guewen:
@mdietrichc2c I fixed a small bug here: https://github.com/mdietrichc2c/bank-statement-reconcile/pull/5

guewen:
Replaced by #196

PR #183, opened 2017-09-13
[10.0] [IMP] base_transaction_id: Add transaction_id on account.payment

lmignon:
This change is required to ease the reconciliation of the account.move
created from the account.payment.
- [ ] required by https://github.com/OCA/sale-workflow/pull/526

### DIFF
--- a/base_transaction_id/models/account_payment.py
+++ b/base_transaction_id/models/account_payment.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class AccountPayment(models.Model):
+
+    _inherit = 'account.payment'
+
+    transaction_id = fields.Char(
+        'Transaction ID',
+        required=False,
+        copy=False,
+        help="Transaction id from the financial institute"
+    )
+
+    @api.multi
+    def _get_liquidity_move_line_vals(self, amount):
+        self.ensure_one()
+        vals = super(AccountPayment, self)._get_liquidity_move_line_vals(
+            amount)
+        if self.transaction_id:
+            vals['name'] = self.transaction_id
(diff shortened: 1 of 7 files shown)


dvdhinesh (approved):
Code review :+1:

base_transaction_id/models/account_payment.py
```diff
@@ -0,0 +1,26 @@
...
+        help="Transaction id from the financial institute"
+    )
+
+    @api.multi
+    def _get_liquidity_move_line_vals(self, amount):
+        self.ensure_one()
+        vals = super(AccountPayment, self)._get_liquidity_move_line_vals(
+            amount)
+        if self.transaction_id:
+            vals['name'] = self.transaction_id
```
adrienpeiffer, line 27: wouldn't it be better to fill transaction_ref (as is done in this module) ?

atchuthan (approved):
Code review :+1:

Cedric-Pigeon:
Travis is red because of another addon

Cedric-Pigeon:
@OCA/accounting-maintainers Does anyone can merge this one?

rousseldenis (approved):
Code review

base_transaction_id/models/account_payment.py
rousseldenis: transaction_id is actually the transaction reference ...

rousseldenis:
@sbidoul This one is ready and in production for years

sbidoul:
> @sbidoul This one is ready and in production for years

@rousseldenis @lmignon is it correct that we'll abandon this approach in 13 (and maybe even 12)? In which case it might be better to not merge it?

rousseldenis:
> > @sbidoul This one is ready and in production for years
> 
> @rousseldenis @lmignon is it correct that we'll abandon this approach in 13 (and maybe even 12)? In which case it might be better to not merge it?

Yes, but the module exists already in 10.0 branch. This is an enhancement/fix for the actual installs.

gurneyalex (approved):

Issue #188, opened 2017-10-30
[9.0] account_mass_reconcile_transaction_ref broken since recent commit


florian-dacosta:
The module won't work since this commit : 
https://github.com/OCA/account-reconcile/commit/91a1aafd849d7a34da8d13dab42409c6950cc14e

The problem is that now, base_mass_reconcile call the method `_base_columns` on `mass.reconcile.base` but account_mass_reconcile_transaction_ref override `_base_columns` on `mass.reconcile.advanced`.
So the code of account_mass_reconcile_transaction_ref is not used anymore.

I don't get why it has changed?
@mreficent @mdietrichc2c

PR #189, opened 2017-10-30
[9.0][FIX] issue with mass_reconcile

MiquelRForgeFlow:
Fix issue https://github.com/OCA/account-reconcile/issues/188.

cc: @jbeficent @florian-dacosta

### DIFF
--- a/account_mass_reconcile/models/base_reconciliation.py
+++ b/account_mass_reconcile/models/base_reconciliation.py
@@ -45,7 +45,6 @@ class MassReconcileBase(models.AbstractModel):
         """
         raise NotImplementedError
 
-    @api.model
     def _base_columns(self):
         """ Mandatory columns for move lines queries
         An extra column aliased as ``key`` should be defined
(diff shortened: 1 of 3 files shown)


florian-dacosta:
Thank you for looking into it!
I don't see how it fix it though, I think, we should replace `self.env["mass.reconcile.base"]._base_columns()` by `self._base_columns()` in account_mass_reconcile/models/base_reconciliation.py at line 68

LoisRForgeFlow (approved):
Tested, works ok :+1:

florian-dacosta (approved):
Thanks for the fix
I did test with account_mass_reconcile_transaction_ref module but not account_mass_reconcile_by_purchase_line

LoisRForgeFlow:
@florian-dacosta FYI my tests were with 'account_mass_reconcile_by_purchase_line' so I think everything works ok now :smile:

florian-dacosta:
Yes, That is what I thought :)

PR #192, opened 2017-12-11
[10.0] Add module for importing automatically stripe payout

sebastienbeau:
This module add the possibiltity to import stripe payout automatically or manually.

TODO

- [ ] Add test with real connection to stripe API


pedrobaeza:
Shouldn't this be in bank-statement-import?

sebastienbeau:
Hi @pedrobaeza 

Seen this repo have been rename, maybe we can move this module in "bank-statement-import" but if we want to move it we should also move the base module "account_move_base_import", this module may look like the module "account_bank_statement_import" but the big difference is that we import only one move with many line. This is perfect for importing cart deposit like in my case with stripe.

In order to avoid confusion with bank importing and deposit importing maybe it will be better to have a new repo called "move import" or "deposit import"?
What do you think?

pedrobaeza:
OK, I see. The repo was renamed due to other request that I don't remember who made it. A repository account-move-import makes sense, but if you don't want to delay more this, go on as now. Any way, `account_move_import_stripe` name follows a better convention and I would remove `base` part from `account_move_base_import`, but this is something more subjective.

sebastienbeau:
It's seem that the base module have been ported but it's not really used for now (the import was really broken see here : https://github.com/OCA/account-reconcile/pull/193)
So I think we can rename the base module and extract this module in a separated branch as for now nobody use it in production.
If you are ok, I will send a mail to the board in order to have a new repo

fcayre:
Hi @sebastienbeau I have a small patch to contribute to your PR but I can't find your original repository to make a PR to your PR. For your information however, stripe.BalanceTransaction.all is to be changed to stripe.BalanceTransaction.list (Stripe API has apparently changed).

sebastienbeau:
Hi @fcayre  The branch is in this repo : https://github.com/akretion/account-reconcile

okuryan:
@sebastienbeau Any plans to Odoo 12 migration?

sebastienbeau:
I thinks I will work on it in september (no customer interested yet to fund the migration), If you are interested to migrate it before, I can review it

PierrickBrun:
@okuryan I did the migration to Odoo 12 here #315

account_move_stripe_import/models/account_move.py
```diff
@@ -0,0 +1,46 @@
...
+        Same as get_from_transaction_id_and_so but uses the reference field
+        of the transaction instead of its ID.
+        """
+        res = {}
+        so_obj = self.env["sale.order"]
+        reference = line.name
+        if line.name.startswith("REFUND"):
+            # Substring between
+            reference = line.name[len("REFUND FOR CHARGE(")+1:-len(")")]
+        sales = so_obj.search([("transaction_ids.reference", "=", reference)])
```
regispirard: With Odoo10 Website E-commerce, when the customer choose a payment acquirer (for example, wire transfer) and then change his mind and choose another payment acquirer like stripe, then we end up with two transactions in the database. The second one's name is the sale order + "x1".

image

Also, we sometimes have the stripe transaction id in line.name (it is the case in belgium with stripe bancontact payments).

This would solve theses two cases : 
sales = so_obj.search(['|',
                              ("payment_tx_id.reference", "=like", reference+'%'),
                              ("payment_tx_id.acquirer_reference", "=", reference)])

account_move_stripe_import/models/account_move.py
```diff
@@ -0,0 +1,46 @@
...
+        partners = sales.mapped("partner_id")
+        if len(partners) > 1:
+            raise ErrorTooManyPartner(
+                _('Line named "%s" was matched by more than ' "one partner.")
+                % line.name
+            )
+        if len(partners) == 1:
+            res["partner_id"] = partners.id
+            res["account_id"] = partners.property_account_receivable_id.id
```
regispirard: I'm not sure we should change account_id to put the receivable account.

The invoice is accounted on receivable account and the payment (account.payment) linked to the invoice is reconciled with the invoice. Receivable account balance is then 0€.

I think we'd better let the account as it is when importing the statement (the account configured on account.journal.receivable_account_id).

account_move_stripe_import/models/account_journal.py
```diff
@@ -0,0 +1,121 @@
...
+        super(StripeParser, self).__init__(journal, *args, **kwargs)
+        self.env = journal.env
+
+    @classmethod
+    def parser_for(cls, parser_name):
+        return parser_name == 'stripe'
+
+    def _get_account(self):
+        return self.env['payment.acquirer'].search(
+            [('provider', '=', 'stripe')])
```
regispirard: Domain should be improved to support multi-company.
We should look for payment acquire provider "stripe" within the same company as the one of the journal for which we import pay-out.

    def _get_account(self):
        return self.env['payment.acquirer'].search(
            [('provider', '=', 'stripe'),
             ('company_id', '=', self.journal.company_id.id)])

account_move_stripe_import/models/account_move.py
regispirard: + the standard field on sale order is "payment_tx_id" instead of "transaction_ids".

regispirard (commented):
Hi, this module is really useful, thanks !
I added some comments and it would be great if it could be merged.

account_move_stripe_import/__manifest__.py
```diff
@@ -0,0 +1,32 @@
...
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "external_dependencies": {
+        "python": [],
+        "bin": [],
+    },
+    "depends": [
+        "base_transaction_id",
+        "account_move_base_import",
```
regispirard, line 22: Add "data/completion_rule_data.xml",

regispirard (changes requested):

account_move_stripe_import/models/account_move.py
PierrickBrun: I'm not sure to understand were to find the data from where we are here.

If you want you can do a PR against mine and I'll merge it

regispirard: Hi @PierrickBrun 

> I'm not sure to understand were to find the data from where we are here.
> 
> If you want you can do a PR against mine and I'll merge it

Is https://github.com/akretion/account-reconcile/pull/2 all good ?

PR #199, opened 2018-03-20
[IMP] assert that move is balanced, recompute amount total, and set reconciled to false to avoid having a null value instead

sebastienbeau:
As ORM is bypassed for perf reason, we need to
- call assert balanced method
- set the total amount of the move
-set reconciled field to false instead of null value


florian-dacosta (approved):
LGTM
Travis error seems unrelated.

florian-dacosta:
LGTM
Travis error seems unrelated.

sebastienbeau:
I have rebased my PR

sebastienbeau:
@lmignon @Cedric-Pigeon

PR #127, opened 2015-10-16
[9.0] [port] base_transaction_id

yvaucher:

PR #200, opened 2018-03-28
[10.0]base_transaction_id: reconciliation process should not be restricted to payable and receivable accounts

Cedric-Pigeon:
My customer wants to 2 reconcile 2 bank accounts.

- one account to pay the invoice or pos_order
- the other one wen he gets back the money for the payment platform

Whe he imports the bank statement, he would like to reconcile the first one with the second one.

### DIFF
--- a/base_transaction_id/models/account_bank_statement_line.py
+++ b/base_transaction_id/models/account_bank_statement_line.py
@@ -12,13 +12,9 @@ class AccountBankStatementLine(models.Model):
     def get_reconciliation_proposition(self, excluded_ids=None):
         """ Look for transaction_ref to give them as proposition move line """
         if self.name:
-            # If the transaction has no partner, look for match in payable and
-            # receivable account anyway
-            overlook_partner = not self.partner_id
             domain = [('transaction_ref', 'ilike', self.name)]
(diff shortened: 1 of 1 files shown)


Cedric-Pigeon:
@yvaucher Any opinion on that ? Thanks

yvaucher:
@Cedric-Pigeon This was added when I migrated the module to Odoo 9.0 [1]. I don't know how that module evolved since.

On a functional side, better to ask @fclementic2c my memory is blurry about it.

I don't remind why we wanted only payable or recevable in case there is no partner defined on the statement line.

And I'm not sure what implies your case with 2 bank accounts. (not done a lot of accounting dev lately)


[1] https://github.com/OCA/account-reconcile/pull/127/commits/3d5e954886d0bd076008d01807248aae1b8d76a2

fclementic2c:
I agree. Only reconcilable accounts make sense. I guess we restricted for performance reasons

Cedric-Pigeon:
@fclementic2c @yvaucher could you review/approve this one pls?

yvaucher (approved):
Not sure to what @fclementic2c agreed here, but code LGTM

PR #201, opened 2018-03-29
[FIX] fix default order for simple reconciliation

sebastienbeau:
Hi, this is a small fix that improve the result of automatic reconciliation.

Indeed in case you have a lot of existing move line unreconciled with the same amount, and you reconcile based on "amount and partner" the system can reconcile old invoice with new payment.
With that fix this problem is really reduce as the reconcilliation will be done on ordered line on the date.

@guewen


sebastienbeau:
Rebased

sebastienbeau:
@lmignon @Cedric-Pigeon

bodedra (approved):
LGTM :+1:

PR #204, opened 2018-04-05
fix red travis on 10.0 branch

damdam-s:
https://travis-ci.org/OCA/account-reconcile/jobs/351116736#L1106


damdam-s:
cc @gurneyalex

pedrobaeza:
Still red I'm afraid

damdam-s:
yep. currency installed by default seems to be USD now ...
tests needs to be adapted according

pedrobaeza:
We have fixed it in other places making the currency change by SQL.

pedrobaeza:
In this case the error is not that one, but using an account that forces to use a secondary currency. Maybe one of your colleagues can help you in fixing it.

PR #205, opened 2018-04-05
[11.0][MIG] bank_statement_foreign_currency

luismontalba:
Bank Statement Foreign Currency
-------------------------------------------
This module makes visible the currency columns in the bank statements lines.

cc @Tecnativa


fclementic2c:
@luismontalba 
By default, bank statement must be populated in the currency of the journal. 
And, if in foreign currency, odoo convert debits/credits using internal odoo exchange rate.

If you make visible these columns you give the impression that user can enter manually the amount in company currency...which will not work I guess.

pedrobaeza:
@fclementic2c I also don't understand too much the sense of this, but it's included in OCA's instance. Maybe @elicoidal or @jbeficent can say why.

elicoidal:
@pedrobaeza I see that indeed: I think originally this is due to Paypal account that we have with multi-currencies transactions.
Nevertheless I would tend to think similarly to @fclementic2c here: if multiple currencies balances in Paypal then we should have multiple journal and handle the transfers in inter-account transfer accounts, managing the currency rate during the transfer. @fclementic2c @jbeficent what do you think

JordiBForgeFlow:
@fclementic2c I have been using this feature for years, and it is not confusing to me. When the bank provides in their statement transactions in a foreign currency, you enter it. Notice that all banks will provide the amounts in the account's main currency and in the foreign currency. You are not supposed to rely on Odoo's currency rate of the moment. The exchange rate is given by the bank.

For example, I buy a widget to Amazon.com for $100.

I can see in my statement
Amazon.com -80€ (-$100)

Then I enter exactly this in Odoo's bank statement. The rate 80€/$100 is the rate that the bank applied to a payment I made in USD in Amazon.com, because my bank account is held in EUR. I need the $100 to be able to reconcile with the invoice. I need the €80 because that's what hit my bank.

pedrobaeza:
OK, thanks for the info to both. @luismontalba I think we can put @jbeficent's example in the README for clarifying the purpose of the module.

pedrobaeza (approved):

fclementic2c:
@jbeficent 
Well I am not a bit fan of this way of posting amount in currency. I had some troubles in the past with this because you can do it with bank statement but not with account.payment or write-off entries (you have to use odoo internal rate)

More generally:
You can use the currency amounts given by your bank if it is provided for all transactions or and use the Odoo internal rates (as long as it is a reliable source - like the ECB - and always the same source (if you change the currency calculation method you must specify it (in the balance sheet notes) and measure the impact  on the P&L for the auditors/tax office in theory).

More or less you do what you want... until you do tax forms or the year-end closing.
So most important is to reevaluate your curr. balances at closing date based on official rate requested (ie: in Switzerland, it is the average rate of the last 30 days before fy closing). This is why we did the module https://github.com/OCA/account-closing/tree/11.0/account_multicurrency_revaluation

I never advise my customers to use the bank currency amounts because:
- it takes more time to enter all these amounts (and it is not mandatory)
- plus import is not always available in currency
- this is not the usual practice in bigger companies
- and you cannot remain consistent with this method everywhere in odoo (ie:  write offs, payments from invoices). At some point odoo will use the internal rate and you will have no chance to populate the equivalent amount in company currency.

JordiBForgeFlow:
@fclementic2c Trying to visualize your point of view here.

I created an example in http://3322150-11-0-c36b58.runbot2.odoo-community.org/web

Customer invoice INV/2018/0006  for 100€
Paid 120€ through the bank. 
image

In reconcile you have to manually write off to the foreign exchange gain/loss account:
image

Is that correct?

The benefit I can see of indicating the foreign currency is that if makes it easy to match the payments with invoices / vouchers, as the amount in foreign currency will match.

We are assuming in this scenario that your bank currency is not in foreign currency.

I understand that you will use https://github.com/OCA/account-closing/tree/11.0/account_multicurrency_revaluation to revaluate both bank accounts that you hold in foreign currency, as well as open receivables and payables that you hold in foreign currency, right?

luismontalba:
I've added the example proposed in the README, for better explanation of the module utility.

okuryan:
@fclementic2c I actually totally agree with @jbeficent that this is absolutely needed module. It will be a nightmare to reconcile with bank statement with payments in mixed currencies. All write-off values have to be entered manually and that delays the process really. 

I want to understand better reasons on why this module should not be here? We use this feature for a long time since Odoo 8 (in Odoo 8 those fields where not hidden) and having no issues.

JordiBForgeFlow (approved):

PR #213, opened 2018-07-25
[10.0][FIX] account mass reconcile: deprecated icon on some buttons

MiquelRForgeFlow:
Old icons from v9 don't exist in v10. Thus, they should be changed by new ones.


AaronHForgeFlow (approved):
:+1: Yes, I noticed that

JordiBForgeFlow (approved):

PR #224, opened 2018-10-26
[9.0][imp][account_mass_reconcile] search for unreconciled entries using a search default method, instead of a domain. 

JordiBForgeFlow:
When the user presses the button 'Go to unreconciled items'

image

A method would generate a domain that contained the move lines pending to reconcile for the account being reconciled. 

We found that when the number of items to reconcile was large (e.g. more than 6000 items), the resulting tree view was causing problems, when applying groupings to the resulting list.

I find it much more solid to apply search default filters based on account and unreconciled journal items.

image

This is anyway how you'd search for unreconciled entries for a given account in the journal items menu.

### DIFF
--- a/account_mass_reconcile/models/mass_reconcile.py
+++ b/account_mass_reconcile/models/mass_reconcile.py
@@ -251,7 +251,7 @@ class AccountMassReconcile(models.Model):
         )
 
     @api.model
-    def _open_move_line_list(self, move_line_ids, name):
+    def _open_move_line_list(self, context, name):
         return {
             'name': name,
             'view_mode': 'tree,form',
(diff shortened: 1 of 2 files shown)


JordiBForgeFlow:
@sebastienbeau @guewen @gurneyalex @pedrobaeza @yvaucher @damdam-s @florian-dacosta As contributors to this module, do you agree to implement this change? It might have an impact because it changes the signature of a method.

JordiBForgeFlow:
@pedrobaeza what do you think of this one?

pedrobaeza (approved):

yvaucher (approved):
I'm not aware of any module for which the signature of that private method would be an issue.

LGTM

PR #226, opened 2018-10-31
[10.0][imp][account_mass_reconcile] search for unreconciled entries using a search default method, instead of a domain. Fixes memory errors.

JordiBForgeFlow:
Forward port of https://github.com/OCA/account-reconcile/pull/224

cc @yvaucher @pedrobaeza


MiquelRForgeFlow:
title should say v10

MiquelRForgeFlow:
travis red

PR #227, opened 2018-10-31
[11.0][imp][account_mass_reconcile] search for unreconciled entries using a search default method, instead of a domain. Fixes memory errors.

JordiBForgeFlow:
Forward port of https://github.com/OCA/account-reconcile/pull/224

cc @pedrobaeza @yvaucher


MiquelRForgeFlow:
heads...

JordiBForgeFlow:
green!

grindtildeath (approved):
cc @fclementic2c Seems just what we talked about this morning :wink:

PR #238, opened 2019-02-12
[10.0] Rollback reconcile operations when automatic reconcile fails.

ecino:
Not sure about this one, review would be good. But in our case we have issues when automatic reconcile jobs are failing. Some invoices are left as paid because it was reconciling its move entries during the process. However, because the job didn't finish, the move lines are not full reconciled and we can reconcile again which looks wrong. With this added rollback, the things go in previous state and we don't end up with partial reconciles and having to manually correct it (unreconcile / reconcile properly).


ecino:
The downside of this is that all reconciles that were done in a job are lost when an error occurs. A solution would be to move the try/except block at the line levels and commit after each line reconciled successfully.

Any opinions?

ecino:
@mreficent @mdietrichc2c @damdam-s

PR #239, opened 2019-02-20
[MIG][12.0] base_transaction_id: migration

i-vyshnevska:
migrate to 12.0


base_transaction_id/models/account_bank_statement_line.py
```diff
@@ -21,5 +23,5 @@ def get_reconciliation_proposition(self, excluded_ids=None):
                 overlook_partner=overlook_partner)
             if match_recs and len(match_recs) == 1:
                 return match_recs
-        _super = super(AccountBankStatementLine, self)
```
simahawk: as it is shorter, you don't need a var anymore

base_transaction_id/models/account_move.py
```diff
@@ -35,7 +37,7 @@ def prepare_move_lines_for_reconciliation_widget(self,
     @api.model
     def domain_move_lines_for_reconciliation(self, str=False):
         """Add transaction_ref in search of move lines."""
-        _super = super(AccountMoveLine, self)
```
simahawk: same

simahawk (approved):

Tonow-c2c (approved):
LGTM

Proposition : 
> It's not in the OCA recommendation but they is not a good idea for the migration to : 
> 1) duplicate the module to migrate to the expected version (eg: duplicate module_x from branch 11 to 12)
> 2) doing the migration to the right version (eg: doing the change for migrated module_x on 12 in the branche 12)
> Like this we have one exploitable diff in the first real commit

pedrobaeza:
@Tonow-c2c you can see easily the diff for the relevant commits thanks to GitHub itself:
seleccion_016
You can use Shift key also for selecting more than 1 commit to see the diff.

PR #243, opened 2019-03-06
[10.0][IMP]account_set_reconcilable

AaronHForgeFlow:
Calculate amount residual also when setting the account as not reconcilable. Otherwise it will be waiting for reconciling in the dashboard, causing confusion.

Ping @mreficent @jbeficent


MiquelRForgeFlow (approved):

JordiBForgeFlow (approved):

sergio-teruel (approved):
¿Can you do a cherrypick to v 11.0?

PR #252, opened 2019-05-02
[IMP] account_move_base_import: don't break without xlrd, add it to dependencies

hbrunn:
otherwise, this breaks when Odoo loads assets (we have a static folder by now) and the library isn't available

### DIFF
--- a/account_move_base_import/__manifest__.py
+++ b/account_move_base_import/__manifest__.py
@@ -26,6 +26,9 @@
         'test/refund.yml',
         'test/completion_test.yml'
     ],
+    'external_dependencies': {
+        'python' : ['xlrd'],
+    },
     'installable': True,
     'license': 'AGPL-3',
 }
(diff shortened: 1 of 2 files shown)


grindtildeath:
@hbrunn How is this needed since `xlrd` is in https://github.com/odoo/odoo/blob/10.0/requirements.txt#L42 ?

hbrunn:
https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#importerror

grindtildeath:
@hbrunn Then, let's get this merged with https://github.com/hbrunn/bank-statement-reconcile/pull/2

hbrunn:
thanks

grindtildeath (approved):

sebalix (approved):
LG, however I don't know what you want to do about the fixup+merge in the history while preserving authorship, no problem to keep it like this? @hbrunn @grindtildeath

grindtildeath:
@sebalix I don't really care about ownership for such a small change :wink:  (i.e I merged both commits in #269 )

i-vyshnevska (approved):

PR #253, opened 2019-05-03
[10.0] Add account_mass_reconcile_partner

grindtildeath:
This module extends the functionality of account_mass_reconcile and adds a new
reconciliation method Advanced. Partner.
With this method credit lines are matched with debit lines in a chronological
order if the partner is the same.

### DIFF
--- a/account_mass_reconcile_partner/models/__init__.py
+++ b/account_mass_reconcile_partner/models/__init__.py
@@ -1,0 +1,2 @@
+from . import advanced_reconciliation
+from . import mass_reconcile
--- a/account_mass_reconcile_partner/models/mass_reconcile.py
+++ b/account_mass_reconcile_partner/models/mass_reconcile.py
@@ -1,0 +1,17 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import models, api
+
+
+class AccountMassReconcileMethod(models.Model):
+    _inherit = 'account.mass.reconcile.method'
+
+    @api.model
+    def _get_all_rec_method(self):
+        res = super(AccountMassReconcileMethod, self)._get_all_rec_method()
+        res.append((
+            'mass.reconcile.advanced.partner',
+            'Advanced. Partner'
+        ))
+        return res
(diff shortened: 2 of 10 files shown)


vrenaville (approved):

account_mass_reconcile_partner/models/mass_reconcile.py
```diff
@@ -0,0 +1,17 @@
...
+
+class AccountMassReconcileMethod(models.Model):
+    _inherit = 'account.mass.reconcile.method'
+
+    @api.model
+    def _get_all_rec_method(self):
+        res = super(AccountMassReconcileMethod, self)._get_all_rec_method()
+        res.append((
+            'mass.reconcile.advanced.partner',
+            'Advanced. Partner'
```
simahawk, line 15: "Advanced. Partner" correct w/ that dot?

account_mass_reconcile_partner/tests/test_reconcile_partner.py
```diff
@@ -0,0 +1,113 @@
...
+import time
+
+from odoo.tests import SavepointCase
+
+
+class TestAccountReconcilePartnerChrono(SavepointCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super(TestAccountReconcilePartnerChrono, cls).setUpClass()
```
simahawk: pls disable tracking suite-wise if you don't need it

sebalix (approved):

simahawk (approved):
LG, non blocking things to address if you can.

i-vyshnevska (approved):

account_mass_reconcile_partner/models/mass_reconcile.py
grindtildeath: Yes, it's to stay consistent with the existing : https://github.com/OCA/account-reconcile/blob/11.0/account_mass_reconcile/models/mass_reconcile.py#L81

account_mass_reconcile_partner/models/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import advanced_reconciliation
```
umiphos, line 1: Can you add the License?

example:
https://github.com/OCA/maintainer-tools/blob/master/template/module/models/__init__.py

sebalix: It's not required for __init__.py where there is only import statements.

guewen: I don't feel that we need a license on import statements.

simahawk: same here :)

fclementic2c:
Functional test ok :+1:

fclementic2c (approved):

PR #254, opened 2019-05-06
[10.0] account_mass_reconcile: Fix display of partially reconciled items

grindtildeath:
Before this commit, only fully reconciled items are stored in the history.

This commit adds a new action to display partially reconciled items and renames the existing action to display fully reconciled items.

### DIFF
--- a/account_mass_reconcile/models/mass_reconcile.py
+++ b/account_mass_reconcile/models/mass_reconcile.py
@@ -159,7 +159,7 @@ class AccountMassReconcile(models.Model):
 
     @api.multi
     def run_reconcile(self):
-        def find_reconcile_ids(fieldname, move_line_ids):
+        def find_full_reconcile_ids(fieldname, move_line_ids):
             if not move_line_ids:
                 return []
             sql = ("SELECT DISTINCT " + fieldname +
@@ -169,6 +169,17 @@ class AccountMassReconcile(models.Model):
             self.env.cr.execute(sql, (tuple(move_line_ids),))
             res = self.env.cr.fetchall()
             return [row[0] for row in res]
+
+        def find_partial_reconcile_ids(move_line_ids):
+            if not move_line_ids:
+                return []
+            reconciles = self.env['account.partial.reconcile'].search([
+                '|',
+                ('credit_move_id', 'in', move_line_ids),
+                ('debit_move_id', 'in', move_line_ids),
+                ('full_reconcile_id', '=', False),
+            ])
+            return reconciles.ids
 
         # we use a new cursor to be able to commit the reconciliation
         # often. We have to create it here and not later to avoid problems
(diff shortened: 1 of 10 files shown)


guewen:
Maybe I'm wrong, but I have some memory of having this feature, and as the partial reconciliations are transient by nature, it was very confusing, not really a history since they change.

grindtildeath:
@guewen `account.partial.reconcile` is not a TransientModel ! 
From what I tested, any reconciliation will create such a record to map debit and credit move lines. The record will then be updated if the reconciliation is fully done by adding the `full_reconcile_id`, but they won't be deleted. Or maybe I miss a use case ?

guewen:
It was not a TransientModel neither before, but the `account.partial.reconcile` records would be deleted and created on every run, which means the "references" would be lost and the history would change over time. It's maybe no longer the case though.

edit: did exist in 7.0 and 8.0: https://github.com/OCA/account-reconcile/blob/6f80d8983b622b078b9478413e967b70132b1854/account_easy_reconcile/easy_reconcile.py#L325, removed in 9.0

account_mass_reconcile/models/mass_reconcile_history.py
```diff
@@ -66,7 +71,33 @@ def _open_move_lines(self):
...
+
+    @api.multi
+    def _open_partial_move_lines(self):
+        """ For an history record, open the view of move line with
+        the reconciled move lines
+
+        :param history_id: id of the history
+        :return: action to open the move lines
+        """
+        partial_credit_move_lines = self.partial_reconcile_ids.mapped(
```
TDu, line 85: Also need an ensure_one ?

account_mass_reconcile/models/mass_reconcile.py
```diff
@@ -280,6 +297,15 @@ def last_history_reconcile(self):
             self._no_history()
         return self.last_history.open_reconcile()
 
+    @api.multi
+    def last_history_partial_reconcile(self):
+        """ Get the last history record for this reconciliation profile
+        and return the action which opens move lines reconciled
+        """
+        if not self.last_history:
```
TDu, line 305: Probably needs an api.ensure_one ?

TDu (commented):
See my couple of comments, otherwise looks good but I am a little out of context.

account_mass_reconcile/models/mass_reconcile_history.py
grindtildeath: There's already one before calling the action here: https://github.com/OCA/account-reconcile/pull/254/files#diff-8ef76c9ad50a6f8b7e4be265e0f827b3R117
I don't think it's worth it to have two separated functions, but as we don't want to break possible dependencies, let's keep it as it is.

simahawk (approved):
eyeball core review. LG

account_mass_reconcile/models/mass_reconcile_history.py
TDu: Ok

i-vyshnevska (approved):

PR #257, opened 2019-05-15
[12.0][add][bank_statement_line_attachment]

JordiBForgeFlow:
Bank Statement Line Attachment
==============================

This module allows to add attachments to bank statement lines, if it has
been deemed as necessary.

It is common that some of the bank statement lines will be reconciled to
an expense account during the reconciliation process. That could be more
convenient to the company than having to enter invoices or vouchers. In that
scenario there's currently no place to store an evidence of the physical
receipts associated to that expense.


bank_statement_line_attachment/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
```
i-vyshnevska, line 1: as you introducing this addon rigths should be the same on all files?

bank_statement_line_attachment/models/account_bank_statement.py
```diff
@@ -0,0 +1,44 @@
...
+    attachment_number = fields.Integer('Attachments',
+                                       compute='_compute_attachment_number')
+    requires_attachment = fields.Boolean('Requires attachment',
+                                         default=False)
+
+    @api.multi
+    def _compute_attachment_number(self):
+        attachment_data = self.env['ir.attachment'].read_group(
+            [('res_model', '=', 'account.bank.statement.line'),
+             ('res_id', 'in', self.ids)], ['res_id'], ['res_id'])
```
i-vyshnevska, line 20:
```suggestion
             ('res_id', 'in', self.ids)], ['res_id'], ['res_id:sum'])
```
it will count on sql level so you don't need to itereate again on python level

bank_statement_line_attachment/models/account_bank_statement.py
```diff
@@ -0,0 +1,44 @@
...
+                          'default_name': self.name,
+                          }
+        return res
+
+    @api.multi
+    def unlink(self):
+        self.env['ir.attachment'].search(
+            [('res_model', '=', 'account.bank.statement.line'),
+             ('res_id', 'in', self.ids)]).unlink()
+        return super(AccountBankStatementLine, self).unlink()
```
i-vyshnevska, line 44:
```suggestion
        return super().unlink()
```

bank_statement_line_attachment/models/account_bank_statement.py
```diff
@@ -0,0 +1,44 @@
...
+
+from odoo import api, fields, models
+
+
+class AccountBankStatementLine(models.Model):
+    _inherit = 'account.bank.statement.line'
+
+    attachment_number = fields.Integer('Attachments',
+                                       compute='_compute_attachment_number')
+    requires_attachment = fields.Boolean('Requires attachment',
```
i-vyshnevska, line 13: seems it might be defined on related journal or bank account to facilitate user work

PR #265, opened 2019-06-15
[FIX][10.0] account_move_base_import currency handling

gurneyalex:
When importing a batch for a journal which has a currency different from the
company currency, the values in the batch files are in the currency of the journal.
The previous code was creating move lines as if the amounts in the batch file were in
the company currency.

We fix this by checking for the case, and converting the amounts the company currency at
the rate of the date of the transaction to fill in the credit / debit columns, and setting the
value of amount_currency to the value in the batch file.


i-vyshnevska (approved):

damdam-s (approved):
LGTM

Tonow-c2c (approved):
Just code review

PR #269, opened 2019-07-10
[11.0][MIG] account_move_base_import

grindtildeath:
Migration to v11.0

Included https://github.com/OCA/account-reconcile/pull/252

PR #270, opened 2019-07-15
[12.0][MIG] account_move_base_import

grindtildeath:
Based on:
 - [ ] (11.0) #269


account_move_base_import/models/account_journal.py
```diff
@@ -0,0 +1,384 @@
...
+
+    split_counterpart = fields.Boolean(
+        string="Split Counterpart",
+        help="Two counterparts will be automatically created : one for "
+             "the refunds and one for the payments")
+
+    def _get_rules(self):
+        # We need to respect the sequence order
+        return sorted(self.rule_ids, key=attrgetter('sequence'))
+
```
i-vyshnevska: @api.model
def _find_values_from_rules

account_move_base_import/models/account_journal.py
```diff
@@ -0,0 +1,384 @@
...
+        if amount > 0.0:
+            account_id = self.default_debit_account_id.id
+            credit = 0.0
+            debit = amount
+        else:
+            account_id = self.default_credit_account_id.id
+            credit = -amount
+            debit = 0.0
+        counterpart_values = {
+            'name': _('/'),
```
i-vyshnevska: AFAIR names can be generated based on journal so empty name should be handled properly

account_move_base_import/models/account_journal.py
```diff
@@ -0,0 +1,384 @@
...
+        transfer_line_count = len(transfer_lines)
+        check_move_validity = False
+        for amount in transfer_lines:
+            transfer_line_count -= 1
+            if not transfer_line_count:
+                check_move_validity = True
+            vals = self._prepare_counterpart_line(move, amount,
+                                                  counterpart_date)
+            move_line_obj.with_context(
+                check_move_validity=check_move_validity
```
i-vyshnevska, line 128: seems  check_move_validity=True will just execute _post_validate on lines, so we can perform this explicitly  and rid of transfer_line_count

account_move_base_import/parser/__init__.py
```diff
@@ -0,0 +1,9 @@
+# © 2011 Akretion
+# © 2011-2016 Camptocamp SA
```
i-vyshnevska:
```suggestion
# © 2011-2019 Camptocamp SA
```

account_move_base_import/models/partner.py
```diff
@@ -0,0 +1,21 @@
+# © 2011-2016 Akretion
+# © 2011-2016 Camptocamp SA
```
i-vyshnevska:
```suggestion
# © 2011-2019 Camptocamp SA
```

account_move_base_import/models/account_move.py
```diff
@@ -0,0 +1,347 @@
...
+    import_partner_id = fields.Many2one('res.partner',
+                                        string="Partner from import")
+
+    @api.depends('line_ids.partner_id', 'import_partner_id')
+    def _compute_partner_id(self):
+        for move in self:
+            if move.import_partner_id:
+                move.partner_id = move.import_partner_id
+            elif move.line_ids:
+                move.partner_id = move.line_ids[0].partner_id
```
i-vyshnevska: doesn't this change default behavior of odoo? move can have different partners in this case we should provide same check as https://github.com/odoo/odoo/blob/1bd434da34731b5467cc17fdffe6f3ed5dfea2d4/addons/account/models/account_move.py#L77

account_move_base_import/parser/generic_file_parser.py
```diff
@@ -0,0 +1,78 @@
+# © 2011 Akretion
+# © 2011-2016 Camptocamp SA
```
i-vyshnevska:
```suggestion
# © 2011-2019 Camptocamp SA
```

i-vyshnevska: and everywhere

account_move_base_import/models/account_journal.py
grindtildeath: I'm not sure to understand what you mean here, but I'd prefer not to change what has been working in previous versions.

account_move_base_import/models/account_journal.py
grindtildeath: ~`_find_values_from_rules` is actually `api.multi`~

EDIT: IMO this must be moved to `account.move.line`, I will rework it...
EDIT 2: Refactored in bd72af0

account_move_base_import/models/account_journal.py
i-vyshnevska: if I read this correctly lines 120-130 can be replaced: 

        for amount in transfer_lines:
            vals = self._prepare_counterpart_line(move, amount,
                                                  counterpart_date)
            lines += move_line_obj.with_context(
                check_move_validity=False
            ).create(vals)
         move._post_validate()

account_move_base_import/parser/generic_file_parser.py
sebalix: I wonder if we can not just put `2011` instead of a year-range. I think it implies this year until today.

account_move_base_import/models/account_move.py
```diff
@@ -273,9 +274,34 @@ def _prepare_insert(self, move, cols):
         Return a copy of statement
         """
         move_copy = move
-        for k, col in move_copy.iteritems():
+        for k, col in move_copy.items():
             if k in cols:
-                move_copy[k] = self._fields[k].convert_to_column(col, None)
+                # In v11, the record is needed to call convert_to_column on a
+                # Monetary field because it must find the currency to apply the
+                # proper rounding. So we mimic the call here as best as we can.
```
sebalix: The same trick than the v11 PR could be applied I suppose.

grindtildeath: This is outdated, was required only for v11.0 because I kept using direct SQL and v12.0 now uses ORM.

account_move_base_import/models/account_journal.py
grindtildeath: `check_move_validate` is only used on the last iteration (i.e last item to be created from `transfer_lines`, so this wouldn't do exactly the same... :man_shrugging:

i-vyshnevska (approved):

TDu (approved):

sebalix (approved):

florian-dacosta:
Thanks a lot for the migration of this module @grindtildeath 
One question though, you did remove the move line creation by sql.
Is it possible to know why? The performances issues are suddenly resolved? You did some tests to compare? Is it possible to know more about this?

grindtildeath:
@florian-dacosta after the refactoring, I guessed we could get rid of the UPDATE statement since it involves extra processing for something that could be written directly on the recordset and makes the module more simple. Then I thought we might just do the same for the INSERT since `api.create_multi` decorator exists in v12.0 (https://github.com/odoo/odoo/blob/12.0/addons/account/models/account_move.py#L1143), and we make use of it here: https://github.com/OCA/account-reconcile/pull/270/files#diff-56235e7a139662a882b141f3a618b84bR321
I didn't benchmark the performance though, but IMO we could still revert to direct SQL if it's really an issue importing lots of records.

florian-dacosta:
@grindtildeath Well, ok, We'll see if we got performance issues or not in version 12.

I made a test with a small file, and I got this warning : 
2019-07-28 15:16:40,849 2071 WARNING db odoo.models: account.move.line.create() with unknown fields: balance_cash_basic, credit_cash_basic, debit_cash_basic

I have no idea what these fields are or from which version they come from, but in v12, they do not exist and should be removed of the module.

yvaucher:
squashed rebased and merged at 03d527a

PR #275, opened 2019-07-21
[FIX] apropriate call _get_reconcilation_methods

i-vyshnevska:
fix call to parent method, fix forgotten test import

### DIFF
--- a/account_mass_reconcile_transaction_ref/tests/test_account_reconcile_transaction_ref.py
+++ b/account_mass_reconcile_transaction_ref/tests/test_account_reconcile_transaction_ref.py
@@ -72,7 +72,6 @@ class TestAccountReconcileTransactionRef(SavepointCase):
         })
         count = reconcile.unreconciled_count
         reconcile.run_reconcile()
-        reconcile._get_total_unrec()
         self.assertEqual(self.cust_invoice.state, 'paid')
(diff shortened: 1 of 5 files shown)


pedrobaeza:
It seems the test is not correct

account_mass_reconcile_transaction_ref/tests/test_account_reconcile_transaction_ref.py
```diff
@@ -72,6 +72,7 @@ def test_mass_reconcile_transaction_ref_vs_ref(self):
         })
         count = reconcile.unreconciled_count
         reconcile.run_reconcile()
+        reconcile._get_total_unrec()
```
grindtildeath: I'm not sure about this one, IMO it's an issue with the dependencies of `_get_total_unrec` : https://github.com/OCA/account-reconcile/blob/11.0/account_mass_reconcile/models/mass_reconcile.py#L119

Maybe we should add `history_ids` to make sure the field is recomputed after `run_reconcile` ?

i-vyshnevska: added

grindtildeath (approved):
LGTM :+1:

tschanzt (approved):

i-vyshnevska:
@yvaucher can we try merge this one again

PR #278, opened 2019-07-26
[10.0][ADD]account_bank_reconcile_hook

AaronHForgeFlow:
This allow to extend a domain when getting the account move lines for banking reconciliation.

PR #219, opened 2018-10-02
[11.0][ADD] account_skip_bank_reconciliation

MiquelRForgeFlow:
Account Skip Bank Reconciliation
================================

This module allows to exclude from bank statement reconciliation
all journal items of a specific reconcilable account.

Usually, you would want to that in accounts like the
_Goods Received Not Invoiced_, which are required to be reconcilable
to be able to have proper traceability in stock received but
their reconciliation is done using the `account_mass_reconcile` module.

PR #286, opened 2019-09-05
[10.0][ADD] account_exclude_from_bank_statement_reconciliation

AdriaGForgeFlow:
This module adds a flag to accounts that prevents them from being proposed
during bank statement reconciliation.

Depends on:

- [x]  #278 

cc ~ @jbeficent @Eficent


MiquelRForgeFlow:
Does it do the same as https://github.com/OCA/account-reconcile/pull/219? If yes, I think it's better if you do a backport. This way, module name and fields names would be the same.

AdriaGForgeFlow:
> Does it do the same as #219? If yes, I think it's better if you do a backport. This way, module name and fields names would be the same.

You're right, I will close this and create a new PR here #289

PR #287, opened 2019-09-19
[10.0][FIX] cleared balance

ovnicraft:
Cleared balance must include starting balance to avoid negative.
Balance is progressive in time


ovnicraft:
Travis fail is not related to this PR.

PR #291, opened 2019-10-01
[11.0] [ADD] account_reconciliation_widget_partial

etobella:
This addons allow to change the amount of a reconciliation line in order to
partially reconcile it.

Usage
=====

On reconciliations, an edit button is created.
Once pressed, you can modify the amount.
By default, it will propose the difference between the original amount and the
current balance if allowed.

@jarroyomorales @jbeficent 

reconciliation_partial


account_reconciliation_widget_partial/static/src/less/account_reconciliation.less
```diff
@@ -0,0 +1,25 @@
...
+        .strike_amount {
+            text-decoration: line-through;
+        }
+    }
+    .o_reconciliation_line .match{
+        .line_info_edit {
+            display: none!important;
+        }
+    }
+}
```
MiquelRForgeFlow: blank line

account_reconciliation_widget_partial/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2019 Creu Blanca
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Account Reconcile Partial',
```
MiquelRForgeFlow: update name

etobella:
@mreficent Comments attended

MiquelRForgeFlow (approved):
LGTM

You know, in v13 this module won't be needed 😜

etobella:
in V13 we will make the greatest reconciliation widget :stuck_out_tongue_closed_eyes:

AdriaGForgeFlow (approved):
Functional test and code review :+1:

PR #292, opened 2019-10-09
[12.0] base_transaction_id: Fix migration

grindtildeath:
Module is broken since its migration as the hooks used for reconciliation
are not used anymore in Odoo 12.0.

Therefore transaction_ref is dropped from account.move.line and transaction_id
is now written into account.move.ref at invoice validation to allow the use
of standard Odoo mechanism in the reconciliation process.

As Odoo defined invoice_payment_ref on account.move in v13.0, avoiding extra
fields in v12.0 is already a step in the right direction.


vrenaville (approved):

damdam-s (approved):
LGTM

Tonow-c2c (approved):
Just code review

yvaucher (approved):

yvaucher:
Travis is failing as it impacts `account_move_transactionid_import` module for which fixes have been done in #293

PR #296, opened 2019-11-13
[FIX][12.0][account_move_transactionid_import] s/transaction_ref/ref

Tonow-c2c:
the field `transaction_ref` doesn't exist any more


grindtildeath (approved):

grindtildeath (changes requested):
Doesn't work as expected, let's check it out together when you're back.

grindtildeath:
@Tonow-c2c https://github.com/Tonow-c2c/account-reconcile/pull/1 :wink:

account_move_base_import/parser/file_parser.py
```diff
@@ -70,13 +72,24 @@ def _parse(self, *args, **kwargs):
...
+        if self.parsed_file is None:
+            if self.ftype == 'csv':
+                self.parsed_file = self._parse_csv()
+            else:
+                self.parsed_file = self._parse_xls()
+        if self.support_multi_moves:
+            if len(self.parsed_file) <= self.current_line:
+                return False
+            else:
+                print(self.current_line)
```
i-vyshnevska:
```suggestion
```

account_move_transactionid_import/parser/transactionid_file_parser.py
```diff
@@ -32,6 +32,7 @@ def __init__(self, profile, ftype='csv', extra_fields=None, header=None,
         super().__init__(
             profile, extra_fields=conversion_dict, ftype=ftype, header=header,
             **kwargs)
+        self.support_multi_moves = True
```
i-vyshnevska, line 45: should be optional?

i-vyshnevska: I don't really understand the message of one commit:
 "parser allows to generate 1 account.move for each line in the imported file"
account move should have at least 2 lines, and second is searched/created on statement reconciliation 
Lint need to be checked

grindtildeath: > should be optional?

Not when using transaction_id parser due to the change from #292 

> "(support_multi_moves) parser allows to generate 1 account.move for each line in the imported file"
account move should have at least 2 lines, and second is searched/created on statement reconciliation

The `support_multi_moves` flag on the parser allows to generate 1 account.move per line in the imported statement. However resulting account move (for one line) will always include at least two move lines (cf https://github.com/OCA/account-reconcile/blob/12.0/account_move_base_import/models/account_journal.py#L319-L324) and even 3 if there's a commission amount (cf https://github.com/OCA/account-reconcile/blob/12.0/account_move_base_import/models/account_journal.py#L325-L326).

@Tonow-c2c Could you fix the lint and commit @i-vyshnevska 's suggestion ?

account_move_base_import/models/account_journal.py
```diff
@@ -245,7 +245,6 @@ def prepare_move_line_vals(self, parser_vals, move):
             'debit_cash_basic': values['debit'],
             'credit_cash_basic': values['credit'],
             'balance_cash_basic': values['debit'] - values['credit'],
```
grindtildeath: @Tonow-c2c Could you please also remove these undefined fields ?

vrenaville (approved):

i-vyshnevska (approved):

grindtildeath (approved):

simahawk:
@yvaucher quick check and merge?

PR #307, opened 2020-03-07
[FIX] account_reconcile_reconciliation_date: Reconciliation Date Not Showing

osi-scampbell:
The reconciliation_date field is being set properly however the field is not showing even though it should be. This solution fixes that bug.


max3903 (approved):

patrickrwilson (approved):
LGTM

PR #311, opened 2020-04-14
[12.0] Create module account_reconcile_blue_lines_restriction

grindtildeath:
This module extends the reconciliation feature of Odoo in order to change
the blue lines appearance conditions in the reconciliation propositions.

When opening the reconciliation widget, Odoo can propose account move lines
linked to an account that is not marked to "Allow Reconciliation", if said
account is defined as "Default Debit Account" or "Default Credit Account" on
the journal from the bank statement line.

This module restricts those proposition to move lines linked to an account
that is marked as "Allow Reconciliation".

PR #316, opened 2020-05-11
[13.0][MIG] account_reconcile_blue_lines_restriction

grindtildeath:
Migrated from:
 - [ ] #311

### DIFF
--- a/account_reconcile_blue_lines_restriction/models/account_reconcile_model.py
+++ b/account_reconcile_blue_lines_restriction/models/account_reconcile_model.py
@@ -1,0 +1,39 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, models
+
+
+class AccountReconcileModel(models.Model):
+
+    _inherit = 'account.reconcile.model'
+
+    @api.multi
+    def _apply_conditions(self, query, params):
+        """Remove AMLs with account not allowed for reconciliation from
+           blue lines appearance conditions"""
+        new_query, new_params = super()._apply_conditions(query, params)
+        if self.rule_type == 'invoice_matching':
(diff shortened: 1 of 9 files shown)


Tonow-c2c (approved):
Add in the Roadmap to add tests

account_reconcile_blue_lines_restriction/models/account_reconcile_model.py
```diff
@@ -0,0 +1,38 @@
...
+                        OR
+                        aml.date > company.account_bank_reconciliation_start
+                    )
+                    -- This line restricts propositions to lines on
+                    -- accounts allowing reconciliation
+                    AND account.reconcile IS TRUE
+                    OR (
+                        -- black lines appearance conditions
+                        account.reconcile IS TRUE
+                        AND aml.reconciled IS FALSE
```
i-vyshnevska, line 34: this condition should already be in the query? I do not really understand how it should work

i-vyshnevska (approved):
LG, some questions

account_reconcile_blue_lines_restriction/models/account_reconcile_model.py
grindtildeath: Yes this part is already there, however it's inside this big `AND` condition: https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_reconcile_model.py#L575-L592
Therefore we must redefine here this whole condition to add `AND account.reconcile IS TRUE`.

I must still add tests though.

grindtildeath:
After doing some testing, we found out that this module was created to avoid the bug we discovered yesterday with @yvaucher : https://github.com/odoo/odoo/pull/51223

So we probably will not need it anymore once the fix is merged.

grindtildeath:
Module not needed anymore after Odoo fix got merged

PR #320, opened 2020-07-06
[13.0] [MIG] bank_statement_foreign_currency: Migration to v13.0

CarlosRoca13:
cc @Tecnativa TT23180

Please @Tardo  @ernestotejeda review it


ernestotejeda (commented):
This module seems not to be necessary in v13 since amount_currency and currency_id fields are not invisible in the view and have option = 'hidden' set. The only weird thing is the user can select those fields even though he/she doesn't belong to the access group group_multi_currency.

Peek 2020-07-06 12-23

pedrobaeza:
Finally not needed as now in core.

PR #322, opened 2020-07-08
[ADD][12.0] account_mass_reconcile_as_job

i-vyshnevska:
As mass reconcile can take a lot of time this addon adds possibility to delay execution with queue

### DIFF
--- a/account_mass_reconcile/models/base_advanced_reconciliation.py
+++ b/account_mass_reconcile/models/base_advanced_reconciliation.py
@@ -5,6 +5,7 @@
 import logging
 from itertools import product
 
+import odoo
 from odoo import models, api, sql_db
 from odoo.tools.translate import _
 
@@ -214,21 +215,60 @@ class MassReconcileAdvanced(models.AbstractModel):
         """
         return False
 
+    def _rec_group(self, reconcile_groups, lines_by_id):
+        reconciled_ids = []
+        for group_count, reconcile_group_ids \
+                in enumerate(reconcile_groups, start=1):
+
+            _logger.debug("Reconciling group %d/%d with ids %s",
+                          group_count, len(reconcile_groups),
+                          reconcile_group_ids)
+            group_lines = [lines_by_id[lid]
+                           for lid in reconcile_group_ids]
+            reconciled, full = self._reconcile_lines(group_lines,
+                                                     allow_partial=True)
+            if reconciled and full:
+                reconciled_ids += reconcile_group_ids
+        return reconciled_ids
+
+    def _rec_group_by_chunk(self, reconcile_groups, lines_by_id, chunk_size):
+        """Commit after each chunk
+
+        :param dict reconcile_grous: all groups to reconcile, will be splitted
+        by chunk
+        :param list lines_by_id: list of dict of move lines values,
+          the move lines we want to search for
+        :return: list of reconciled lines
+        """
+        reconciled_ids = []
+
+        _logger.info("Reconciling by chunk of %d", chunk_size)
+
+        # Copy and commit current transient model before creating a new cursor
+        # This is required to avoid CacheMiss when using data from `self`
+        # which is created during current transaction.
+        with odoo.api.Environment.manage():
+            with odoo.registry(self.env.cr.dbname).cursor() as new_cr:
+                new_env = api.Environment(new_cr, self.env.uid, self.env.context)
+                self_env = self.with_env(new_env)
+                rec = self_env.create(self.copy_data())
+        for i in range(0, len(reconcile_groups), chunk_size):
+            with odoo.api.Environment.manage():
+                with odoo.registry(self.env.cr.dbname).cursor() as new_cr:
+                    new_env = api.Environment(new_cr, self.env.uid, self.env.context)
+                    chunk = reconcile_groups[i:i + chunk_size]
+                    # Re-use the commited transient we just commited
+                    self_env = self.with_env(new_env).browse(rec.id)
+                    reconciled_ids += self_env._rec_group(chunk, lines_by_id)
+        return reconciled_ids
+
     @api.multi
     def _rec_auto_lines_advanced(self, credit_lines, debit_lines):
         """ Advanced reconciliation main loop """
         reconciled_ids = []
         for rec in self:
+            commit_every = rec.account_id.company_id.reconciliation_commit_every
             reconcile_groups = []
-            ctx = self.env.context.copy()
-            ctx['commit_every'] = (
-                rec.account_id.company_id.reconciliation_commit_every
-            )
-            if ctx['commit_every']:
-                new_cr = sql_db.db_connect(self.env.cr.dbname).cursor()
-            else:
-                new_cr = self.env.cr
-
             _logger.info("%d credit lines to reconcile", len(credit_lines))
             for idx, credit_line in enumerate(credit_lines, start=1):
                 if idx % 50 == 0:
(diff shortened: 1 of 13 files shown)


account_mass_reconcile_as_job/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2017 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
+{
+    'name': 'Account Mass Reconcile as Jobs',
+    'version': '12.0.0.1.0',
+    'category': 'Accounting',
+    'depends': [
+        'connector',
```
guewen: you don't need `connector`, only `queue_job`
```suggestion
        'queue_job',
```

(`queue_job` was extracted from `connector` in version 10.0)

oca_dependencies.txt
```diff
@@ -1,2 +1,3 @@
 account-financial-tools
 bank-payment
+connector
```
guewen:
```suggestion
queue
```

guewen (approved):

gurneyalex (approved):

account_mass_reconcile_as_job/models/mass_reconcile.py
```diff
@@ -0,0 +1,41 @@
...
+            'account.mass.reconcile.as.job', default=False
+        )
+        try:
+            as_job = ast.literal_eval(as_job) if as_job else False
+        except ValueError:
+            as_job = False
+
+        for rec in self:
+            if as_job and self.env.context.get('mass_reconcile_as_job', True):
+                self.with_delay().reconcile_as_job()
```
guewen:
```suggestion
                rec.with_delay().reconcile_as_job()
```

account_mass_reconcile_as_job/models/mass_reconcile.py
```diff
@@ -0,0 +1,41 @@
...
+                return True
+            else:
+                return super(AccountMassReconcile, rec).run_reconcile()
+
+    @job(default_channel='root.mass_reconcile')
+    def reconcile_as_job(self):
+        """Run reconciliation with appropriate context key to interapt new job
+        creation"""
```
guewen:
```suggestion
        """Run reconciliation on a single account"""
```

yvaucher:
@i-vyshnevska can you squash / rebase (Travis should look slighly better now)

account_mass_reconcile/models/base_advanced_reconciliation.py
```diff
@@ -224,6 +224,11 @@ def _rec_auto_lines_advanced(self, credit_lines, debit_lines):
             ctx['commit_every'] = (
                 rec.account_id.company_id.reconciliation_commit_every
             )
+            if ctx['commit_every']:
+                new_cr = sql_db.db_connect(self.env.cr.dbname).cursor()
```
yvaucher: :warning: the creation of the new cursor must be in the same loop where the cursor is closed.

Otherwise it will raise an error such as "Unable to use a closed cursor".

The cursor seems also badly used. What is the purpose of having a new_cr but not using it anywhere before commit?

yvaucher (changes requested):

account_mass_reconcile/models/base_advanced_reconciliation.py
i-vyshnevska: mentioned code is removed in next commit

account_mass_reconcile/models/mass_reconcile.py
```diff
@@ -166,7 +166,7 @@ def _last_history(self):
     last_history = fields.Many2one(
         'mass.reconcile.history',
         string='Last history', readonly=True,
-        compute='_last_history',
+        compute='_compute_last_history',
```
grindtildeath, line 175: These changes of method names should be avoided in published versions as it could break existing installations.

At the very least, `_last_history = _compute_last_history` should be added somewhere for compatibility if the change is needed.

account_mass_reconcile/models/base_advanced_reconciliation.py
```diff
@@ -214,16 +215,60 @@ def _skip_line(self, move_line):
...
+        _logger.info("Reconciling by chunk of %d", chunk_size)
+
+        # Copy and commit current transient model before creating a new cursor
+        # This is required to avoid CacheMiss when using data from `self`
+        # which is created during current transaction.
+        with odoo.api.Environment.manage():
+            with odoo.registry(self.env.cr.dbname).cursor() as new_cr:
+                new_env = api.Environment(new_cr, self.env.uid, self.env.context)
+                self_env = self.with_env(new_env)
+                rec = self_env.create(self.copy_data())
```
grindtildeath, line 254: A bit of an ugly solution, but if it works and avoids the bare `cr.commit`, it's still less ugly than the actual code. :+1: 

Maybe add a TODO to refactor this so we don't have to create and persist actual wizard but use the result of `copy_data` with a new instead of a `browse` below. :question:

account_mass_reconcile/models/mass_reconcile_history.py
```diff
@@ -42,7 +42,7 @@ def _get_reconcile_line_ids(self):
         comodel_name='account.move.line',
         relation='account_move_line_history_rel',
         string='Reconciled Items',
-        compute='_get_reconcile_line_ids',
+        compute='_compute_reconcile_line_ids',
```
grindtildeath, line 48: Same as above

account_mass_reconcile/models/base_advanced_reconciliation.py
yvaucher: The wizard will be garbage collected anyway. I doubt this wizard will be called enough time per day even with a cron to cause an issue in term of DB size. For version 12.0 we won't invest time on refactoring the whole mass reconcile process, it would mean messing with the method signatures to use staticmethods instead. I would prefer to keep the code stable.

gurneyalex:
@yvaucher @grindtildeath  what is the current status of this PR? I know some version of this code is running in production.Can we merge this PR?

grindtildeath:
@gurneyalex From my POV it could be merged if the renaming of compute function is fixed (cf: https://github.com/OCA/account-reconcile/pull/322#discussion_r779431295 )

gurneyalex:
@grindtildeath could you propose this fix in a PR on @i-vyshnevska 's branch?

grindtildeath:
@gurneyalex I added an extra commit since I had write access :wink:

grindtildeath (approved):

account_mass_reconcile/models/base_advanced_reconciliation.py
```diff
@@ -5,7 +5,8 @@
 import logging
 from itertools import product
 
-from odoo import models, api
+import odoo
+from odoo import models, api, sql_db
```
ivantodorovich:
```suggestion
from odoo import models, api
```

Travis complains:
```
account_mass_reconcile/models/base_advanced_reconciliation.py:9:1: F401 'odoo.sql_db' imported but unused
```

account_mass_reconcile/models/base_advanced_reconciliation.py
```diff
@@ -214,16 +215,65 @@ def _skip_line(self, move_line):
...
+                new_env = api.Environment(new_cr, self.env.uid, self.env.context)
+                self_env = self.with_env(new_env)
+                rec = self_env.create(self.copy_data())
+        for i in range(0, len(reconcile_groups), chunk_size):
+            chunk = reconcile_groups[i:i + chunk_size]
+            _logger.debug("Reconcile group chunk %s", chunk)
+            try:
+                with odoo.api.Environment.manage():
+                    with odoo.registry(self.env.cr.dbname).cursor() as new_cr:
+                        new_env = api.Environment(new_cr, self.env.uid, self.env.context)
```
ivantodorovich:
```suggestion
                        new_env = api.Environment(
                            new_cr, self.env.uid, self.env.context
                        )
```

```
account_mass_reconcile/models/base_advanced_reconciliation.py:261:89: E501 line too long (89 > 88 characters)
```

Travis complains

ivantodorovich (approved):
LG but travis linting is red, see comments below
pre-approve

PR #344, opened 2020-10-07
[11.0] [ADD] account_reconcile_payment_order: Inbound and outbound payment's support

vinclar:
If self.amount is positive, the account payment order which is an outgoing payment will never be proposed on the bank statement line to match.
If self.amount is negative it will be proposed to be reconciled on the bank outgoing payment from the funds of the account.

This is what happens actually, if the proposed value is changed to negative it will be suggested to reconcile properly on the bank statement.

@pedrobaeza  ¿What do you think?

Before changing the value

CapturaA
CapturaB
CapturaC

After changing the value

CapturaD

### DIFF
--- a/account_reconcile_payment_order/models/account_bank_statement_line.py
+++ b/account_reconcile_payment_order/models/account_bank_statement_line.py
@@ -15,11 +15,12 @@ class AccountBankStatementLine(models.Model):
         return self.env['account.payment.order'].search([
             ('total_company_currency', '=', self.amount),
             ('state', 'in', ['done', 'uploaded']),
-			('payment_type', '=', ['inbound']),
-		]), self.env['account.payment.order'].search([
+            ('payment_type', '=', ['inbound']),
+        ]),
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I think you are checking the opposite case that I have programmed this for: a payment order instead of a debit order, but we should cover both.

pedrobaeza:
The failed test is reflecting this problem I comment.

vinclar:
Dear @pedrobaeza Pedro, 
I'm checking the errors on the continuous-integration/travis-ci/pr — The Travis CI build failed checks, and none of them are from the pull request module. Are module dependencies perhaps?

vinclar:
@pedrobaeza Now test have passed, and for me it's working in both cases. Inbound and Outbound payments.

pedrobaeza:
Seems correct. Please squash all your commits into one.

vinclar:
Rebased in #349

PR #347, opened 2020-10-16
[12.0] account_move_transactionid_import: Restore feature to import single move

grindtildeath:
TODO:
 - [x] Add tests

Fixing changes done on https://github.com/OCA/account-reconcile/pull/296

### DIFF
--- a/account_move_transactionid_import/parser/transactionid_file_parser.py
+++ b/account_move_transactionid_import/parser/transactionid_file_parser.py
@@ -73,8 +73,6 @@ class TransactionIDFileParser(FileParser):
 
     def get_move_vals(self):
         res = super().get_move_vals()
-        if "ref" in res:
-            res.pop("ref")
         if res.get("name") == "/":
             res["name"] = self.move_ref
         return res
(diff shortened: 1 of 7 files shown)


account_move_transactionid_import/parser/transactionid_file_parser.py
```diff
@@ -36,20 +36,19 @@ def __init__(
...
             ftype=ftype,
             header=header,
             **kwargs
         )
-        self.support_multi_moves = True
 
     @classmethod
     def parser_for(cls, parser_name):
         """Used by the new_bank_statement_parser class factory. Return true if
         the providen name is generic_csvxls_transaction
```
i-vyshnevska: hook to check parser .....

i-vyshnevska (approved):

account_move_transactionid_import/parser/transactionid_file_parser.py
grindtildeath: I did removed it altogether since the classes it's inheriting from already return False ;)

guewen (approved):

PR #321, opened 2020-07-08
[12.0][IMP] base_transaction_id: multiple transaction ref

i-vyshnevska:
do not overwrite ref as there can be few transactions related to
one move

PR #352, opened 2020-11-02
[12.0] account_move_transactionid_import fix

gurneyalex:
in #347 and #321 we adapt the way the transactionID is propagated on the
account.move / account.move.line: using the move.ref is not efficient in
case of payment remittance import -> we move the transactionID to the
name of the account.move.line (which is how it is handled in
odoo >= 13.0).

This PR follows to adapt the autocompletion based on transactionID.


gurneyalex:
move to #347

PR #353, opened 2020-11-03
[IMP] account_mass_reconcile advanced.label

gurneyalex:
Added the 'mass.reconcile.advanced.label' model to at least be able to
reconcile payments as Odoo would. In Odoo the customer invoice has the
payment reference in the account move ref field, and the payment
typically has this in the move line name field. Since the invoice
account move line that we want to reconcile for customer invoices is a
debit line, and the payment we want to find is a credit line, we need to
invert the current implementation of the reconciliation

'mass.reconcile.advanced.ref' should be used to reconcile refunds.

### DIFF
--- a/account_mass_reconcile/models/advanced_reconciliation.py
+++ b/account_mass_reconcile/models/advanced_reconciliation.py
@@ -1,4 +1,4 @@
-# Copyright 2012-2016 Camptocamp SA
+# Copyright 2012-2020 Camptocamp SA
 # Copyright 2010 Sébastien Beau
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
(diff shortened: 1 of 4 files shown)


account_mass_reconcile/models/advanced_reconciliation.py
```diff
@@ -100,5 +100,26 @@ def _opposite_matchers(move_line):
         :yield: matchers as tuple ('matcher key', value(s))
         """
         yield ('partner_id', move_line['partner_id'])
-        yield ('ref', ((move_line['ref'] or '').lower().strip(),
-                       move_line['name'].lower().strip()))
+        yield ('ref', ((move_line['name'] or '').lower().strip(),
+                       move_line['ref'].lower().strip()))
```
guewen: FTR, I thought we should write `(move_line['ref'] or '')` because the ref is not required, but it can't happen here since `_skip_line` exclude lines without ref.

guewen (approved):

account_mass_reconcile/models/advanced_reconciliation.py
guewen: Note: we should indeed check for empty ref because `_skip_line` exclude only the credit move line, not the lines to match.

damdam-s (approved):

PR #355, opened 2020-11-09
[12.0] Add field to optionally define analytic account on commission line

gurneyalex:
this feature was present in 9.0 but had not been migrated to later versions.


jcoux (approved):
Technical review only.

PR #361, opened 2020-12-29
[14.0][MIG] account_move_reconcile_helper: Migration to 14.0

hildickethan:
Standard migration to 14.0.

Based on https://github.com/OCA/account-reconcile/pull/301 13.0 migration, ideally that one would be merged first but it does not strictly depend on it.

Depends on 
- [X] [account_balance_line](https://github.com/OCA/account-financial-tools/pull/1105)

I removed the `partial_reconciliation_in_progress` as the new computed `matching_number` [does basically the same thing](https://github.com/odoo/odoo/blob/14.0/addons/account/models/account_move.py#L3486-L3494)


CasVissers-360ERP (approved):
Functional review.

hildickethan:
rebased to 14.0 to rerun travis, as its previous fail was before the dependency was merged

fuentes73 (approved):

hildickethan:
Updated as I detected that in the tax audit tree, the `matching_number` field I use in the domain is replaced in base Odoo so I need to add it myself. The replace in `view_move_line_tax_audit_tree` has been there for years so logically this should have been breaking since the beginning, but when trying to replicate the error locally it didn't happen for me until I updated base Odoo code. 
Unfortunately I didn't pay attention to what commit I was on before updating so I can't be precise, but I believe it was some commit from March of this year, so since then something was done to fix this tree working when it shouldn't have

CasVissers-360ERP:
We had the same issue in the tree view, which is fixed by the last commit. Thnx.

Abranes (approved):
Tested with multiple customers and it works correctly

xavier-bouquiaux:
Hi @hildickethan 
I pushed a commit at your side.
I've just made a small change in the way you call the action so the admistration-setting right are no longer required
KR
Xbo.

hildickethan:
> Hi @hildickethan I pushed a commit at your side. I've just made a small change in the way you call the action so the admistration-setting right are no longer required KR Xbo.

Thanks, I merged it

hildickethan:
Rebased to oca/14.0

hildickethan:
Don't know what this pre-commit error is, it passes locally

dreispt:
@hildickethan You need to install and run run pre-commit checks locally in your Git checkout .

hildickethan:
@dreispt I did run it locally before and it passed with no issues, it seems like it was failing due to some cache or library version conflict problem at the time
I have rebased to current 14.0 and it passed this time, so everything should be ok now

PR #365, opened 2021-02-04
[14.0] account_reconciliation_widget: Use payment_ref for the statement line label

alexis-via:
This fixes a problem following the merge of PR #359 : in the reconcile interface of the bank statement, the move name was displayed instead of the label for the bank statement line. 
This is because the label of the bank statement line was renamed from 'name' to 'payment_ref' between odoo v13 and v14.

### DIFF
--- a/account_reconciliation_widget/models/reconciliation_widget.py
+++ b/account_reconciliation_widget/models/reconciliation_widget.py
@@ -1033,6 +1033,7 @@ class AccountReconciliation(models.AbstractModel):
             # FIXME: where to fill?
             # 'note': st_line.note or "",
             "name": st_line.name,
+            "payment_ref": st_line.payment_ref,
             "date": format_date(self.env, st_line.date),
             "amount": amount,
             "amount_str": amount_str,  # Amount in the statement line currency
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Can you please put a before and after screenshots?

account_reconciliation_widget/models/reconciliation_widget.py
```diff
@@ -1033,6 +1033,7 @@ def _get_statement_line(self, st_line):
             # FIXME: where to fill?
             # 'note': st_line.note or "",
```
StefanRijnhart: While we here, I believe this should be `st_line.narration` as inherited from `account.move`. The `note` field was removed in Odoo 14: https://github.com/OCA/OpenUpgrade/blob/14.0/openupgrade_scripts/scripts/account/14.0.1.1/upgrade_analysis.txt#L47

(better to rename the key as well, both here and in the qweb template).

StefanRijnhart:
@pedrobaeza I can confirm this data model change. Odoo will only allow the `name` field (now inherited from account.move) to be used for the sequential journal entry number as enforced by the `sequence.mixin`.

alexis-via:
Before:

before

After: you have the proper label

pedrobaeza:
Great, if you perform the change commented by Stefan, we proceed with the merge.

StefanRijnhart (approved):

StefanRijnhart:
Oops needs pre-commit

pedrobaeza (approved):

PR #370, opened 2021-03-26
[13.0][FIX] account_reconcile_payment_order: Fix tests

joao-p-marques:
Create journal instead of searching
Use computed value for total price instead of hardcoded (helps avoid integration tests when another module changes fiscal plan)

@Tecnativa
TT28962

ping @victoralmau @pedrobaeza

### DIFF
--- a/account_reconcile_payment_order/tests/test_account_reconcile_payment_order.py
+++ b/account_reconcile_payment_order/tests/test_account_reconcile_payment_order.py
@@ -12,14 +12,8 @@ class TestAccountReconcilePaymentOrder(TestPaymentOrderInboundBase):
     def setUpClass(cls):
         super().setUpClass()
         cls.widget_obj = cls.env["account.reconciliation.widget"]
-        cls.bank_journal = cls.env["account.journal"].search(
-            [
-                ("type", "=", "bank"),
-                "|",
-                ("company_id", "=", cls.env.company.id),
-                ("company_id", "=", False),
-            ],
-            limit=1,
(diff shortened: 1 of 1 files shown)


victoralmau (approved):

account_reconcile_payment_order/tests/test_account_reconcile_payment_order.py
```diff
@@ -31,17 +25,29 @@ def setUpClass(cls):
...
                 "journal_id": cls.bank_journal.id,
                 "line_ids": [
-                    (0, 0, {"date": "2019-01-01", "name": "Test line", "amount": 230}),
+                    (
+                        0,
+                        0,
+                        {
+                            "date": "2019-01-01",
+                            "name": "Test line",
+                            "amount": payment_order.total_company_currency,
```
pedrobaeza: This goes against the nature of the tests. We should fix data.

joao-p-marques: Ok, I think it should be good now, along with https://github.com/OCA/bank-payment/pull/797 (this tests use the base from that module)

victoralmau: PR it's complete now, it's ready to merge?

joao-p-marques: It should be completed with the latest change, although test here will only pass after https://github.com/OCA/bank-payment/pull/798 is merged

PR #373, opened 2021-04-07
[14][MIG] account_move_base_import

florian-dacosta:
Standard migration
+ Fix commission feature to make it compatible with split counterpart option and with custom parsers
+ Make it possible to automatically validates created move(s) in case it has been fully auto-completed.


rvalyi (approved):
OK for me overall. But hey @florian-dacosta we are not at Tiny SPRL, data is plural already (the singular is datum), so would be great if you change this `_get_attachment_datas` to `_get_attachment_data`, it hurts my eyes...

rvalyi:
thanks for the fix. Still ok for me then.

hildickethan (approved):
LGTM. To keep commit guidelines, I would prefix the 2 new commits with [IMP]

rvalyi:
@hildickethan I changed the 2 last commit labels as you requested (rebase -i). So now this PR is approved let's merge it...

florian-dacosta:
@rvalyi Thanks for the merge try. It seems it failed though, I just rebased it in case it solve the issue
Could you try again?

rvalyi:
@florian-dacosta I will retry now. In any case the merge bot does an automatic rebase but let's see if it works this time.

rvalyi:
@florian-dacosta no luck, same error:



tests seems to work on HEAD of 14.0 and some other 14.0 PRs however so may be there is some kind of incompatibility or may be an issue with the data created by the tests, not sure. Can you find out?

florian-dacosta:
@rvalyi 
I looked into it. And I really did not understand the issue... Let me try to explain with an example
In `account_reconcile_reconciliation_date` tests use journals, it finds the journal with a search and LIMIT 1. The test user has 2 allowed companies. So, the journal search domain correspond to 2 journals in Odoo database, but only one is returned because of the limit.
So test user is allowed on company 2 and 3.
we search a journal with type "sale", with limit 1. Odoo return the journal from company 2. => The test work

Now, if we install account_move_base_import, the same tests fails because instead of returning the journal from company 2, it returns the journal from company 3...
I dug into it and I found that this line causes the issue : https://github.com/akretion/account-reconcile/blob/14.0-mig--account-move-base-import/account_move_base_import/models/account_journal.py#L69
It makes no sense to me, but adding a boolean field on the journal with default = True, changes the result of the journal search with LIMIT 1.
Even stranger, this change of order is only effective the first time the tests are launched at installation. If you install the modules and then lauch tests again, the behavior is not changed but the default True boolea field and the tests are ok.

Anyway, even if I feel I am missing something around here, IMO, the root problem is that the `account_reconcile_reconciliation_date` module make a search on journals without specifying the company, although there are multiple journals. So the result could be kind of random, since the order is not enough to always get the same result. (both journals from tests have the same sequence, type and code, which is the default order on journals).
What I did to fix this was just to add the company in the journal search domain...

I think it solves our issue.

rvalyi (approved):
@florian-dacosta I think this workaround is legit. At least we are pretty sure it's a test issue and not some fundamental module design incompatibility. Thank you for the work...

PR #380, opened 2021-04-20
[14.0][FIX] account_reconciliation_widget: error checking res.partner.bank on  account.bank.statement.line

ozono:
bank_account_id is now partner_bank_id on account.bank.statement.line (inheriting from move)

### DIFF
--- a/account_reconciliation_widget/models/account_bank_statement.py
+++ b/account_reconciliation_widget/models/account_bank_statement.py
@@ -200,7 +200,7 @@ class AccountBankStatementLine(models.Model):
             )
 
         # create the res.partner.bank if needed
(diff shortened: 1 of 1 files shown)


bosd (approved):
Tested on Runbot
LGTM! :+1: 
Now we are able to reconcile :tada: 
Hope we can get this fix merged soon @pedrobaeza :pray:

bosd:
Thinking out loud; Should this scenario have been covered by some tests in the first place?

pedrobaeza (approved):
Regression tests are indeed welcome, but let's not block it for the lack of them, as this module is yet a bit low covered (and the problem comes from Odoo itself).

PR #382, opened 2021-04-21
[14.0] account_reconciliation_widget: remove payment_id <> false

alexis-via:
Remove the condition on payment_id <> false in the method _domain_move_lines_for_reconciliation()
because the OCA module account_payment_order doesn't generate account.payment objects.

PR #390, opened 2021-05-17
[14.0][IMP] account_reconcile_widget: Payments without payment_id related aren't showed in statement reconciliation

joao-p-marques:
Port of https://github.com/OCA/OCB/pull/678

`account.reconciliation.widget` has been removed from Odoo community in v14 and ported to `account_reconcile_widget`

@pedrobaeza @carlosdauden can you review please?

### DIFF
--- a/account_reconciliation_widget/models/reconciliation_widget.py
+++ b/account_reconciliation_widget/models/reconciliation_widget.py
@@ -761,10 +761,8 @@ class AccountReconciliation(models.AbstractModel):
         domain_reconciliation = [
             "&",
             "&",
-            "&",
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Ouch, this is duplicated with #382

joao-p-marques:
Oops, true, although the original patch from @carlosdauden also changes the order of the operations and puts the if right after the domain definition, and not at the end...

pedrobaeza:
Please comment it in the other thread, although I want to do in the OCA Spanish Days the PoC for the `account.payment` generation from payment orders, doing this unneeded.

PR #371, opened 2021-04-01
[14][MIG] account_mass_reconcile

florian-dacosta:
Standard migration, no real feature added, but still lot of changes because of the difference in Odoo core about accounting.
Main changes concerns the write-off management as it is not managed by Odoo account module anymore in reconcile method and manage multi currency better with write off.

@AdriaGForgeFlow @LoisRForgeFlow @AaronHForgeFlow

PR #394, opened 2021-06-29
[14.0] account_mass_reconcile: Extra options

atchuthan:
https://github.com/OCA/account-reconcile/pull/371

We added extra options based on the PR

cc: @SodexisTeam


AaronHForgeFlow:
Can you please let us know what are those extra options? it is hard to know by just looking at the code.

PR #396, opened 2021-07-14
[FIX] Prefill partner in reconciliation widget if matched by rule

florian-dacosta:
During the reconcile model rule evaluation, Odoo can find a partner from the rules : 
https://github.com/OCA/OCB/blob/14.0/addons/account/models/account_reconcile_model.py#L486
Then, if candidates have been found the the statement line, the partner is added to the result : https://github.com/OCA/OCB/blob/14.0/addons/account/models/account_reconcile_model.py#L511

Today, if a partner is found this way, it is still empty in the reconciliation widget. With this fix, it will be prefilled.

@alexis-via @pedrobaeza @ozono

### DIFF
--- a/account_reconciliation_widget/models/reconciliation_widget.py
+++ b/account_reconciliation_widget/models/reconciliation_widget.py
@@ -269,14 +269,19 @@ class AccountReconciliation(models.AbstractModel):
                     and matching_amls[line.id]["model"].id,
                     "write_off": matching_amls[line.id].get("status") == "write_off",
                 }
-                if not line.partner_id and partner_map.get(line.id):
-                    partner = self.env["res.partner"].browse(partner_map[line.id])
-                    line_vals.update(
-                        {
-                            "partner_id": partner.id,
-                            "partner_name": partner.name,
-                        }
-                    )
+                if not line.partner_id:
+                    partner = False
+                    if matching_amls[line.id].get("partner"):
+                        partner = matching_amls[line.id]["partner"]
+                    elif partner_map.get(line.id):
+                        partner = self.env["res.partner"].browse(partner_map[line.id])
+                    if partner:
+                        line_vals.update(
+                            {
+                                "partner_id": partner.id,
+                                "partner_name": partner.name,
+                            }
+                        )
                 results["lines"].append(line_vals)
 
         return results


ozono (approved):
Thanks Florian. LGTM.

pedrobaeza:
Not sure of wanting this behavior, as it's very annoying when reconciling a payment/debit order for several partners, that you have to explicitly clear the partner field, as it has been filled automatically with the first selected line.

florian-dacosta:
> Not sure of wanting this behavior, as it's very annoying when reconciling a payment/debit order for several partners, that you have to explicitly clear the partner field, as it has been filled automatically with the first selected line.

I am not sure what is this case, could you explain more ?
AFAIK, the only case where the partner will be prefilled (in addition to the already implemented prefilled cases) will be if you have configured your reconciliation model with a specific partner mapping.
If you do not configure any rule with partner mapping, it should not set any partner, unless I missed something?

Today, I think it is really confusing, you have reconciliation models with possibility to add partner mapping, but it has no effect !
Worse, if you set a partner mapping, Odoo could find the partner making the rule eligible, use it for the reconciliation widget, and still put an empty partner.

ozono:
I agree with Florian.  (IMO) You need specific  mapping. This not affects default behavior.

pedrobaeza (approved):
OK, it seems not the same I'm talking about

rvalyi:
> Not sure of wanting this behavior, as it's very annoying when reconciling a payment/debit order for several partners, that you have to explicitly clear the partner field, as it has been filled automatically with the first selected line.

@pedrobaeza , just to know: were you talking about the case like your bank statement has not a lot of details and one bank statement line actually matches serval invoices from different partners? If so we find it to be a serious usability issue for the user to think to clear the statement line partner before searching the other counterparts indeed...

pedrobaeza:
> @pedrobaeza , just to know: were you talking about the case like your bank statement has not a lot of details and one bank statement line actually matches serval invoices from different partners? If so we find it to be a serious usability issue for the user to think to clear the statement line partner before searching the other counterparts indeed...

Yes, I'm talking about such case, that happens when you issue a debit order for 10 partners for example, and your bank exposes only one transaction for the whole amount. You look for the first invoice/transfer journal item, select it, and the partner of such line is filled on the partner field. You have to clear it, and then continue looking for the rest of items.

PR #397, opened 2021-07-15
[IMP] account_reconciliation_widget: reduce heavy queries count

yajo:
This specific query is quite expensive, and what's worse, it's also done twice: once to count records and once to get them.

By using a window function, we can get the count and results in one query, thus almost doubling performance in this operation.

Related resources:

- [Explanation](https://stackoverflow.com/a/22353886/1468388).
- [Explain of the 1st SELECT COUNT(*), before the patch](https://explain.dalibo.com/plan/TgL).
- [Explain of the 2nd SELECT, before the patch](https://explain.dalibo.com/plan/Yc9).
- [Explain of the new, unified SELECT, after the patch](https://explain.dalibo.com/plan/9MA).

@Tecnativa TT28595 TT29916 https://github.com/odoo/odoo/pull/73765

### DIFF
--- a/account_reconciliation_widget/models/reconciliation_widget.py
+++ b/account_reconciliation_widget/models/reconciliation_widget.py
@@ -102,14 +102,13 @@ class AccountReconciliation(models.AbstractModel):
             search_str=search_str,
             mode=mode,
         )
-        recs_count = self.env["account.move.line"].search_count(domain)
 
         from_clause, where_clause, where_clause_params = (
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

joao-p-marques (approved):

carlosdauden (approved):

yajo:
May somebody merge?

PR #403, opened 2021-09-07
[IMP] Account reconcilation widget allows statement in open state

LuceEasyPme:
Allow partial reconciliation for monthly bank statements.
Before PR: After importing a bank statement, you have to post it before being able to reconcile.
After PR: After importing a bank statement, you have can reconcile items, even if the statement is in open state. It allows to import a partial bank statement and reconcile the items, then import new lines in the same bank statement, and post it when the statement is complete.

Use case: our client wants to import and reconcile the items daily (so the bills are paid), even thought his official bank statement is available one a month.

cc @cvinh

### DIFF
--- a/account_reconciliation_widget/models/reconciliation_widget.py
+++ b/account_reconciliation_widget/models/reconciliation_widget.py
@@ -312,7 +312,7 @@ class AccountReconciliation(models.AbstractModel):
              FROM account_bank_statement_line line
              LEFT JOIN res_partner p on p.id = line.partner_id
              INNER JOIN account_bank_statement st ON line.statement_id = st.id
-                AND st.state = 'posted'
+                AND st.state IN ('posted', 'open')
              WHERE line.is_reconciled = FALSE
              AND line.amount != 0.0
              AND line.id IN %(ids)s
--- a/account_reconciliation_widget/views/account_bank_statement_view.xml
+++ b/account_reconciliation_widget/views/account_bank_statement_view.xml
@@ -12,7 +12,7 @@
                         string="Reconcile"
                         type="object"
                         class="oe_highlight"
-                        attrs="{'invisible':['|','|',('all_lines_reconciled','=',True),('line_ids','=',[]),('state', '!=', 'posted')]}"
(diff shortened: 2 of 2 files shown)


pedrobaeza:
This goes against v14 way of working, so this shouldn't be allowed.

cvinh:
@pedrobaeza please note that when you reconcile entries, the account moves get posted at the same time... so IMHO it can be accepted...
Do you know how EE works with bank sync ? Are there daily/hourly bank statements ? How do they validate initial/ending balances in the case where they have monthly legal paper/pdf statements ? 
Thanks for your help

pedrobaeza:
The post operation makes the outstanding entries, so you can't skip them, as dashboard and so rely on these entries for showing correct data. Both OCA and enterprise fill draft statements, but for reconciling you have to post before, and then ending balance is adjusted by Odoo on the rest of the statements.

LuceEasyPme:
Hello @pedrobaeza ,
Actually, the outstanding entries are already created, they are in draft state. When we reconcil them, they are posted.
You can see that in this video : https://www.youtube.com/watch?v=kBxx0gYMGwY&feature=youtu.be
Thank you very much for your help!
cc @alexis-via

pedrobaeza:
I'm still reluctant of this change and possible side effects. At least, it should be configurable and disabled by default. I let others also to give their opinion. I will come back anyways in a couple of weeks after further testing.

tv-openbig:
@LuceEasyPme @pedrobaeza In general it isn't a bad idea to work already on non posted statement lines to allow them to reconcile against open invoices before final posting of completed statement. However this scenario shows that it works with non reconciled open invoices. Does it also work as expected if we have non posted statement lines and invoices proceeded in a first step by creating a payment entry directly on the invoice (counterpart account would have to be the "Outstanding Receipts" account instead of the partners debit account automatically proposed in your video) ?

pedrobaeza:
I think it can be acceptable if put it as configurable.

cvinh:
> I think it can be acceptable if put it as configurable.

Good idea

cvinh:
> Does it also work as expected if we have non posted statement lines and invoices proceeded in a first step by creating a payment entry directly on the invoice (counterpart account would have to be the "Outstanding Receipts" account instead of the partners debit account automatically proposed in your video) ?

You are right, it has to be tested

PR #405, opened 2021-10-03
[13.0][account_reconcile_only_posted][new]

JordiBForgeFlow:
This module ensures that only posted entries can be reconciled together. This ensures the consistency of accounting numbers. Allows, for example, that reports such as the Open Items will match with the Trial Balance for a reconcilable account.

@ForgeFlow

### DIFF
--- a/account_reconcile_only_posted/models/account_partial_reconcile.py
+++ b/account_reconcile_only_posted/models/account_partial_reconcile.py
@@ -1,0 +1,35 @@
+# Copyright 2021 ForgeFlow
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo import _, api, models
+from odoo.exceptions import ValidationError
+
+
+class AccountPartialReconcile(models.Model):
+
+    _inherit = "account.partial.reconcile"
+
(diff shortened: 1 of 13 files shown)


JordiBForgeFlow:
@fclementic2c Perhaps you have come across this issue sometime. It took me quite a few weeks to realize this. The auditor found that the open items report was not matching with the trial balance, for the accounts receivable. Go figure out what could happen...

account_reconcile_only_posted/models/account_partial_reconcile.py
```diff
@@ -0,0 +1,35 @@
...
+            return True
+        for rec in self.sudo():
+            cr_move = rec.credit_move_id.move_id
+            dr_move = rec.debit_move_id.move_id
+            active_model = self.env.context.get("active_model")
+            if active_model == "account.bank.statement":
+                return True
+            if dr_move.state != "posted" or cr_move.state != "posted":
+                raise ValidationError(
+                    _(
```
AaronHForgeFlow, line 29: Draft moves has '/' as name so perhaps we can also show the name of the journal item instead:

AaronHForgeFlow (approved):
LGTM! Just some minor comment

PR #407, opened 2021-10-12
[13.0][FIX] account_reconcile_payment_order: search payment order according statement amount

celm1990:
* on payment order total_company_currency is always positive but when order is outbound on bank statement amount is entered on negative, so at search result are not found and nothing proposition

this feature was lossed on commit https://github.com/OCA/account-reconcile/commit/e2d7ced32f9b17445fc4e87e4c5b73747a127297#diff-3023bdfeffb6a6bb309395eaf9e8ee6b560870e230d5f735b9b4e4e935d20afeL41

### DIFF
--- a/account_reconcile_payment_order/models/account_reconciliation_widget.py
+++ b/account_reconcile_payment_order/models/account_reconciliation_widget.py
@@ -12,12 +12,17 @@ class AccountReconciliationWidget(models.AbstractModel):
         """Find orders that might be candidates for matching a statement
         line.
         """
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Can you please add any tests for the new use case?

celm1990:
> Can you please add any tests for the new use case?

Hi @pedrobaeza sorry for delay, I add test but need changes on PR https://github.com/OCA/bank-payment/pull/891
can you help me with this PR Please?

Issue #404, opened 2021-09-29
[14.0] account_reconciliation_widget regex models client error


TeoGoddet:
Hi ! 
you can provide regex to find the amount of a suggested writeoff line.
But it seems the implementation is different between the backend (python) and the frontend (js)

Here we are matching the statement_line name (wich contain no useful infos (sequence generated move name)
    var matching = line.st_line.name.match(
           new RegExp(values.amount_string)
      );

https://github.com/OCA/account-reconcile/blob/bbbffd3eb47dcdfda5ab75dfbc89df2542e663ff/account_reconciliation_widget/static/src/js/reconciliation/reconciliation_model.js#L1601

and there they are matching the statement payment_ref
                match = re.search(line.amount_string, st_line.payment_ref)
https://github.com/odoo/odoo/blob/b36809315f12142ca9589b34b38c725100528909/addons/account/models/account_reconcile_model.py#L346

I will locally test
    var matching = line.st_line.payment_ref.match(
           new RegExp(values.amount_string)
      );

Is there a reason for redoing this calculation on the server ? Should not we retrieve it from the server ? 

Need help to choose the fix (small fix or rewrite to load from backend) and then may be able to submit PR

TeoGoddet:
I'm also sceptic about this line which mess up my number (i have . as decimal separator in my model)
https://github.com/OCA/account-reconcile/blob/bbbffd3eb47dcdfda5ab75dfbc89df2542e663ff/account_reconciliation_widget/static/src/js/reconciliation/reconciliation_model.js#L1606

TeoGoddet:
The fix is working on my installation, should i open a MR ?

Tardo:
Hi @TeoGoddet
As you say, I don't see any inconvenience in moving the logic to Python. But I think it is better to do it in a new version since it involves having to change the sync. calls to async.

In current versions I think it is more appropriate to apply the fix for the construction of the regex.

TeoGoddet:
Ok, i'll do the PR

TeoGoddet:
See #412

PR #412, opened 2021-10-22
[14.0][FIX] account_reconcile_widget : Fix generated reconcile proposal

TeoGoddet:
- fix the value on wich the regex is matched (search on payment_ref instead of name)

### DIFF
--- a/account_reconciliation_widget/static/src/js/reconciliation/reconciliation_model.js
+++ b/account_reconciliation_widget/static/src/js/reconciliation/reconciliation_model.js
@@ -1598,7 +1598,7 @@ odoo.define("account.ReconciliationModel", function (require) {
                     amount = (line.balance.amount * values.amount) / 100;
                     break;
                 case "regex":
-                    var matching = line.st_line.name.match(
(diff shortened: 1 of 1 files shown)


TeoGoddet:
Fixes #404 

With this comment https://github.com/OCA/account-reconcile/pull/359#issuecomment-773409956 on #359 i understood that the error is from the 13 to 14 change made on statement lines (name -> payment_ref)

pedrobaeza:
Please provide a pull request for each of the 3 things, as some of them are very straight. but not others.

TeoGoddet:
Hi ! Is it small enough ?

pedrobaeza:
I still see 2 changes, and I don't understand the decimal separator one. If you leave the other, I will fast-track it, and then we can discuss about the other in a new PR.

TeoGoddet:
Ok, I let only the first change
 let's discuss the decimal separator on #420

pedrobaeza (approved):

pedrobaeza:
Fast-tracking this one as the recurring `name`>`payment_ref` bug:

/ocabot merge patch

PR #417, opened 2021-10-30
[14.0][FIX] account_reconcile_widget : Fix foreign_currency_id vs currency_id on bank statement

TeoGoddet:
Fix 13 to 14 migration that broke because upstream field change on bank.statement.line
currency_id was renamed foreign_currency_id
and a new currency_id field was created (wich seem to contain the statement value)

### DIFF
--- a/account_reconciliation_widget/models/account_bank_statement.py
+++ b/account_reconciliation_widget/models/account_bank_statement.py
@@ -150,7 +150,7 @@ class AccountBankStatementLine(models.Model):
 
         # Fully reconciled moves are just linked to the bank statement
         total = self.amount
-        currency = self.currency_id or statement_currency
+        currency = self.foreign_currency_id or statement_currency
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Indeed, checked in https://github.com/odoo/odoo/commit/caeb782841fc5a7ad71a196e2c9ee67644ef9074#diff-96c5c4fc954f9deb57474d052d9c59b68cc41b610205d342949727ece08ac404L489 and https://github.com/odoo/odoo/commit/caeb782841fc5a7ad71a196e2c9ee67644ef9074#diff-96c5c4fc954f9deb57474d052d9c59b68cc41b610205d342949727ece08ac404R506 that this change happened from 13. to 14.0.

pedrobaeza:
Fast-tracking the patch for solving the problem:

/ocabot merge patch

PR #418, opened 2021-10-30
[14.0][FIX] account_reconcile_widget :  Fix analytic tags formatting

TeoGoddet:
Analytic tags are not correctly formatted when sent to the server after reconciliation, that lead to an error being raised


pedrobaeza (approved):
I have been hit now by this problem with the following steps to reproduce:

- Create a reconciliation model of type "Manually create a write-off on clicked button".
- Show analytic on the configuration and show the optional column "Analytic tags".
- Put any analytic tag.

When using such reconciliation model on the reconciliation, the analytic tag is not well formatted and you can't complete the reconciliation.

@TeoGoddet next time, that steps are the crucial one for reviewing the PR.

Merging for solving the problem:

/ocabot merge patch

PR #419, opened 2021-10-30
[14.0][IMP] account_reconcile_widget : Look for amount in all bank fields

TeoGoddet:
When using a regex to obtain an amount from the statement, it look only on payment_ref (when #412 will be merged) 
It would be nice if it's also look in ref and narration (other field that come from the bank files and that may contain infos


pedrobaeza:
I'm afraid this affects performance. Isn't the bank importer which should provide the relevant information in payment_ref? And also why looking for amounts in text fields instead on the proper one?

TeoGoddet:
This feature is used too find amounts in the text such as a gross amount when an acquirer pays the net amount or things like this.

PR #420, opened 2021-10-30
[14.0][FIX] Fix value cleaning and decimal separator choice

### LINKED ISSUE #478: account_reconciliation_widget label match decimal_separator broken
The regexp that is meant to remove everything that is not a digit or the decimal separator from the label match is broken.

https://github.com/OCA/account-reconcile/blob/a29c6fc9c1cccbb287334a78cd0f17eee9013f5b/account_reconciliation_widget/static/src/js/reconciliation/reconciliation_model.js#L1606

This was fixed in the server side model in https://github.com/odoo/odoo/commit/d4da8be3669fbfdfdcad752af1f0311accaf34a3

The same fix should be applied in the `js`.

Tested that this works and that it fixes the issue.

TeoGoddet:
When decimal separator is a dot, it has a special meaning in the regex (any character) that cause problem 
For instance  in 123.40  the regex will match .4 and remove it.
And then the proposal will be base on the number 120 instead of 123.40

The other problem with the actual regex is that minus sign get removed

In the fix, we start by replacing the decimal_separator by a dot (before it was done after the cleaning) and the clean with a character class that is clearly specified (dot, minus, numbers)

### DIFF
--- a/account_reconciliation_widget/static/src/js/reconciliation/reconciliation_model.js
+++ b/account_reconciliation_widget/static/src/js/reconciliation/reconciliation_model.js
@@ -1603,7 +1603,10 @@ odoo.define("account.ReconciliationModel", function (require) {
                     );
                     if (matching && matching.length == 2) {
                         matching = matching[1].replace(
-                            new RegExp("\\D" + reconcileModel.decimal_separator, "g"),
+                            new RegExp(
+                                "[^-d" + reconcileModel.decimal_separator + "]",
+                                "g"
+                            ),
                             ""
                         );
                         matching = matching.replace(


pedrobaeza:
What I don't get is why this has broken, being the same as in 13.0.

pedrobaeza:
Can you also add some steps to reproduce the problem before this patch?

TeoGoddet:
I just tested on odoo entreprise v13 
And it's seem broken too : 
Screenshot_20211030-223054

Screenshot_20211030-223125


This is the statement I tested on : 
https://10809030-13-0-all.runbot80.odoo.com/web#id=1&action=660&active_id=8&model=account.bank.statement&view_type=form&cids=1&menu_id=1019

TeoGoddet:
@pedrobaeza do you need a better explanation on how to reproduce ?

account_reconciliation_widget/static/src/js/reconciliation/reconciliation_model.js
```diff
@@ -1603,13 +1603,10 @@ odoo.define("account.ReconciliationModel", function (require) {
...
                     if (matching && matching.length == 2) {
                         matching = matching[1].replace(
-                            new RegExp("\\D" + reconcileModel.decimal_separator, "g"),
-                            ""
-                        );
-                        matching = matching.replace(
                             reconcileModel.decimal_separator,
                             "."
                         );
+                        matching = matching.replace(new RegExp("[^0-9.-]", "g"), "");
```
jordens: This is wrong. It breaks when "." is the thousands separator and "," decimal separator.
The code should be in the original location and the regex should be just ```new RegExp("[^-\d" + reconcileModel.decimal_separator + "]", "g"),```

See #478 and the code in https://github.com/odoo/odoo/commit/d4da8be3669fbfdfdcad752af1f0311accaf34a3

jordens (changes requested):

jordens:
Thanks. I overlooked this.

closes #478

account_reconciliation_widget/static/src/js/reconciliation/reconciliation_model.js
jordens: ~Maybe~Probably even without the `-`. I'm not entirely sure. All this looks like somewhat fragile code with a lot of overlap w.r.t. the host side python code.

TeoGoddet: This module should be rewritten to use the python backend for most operations and js only for consuming it, probably on v16 migration as the odoo module was completly rewritten there.

TeoGoddet: But yeah I did not see the problem for the thousand separator, indeed need to be changed, but i'm in favor of keeping the minus sign, as you can easily remove it using the main regex if you need to

jordens: Ack

pedrobaeza: We should fix it here for now on the best possible way, and don't wait on v16 refactoring, as all of that is enterprise and we can't access such code legally. The community part won't affect this JS part too much.

TeoGoddet: @pedrobaeza please check this https://github.com/odoo/odoo/commit/080d41a34119191d20b3e07f72f6aeec677ee1a8?s=09

**the regex should be fixed here and now** but regarding this : 
> All this looks like somewhat fragile code with a lot of overlap w.r.t. the host side python code.


i think that it would be a mistake to write code that depend on things that are removed in v16
I don't know how much we want to depend on odoo SA, we can either wait for v16 to use their backend code or refactor the module with OCA front/backend code completely. This should have an issue for itself

jordens (approved):

account_reconciliation_widget/static/src/js/reconciliation/reconciliation_model.js
pedrobaeza: The question here is that there are comments saying that this fix is not enough, so it can't be merged as is, and also the pre-commit CI is red.

account_reconciliation_widget/static/src/js/reconciliation/reconciliation_model.js
```diff
@@ -1603,7 +1603,10 @@ odoo.define("account.ReconciliationModel", function (require) {
...
                     if (matching && matching.length == 2) {
                         matching = matching[1].replace(
-                            new RegExp("\\D" + reconcileModel.decimal_separator, "g"),
+                            new RegExp(
+                                "[^-d" + reconcileModel.decimal_separator + "]",
+                                "g"
+                            ),
                             ""
                         );
                         matching = matching.replace(
```
pedrobaeza, line 1612: Why this second replace then?

jordens: That's unrelated. It seems to me that discussing other areas of the original code and why it does does things the way it does them is a different question. If you think it's wrong, file an issue and submit a PR.

But this fix here is certainly correct. Let's get it merged!

TeoGoddet: The second replace allow to internationalize the float number (by setting it's decimal separator to a dot)

pedrobaeza:
OK, let's merge it for improving the situation. Please fw-port it to v15:

/ocabot merge patch

TeoGoddet:
FW port to 15 : #482

account_reconciliation_widget/static/src/js/reconciliation/reconciliation_model.js
```diff
@@ -1603,7 +1603,10 @@ odoo.define("account.ReconciliationModel", function (require) {
                     );
                     if (matching && matching.length == 2) {
                         matching = matching[1].replace(
-                            new RegExp("\\D" + reconcileModel.decimal_separator, "g"),
+                            new RegExp(
+                                "[^-d" + reconcileModel.decimal_separator + "]",
```
jordens, line 1607: In fact, this isn't correct anymore. Needs to be `"[^-\d" + reconcileModel.decimal_separator + "]",`

PR #422, opened 2021-11-29
[14.0] [MIG] base_transaction_id: Migrate to version 14.0

sonhd91:


leemannd (approved):

base_transaction_id/models/invoice.py
```diff
@@ -0,0 +1,21 @@
...
+        index=True,
+        copy=False,
+        store=True,
+        help="Transaction ID from the financial institute",
+    )
+
+    def _get_computed_reference(self):
+        if self.transaction_id:
+            return self.transaction_id
+        return super()._get_computed_reference()
```
hailangvn: I've just found that `_get_computed_reference` had been changed to `_get_invoice_computed_reference` with new helpers. So the code should be updated accordingly. (https://github.com/odoo/odoo/commit/bc131c0cfb51c953de8ec41fb820c8c7831eefb5)
```suggestion
    def _get_invoice_reference_odoo_invoice(self):
        if self.transaction_id:
            return self.transaction_id
        return super()._get_invoice_reference_odoo_invoice()
```

PR #440, opened 2022-01-31
[14.0][FIX] account_reconcile_widget: Apply the tax_tag_ids correctly to the lines

### LINKED ISSUE #432: [14.0] account_reconcile_widget missing tax grid entry in created journal items
In day-to-day accounting work, there can always be direct debits from suppliers for whom there are no direct incoming invoices (contracts, etc.). These are usually posted to an expense account using the bank statement instead of a supplier invoice. In addition to the expense account selection on manual operations tab, the reconcile widget also allows you to choose a tax (e.g. 19% input tax incl. In price). The posting to the input tax account takes place smoothly, but the field tax_tag_ids (field is named as "tax grids") must also be filled. With "19 input tax inl. In price" this would be, for example, "-66". This code is required for the monthly tax report (advance VAT return), otherwise incorrect values ​​would be transmitted.

This function is included in the Reconcile Widget of the Odoo Enterprise Edition. 

The video shows this clearly and compares to the correct way in Odoo EE:
https://drive.google.com/file/d/15kJe-KS1u2GDRaXLWi6o5XJKGjdwUOXs/view

fkantelberg:
This attempts to fix the issue #432 where the tax_tag_ids aren't correctly copied from the tax to the account.move.line. This causes empty tags grids.

### DIFF
--- a/account_reconciliation_widget/static/src/js/reconciliation/reconciliation_model.js
+++ b/account_reconciliation_widget/static/src/js/reconciliation/reconciliation_model.js
@@ -1635,7 +1635,7 @@ odoo.define("account.ReconciliationModel", function (require) {
                 ),
                 journal_id: this._formatNameGet(values.journal_id),
                 tax_ids: this._formatMany2ManyTagsTax(values.tax_ids || []),
-                tag_ids: values.tag_ids,
+                tax_tag_ids: this._formatMany2ManyTagsTax(values.tax_tag_ids || []),
                 tax_repartition_line_id: values.tax_repartition_line_id,
                 debit: 0,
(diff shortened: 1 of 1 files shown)


tv-openbig:
@fkantelberg I have checked the fix. It now correctly assigns the tax_tag_ids to the move lines of a bank statement if an expense / income + a tax will be used on a manual operation to create a counterpart booking transaction towards the bank account directly, instead of creating a supplier bill with a tax and the reconciling with a bank statement line (outbound).

pedrobaeza (approved):
Clearly an old field name that should be changed.

PR #443, opened 2022-02-14
[14.0][IMP] account_reconcile_widget: Allow to reconcile payment/debit orders from OCA/bank-payment

pedrobaeza:
Payments without payment_id related aren't showed in statement reconciliation.

With this change, we show them as blue lines, the same as other coming from Odoo core payments.

Similar to OCA/OCB#884

@Tecnativa TT34508


carlosdauden (approved):

pedrobaeza:
Cherry-pick to v15 at 7c98fd89fa68f9836d605d38808c7a369d4d4866

PR #454, opened 2022-03-17
[14.0][FIX] account_reconciliation_widget: avoid unbalanced error on recompute

ozono:
Some module inheriting 'account.move' and raising a recompute over (unfinished) original statement move cause an undesired 'Cannot create unbalanced journal entry' error.

### DIFF
--- a/account_reconciliation_widget/models/account_bank_statement.py
+++ b/account_reconciliation_widget/models/account_bank_statement.py
@@ -241,7 +241,9 @@ class AccountBankStatementLine(models.Model):
         aml_to_reconcile = []
         for aml_dict in counterpart_aml_dicts:
             if not aml_dict["move_line"].statement_line_id:
-                aml_dict["move_line"].write({"statement_line_id": self.id})
+                aml_dict["move_line"].with_context(check_move_validity=False).write(
+                    {"statement_line_id": self.id}
+                )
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Isn't the problem in the module doing such recomputation?

ozono:
Maybe. in my case occurs with https://github.com/OCA/l10n-spain/pull/2044 (I know this is an unapproved alpha)

But i think this is not a bad practice. We are writing an unrelated-to-balabace field while working on an unbalanced move...

pedrobaeza:
Can you point in which line of 190 module happens this for trying to trace what is happening? I think such module doesn't write on move lines.

ozono:
This one.

https://github.com/ForgeFlow/l10n-spain/blob/b31734be40699562a6370e215035c2768f3fcbd4/l10n_es_aeat_mod190/models/account_move_line.py#L41

pedrobaeza:
:scream: A compute must never write anything!! The `write` must be replaced by an `update`. That's the reason.

pedrobaeza:
I have pushed OCA/l10n-spain@f950b4577185794da8017e429b77d93cc0fa1b1e for v13, as the same problem is there. This can be closed IMO. I can't think of a situation where writing this field in this module can lead to non balanced moves.

ozono:
Thanks Pedro

PR #459, opened 2022-04-18
[WIP][14.0][IMP] account_reconciliation_widget: Use in_payment

pedrobaeza:
This was forced to "paid" on community, but we can make use of the feature.

See https://github.com/odoo/odoo/blob/c8c3c133385a36460f3da7ed00b59116c9bcf693/addons/account/models/account_move.py#L1346-L1348 for more info.

### DIFF
--- a/account_reconcile_reconciliation_date/tests/test_account_reconcile_reconciliation_date.py
+++ b/account_reconcile_reconciliation_date/tests/test_account_reconcile_reconciliation_date.py
@@ -187,28 +187,4 @@ class TestAccountReconcileReconciliationDate(AccountTestInvoicingCommon):
         register_payments.flush()
         register_payments.action_create_payments()
         payment = self.payment_model.search([], order="id desc", limit=1)
-
-        self.assertAlmostEqual(payment.amount, 200)
-        self.assertEqual(payment.state, "posted")
-        self.assertEqual(payment.state, "posted")
-        self.assertEqual(inv_1.payment_state, "paid")
-        self.assertEqual(inv_2.payment_state, "paid")
-
-        self.assertRecordValues(
-            payment.line_ids,
-            [
-                {
-                    "journal_id": self.bank_journal_euro.id,
-                    "debit": 200.0,
-                    "credit": 0.0,
-                },
-                {
-                    "journal_id": self.bank_journal_euro.id,
-                    "debit": 0.0,
-                    "credit": 200.0,
-                },
-            ],
-        )
-
-        self.assertEqual(payment.state, "posted")
         self.assertEqual(payment.reconciliation_date, inv_1.reconciliation_date)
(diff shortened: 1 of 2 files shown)


ozono (approved):
LGTM. Maybe a better way would be enable this state in a separated module / add a dependency. But, in the other way, if another module needs this state can override this  method too...Itś OK. Thanks @pedrobaeza

pedrobaeza:
I'm still seeing the consequences, and yes, I will probably put not a separate module, but an option like the one in journals. Now Odoo always consider `in_payment` state when registering payments, so that leads to "non finished" flows without reconciliation, so I don't like it too much, but as said, I'm yet gathering all the information.

OSevangelist:
Hi @pedrobaeza any news on this matter. As far a i understand this, the problem is hindering the OCA (and probably others) to finish the annual accounting because we can properly reconcile due to the changes done in invoice / payment process in Odoo v14. Although i have had my head scratched more than enough with that design change i am a bit lost do think of a solution, as i don't fully understand the root cause of the problem with regard to the reconciliation. If you have an idea how to get arround this problem and get the reconciliation done, i would be happy if you could share you thoughts. If in turn we need engineering i would also be happy to let somebody of my team contribute to a solution but i am afraid nobody of my team is too familiar with problems that derive from that initial design change in v14. So maybe we can catch up. Any thoughts are welcome

pedrobaeza:
This is not the OCA problem, Frederik. This can be considered an extra feature. I'm still searching the problem about OCA, that is related to use the payment flow. May be related to https://github.com/OCA/OpenUpgrade/pull/3226

OSevangelist:
Ok @pedrobaeza that explains why i didn't understand the connection ;-) Thanks for clarification

TeoGoddet:
To skip the in_payment state and have 2 moves (bank->payable->expense) instead of 3 (bank->outstanding->payable->expense), one need to set the outstanding accounts to the bank account.
image
image

When you use register payment with this setup, you a got fully paid invoice directly.
See this invoice http://oca-account-reconcile-14-0-pr459-66f35e640d49.runboat.odoo-community.org/web#id=11&action=207&model=account.move&view_type=form&cids=&menu_id=107 (check the cash account config)

Vs this one http://oca-account-reconcile-14-0-pr459-66f35e640d49.runboat.odoo-community.org/web#id=12&action=209&model=account.move&view_type=form&cids=&menu_id=107

I think that's the correct behavior

TeoGoddet (approved):
Code and Functional review OK ! Thanks !
Maybe you should add some docs for the journal configuration and reconciliation workflows

pedrobaeza:
This is still WIP

marcelsavegnago:
@pedrobaeza Are you planning to continue with this PR?

What do you think about implementing it in a separate module? If that's the case, I can do it and submit a PR. What do you think? And if you have a name suggestion, I appreciate it.

pedrobaeza:
Still in my TODO, but I think it can be in the same module. If you can take a look, please tell us.

PR #462, opened 2022-05-17
[14.0][FIX] account_reconciliation_widget: set statement name to move ref

MiquelRForgeFlow:
When the Odoo module was ported to OCA, there was an oversight.

image

As you can see in the deleted code, the move that was being created, was using the dict values of `_prepare_reconciliation_move` method.

This `_prepare_reconciliation_move` method was this one:
Selection_388

As you can see, this method doesn't make sense now because in v14 the values of this dict are already in the move of the statement. But, except the `ref`, here is the oversight.

Thus, in order to restore this "set ref" functionality, this PR adds the method `_update_move_ref`.

### DIFF
--- a/account_reconciliation_widget/models/account_bank_statement.py
+++ b/account_reconciliation_widget/models/account_bank_statement.py
@@ -222,6 +222,7 @@ class AccountBankStatementLine(models.Model):
         aml_obj.with_context(check_move_validity=False).create(liquidity_aml_dict)
 
         self.sequence = self.statement_id.line_ids.ids.index(self.id) + 1
+        self.move_id.ref = self._get_move_ref(self.statement_id.name)
         counterpart_moves = counterpart_moves | self.move_id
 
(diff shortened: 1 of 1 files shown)


mariadforgeflow (approved):
Tested and code review. LGTM!

TeoGoddet (approved):

account_reconciliation_widget/models/account_bank_statement.py
```diff
@@ -271,6 +272,12 @@ def _create_counterpart_and_new_aml(
 
         return counterpart_moves
 
+    def _update_move_ref(self, move_ref):
```
pedrobaeza: As this is returning a value, not setting it, it should be `_get_move_ref`

MiquelRForgeFlow: done

pedrobaeza (approved):

PR #480, opened 2022-07-05
[FIX] account_move_reconcile_forbid_cancel: make it compatible with tests of other modules

ernestotejeda:
TT37518

- Make it compatible with tests of other modules
- Fix payable account type detection and add tests for vendor invoice

### DIFF
--- a/account_move_reconcile_forbid_cancel/models/account_move.py
+++ b/account_move_reconcile_forbid_cancel/models/account_move.py
@@ -10,7 +10,7 @@ class AccountMove(models.Model):
 
     def _get_receivable_payable_lines(self):
         return self.line_ids.filtered(
-            lambda l: l.account_internal_type in ["receivable", "acc_type_pay"],
+            lambda l: l.account_internal_type in ["receivable", "payable"],
         )
 
     def button_draft(self):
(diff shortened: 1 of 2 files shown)


ernestotejeda:
Ping @carlosdauden @pedrobaeza

account_move_reconcile_forbid_cancel/models/account_move.py
```diff
@@ -10,17 +10,25 @@ class AccountMove(models.Model):
 
     def _get_receivable_payable_lines(self):
         return self.line_ids.filtered(
-            lambda l: l.account_internal_type in ["receivable", "acc_type_pay"],
+            lambda l: l.account_internal_type in ["receivable", "payable"],
```
pedrobaeza, line 13: Why this change?

ernestotejeda: This is another fix. I have extracted it to another commit.

ernestotejeda:
I splitted the commit into 2 commit to have each functionality in a different commit.
ping @pedrobaeza @carlosdauden

pedrobaeza (approved):

pedrobaeza:
Fast-tracking this to fix the bug and the integration tests:

/ocabot merge patch

PR #487, opened 2022-08-01
[14.0] account_reconciliation_widget: fix notifications, show analytic and add ability to edit 'ref'

alexis-via:
With this PR, 'ref' of account.move can now be customized in the reconciliation widget and 'name' of account.move is displayed for information.

One of my users asked me for this ; he wanted this for his classification/archive methodology. I decided to propose this development here: if you think it's useful for the module, we can review/merge it. If you think this feature is not useful for the module, I'll maintain this branch for that specific user and we cancel the PR.

reconcile_widget-edit_ref

### DIFF
--- a/account_reconciliation_widget/models/account_bank_statement.py
+++ b/account_reconciliation_widget/models/account_bank_statement.py
@@ -245,6 +245,9 @@ class AccountBankStatementLine(models.Model):
             if aml_dict["move_line"].partner_id.id:
                 aml_dict["partner_id"] = aml_dict["move_line"].partner_id.id
             aml_dict["account_id"] = aml_dict["move_line"].account_id.id
+            aml_dict["analytic_account_id"] = (
+                aml_dict["move_line"].analytic_account_id.id or False
+            )
 
             counterpart_move_line = aml_dict.pop("move_line")
             new_aml = aml_obj.with_context(check_move_validity=False).create(aml_dict)
--- a/account_reconciliation_widget/models/account_bank_statement.py
+++ b/account_reconciliation_widget/models/account_bank_statement.py
@@ -222,7 +222,6 @@ class AccountBankStatementLine(models.Model):
         aml_obj.with_context(check_move_validity=False).create(liquidity_aml_dict)
 
         self.sequence = self.statement_id.line_ids.ids.index(self.id) + 1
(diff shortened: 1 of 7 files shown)


pedrobaeza:
The screenshot is of the "after the PR"? Isn't the labels unaligned?

alexis-via:
Yes, the screenshot is after the PR ; as you can see, the "Reference" field appear at the top and there is the move number next to it.

The label alignement problem at the bottom of the screen is unrelated. I reduced the screen width to make the screenshot ; on a regular PC screen, the fields are spread access 2 columns, so you don't see this alignment problem.

pedrobaeza:
OK, I see. Not sure if this should go in the main module. Let see others what think.

alexis-via:
IMHO:

Pros : 

- adds a possibility for accountants

Cons:

- 67 more lines of code to maintain. But this module has 10k lines of code (5k if you exclude tests), so it's not very significant.

Remark:

- without this patch, the 'ref' field contains the 'name' of the account.bank.statement, which, in real life, doesn't contain any useful information. With this PR, the 'ref' field will be empty by default ; it will have a value only if the accountant sets the value in the JS reconciliation interface.

andreampiovesana (approved):
LGTM

alexis-via:
Another killer feature of this PR: notifications when starting the reconcile interface are working again (they were working on v10 and, according to my tests, broken in v12 and v13). Here is an example on v14 with this PR when importing an OFX file that had 7 transactions that were already imported in a previous OFX file:

bank_statement_v14-notifications

clementmbr (approved):
LGTM! Thanks

alexis-via:
I also added the display of the analytic account on the counterpart line. Before that, when you had several counter-part lines, you could not see the analytic account you had set on the previous counterpart lines.

bank-statement-rec-show_analytic

I display the analytic account between the due date and the label. If you think another position would be better, please suggest and explain why you think it's better... I'm open to improvement !

More details tomorrow after a good night of sleep :)

njeudy (approved):
Great job I just tested it and think accountant will love it :) 👍

pedrobaeza:
Can you please fw-port it to v15?

pedrobaeza:
@alexis-via please fw-port it to v15

PR #489, opened 2022-08-03
[14.0][IMP] account_reconcile_reconciliation_date: reconciliation date should not be copied

cvinh:
Steps to reproduce : 
- duplicate a journal entry that has been reconciled
- reconciliation date is copied

Right behaviour 
- reconciliation date should be empty


pedrobaeza (approved):
Let's merge it for not losing this patch:

/ocabot merge patch

PR #491, opened 2022-08-10
[14.0] account_reconciliation_widget: dev. status=mature

alexis-via:
We've been deploying and using account_reconciliation_widget on odoo v14 for quite some time now, so maybe we can set it to mature ?


pedrobaeza:
Not sure about complying with all the requirements to be Mature:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/oca_module_lifecycle_development_status.rst#mature-modules

alexis-via:
ok, no pb

PR #492, opened 2022-08-11
[13.0][FIX] account_mass_reconcile: Compare threshold level properly

pedrobaeza:
Not using `float_compare` can make this to produce incorrect results.

@Tecnativa TT38529


grindtildeath (approved):
:+1:

PR #500, opened 2022-11-01
[ADD] account_reconcile_oca

etobella:
It is still a Draft project
Todo:
- [x] Bank statement reconcile widget
  - [x] Manual reconcile
  - [x] Invoice reconcile
  - [x] Configure the domain of invoices
  - [x] Chatter from the move
  - [x] Improve how to force a different amount on an invoice
  - [x] Launch invoice match automatically
  - [x] Launch writeoff match automatically
  - [x] Reconcile models buttons
  - [x] Remove a line (discard button)
  - [x] Use amount format according to currency
  - [x] Access reconcile widget from Move
  - [x] Adding a button for adding new lines
  - [x] Select next line when reconciling (& refresh!)
  - [x] To check functionality. It is interesting, isn't it?
- [x] Improve creation of statements 
- [x] Account reconcile
- [x] Readme
- [x] Review what happens with multi-companies
- [x] Review what happens with multi-currencies --> Add a test for that
- [x] Tests
- [x] Add an option in order to not delete original move lines and create a new move.
- [x] refreshing is a bit confusing.... Improve it somehow

Detected bugs:

- [ ] Chatter does not update when we add an activity
- [x] Name of the reconcile widget is of the selected move (maybe it is right for us)

I will try to add a lot of commits. And squash them at the end.


elvise:
@etobella that’s amazing

etobella:
@pedrobaeza @cvinh comments noted :smile:

etobella:
Well, not it begins to work properly.
image
There is still some work on navigation, but the bank reconcile widget will be completly functional soon

pedrobaeza:
It would be good to have the chatter on the bottom of the right part to add comments and attachments.

etobella:
@pedrobaeza Your wishes have been fullfilled :rofl:

pedrobaeza:
You're the best!

etobella:
On the bank reconcile widget, the only missing important thing is that it is not refreshing automatically once we press the reconcile button. Anyway, it is usable if you want to try it :smile:

elvise:
@etobella when i try to create a new activity, it doesn’t appear in the chatter

etobella:
Nice catch!! Let me review why 😁

etobella:
After checking how mail works, I was unable to reload the activity data. It might be possible, but I think too much JS just for that :cry:

etobella:
Well, I think Bank reconcile widget is ready for testing. There is already a bug detected on the chatter, but I don't know how to fix it :cry:

elvise:
@etobella to simplify testing you could add Mitchell in the Technical/Show Full Accounting group ?

etobella:
Sure, group added on demo data :smile:

etobella:
@pedrobaeza Can you enlighten me? It has sense to add a reconcile model when doing conciliation of accounts? It is really weird IMO, but maybe you know a case :thinking:

pedrobaeza:
I don't have any in mind.

etobella:
So at the end, only the lines reconcile makes sense :smile:

pedrobaeza:
If it doesn't cost you too much, you can leave it in case anyone wants to do manual write-off items, but if not, go out with it.

elvise:
@etobella i’m testing and i have a quick question for you:
what happen if the user click on button “Riconcile” and after this the user realizes he has made a mistake. 🤔

etobella:
Good question.
If it is account reconcile you need to search for the reconciled lines and unreconcile them. 
If it is bank reconcile, you probably need to do some manual moves in order to fix that :cry: 

I think the same happens with previous versions or EE

OSevangelist:
Hi @etobella thanks for the massive work on this. Looks really promising and already quite complete already. However when i try to enter a new reconcile record i get the following error. Do you have an idea what is wrong with it 
Odoo - Reconcile - Create

etobella:
Button shouldn't be there. Sorry :cry: 

It is gone right now. in order to reconcile accounts, the system must find that some unreconciled lines for the same account exists, some on debit and some on credit.

OSevangelist:
Thanks for fixing @etobella I can confirm that the button is no longer there. Will continue my functional testing

elvise:
Hey @etobella ! 
i would like to port your great module in V14, please let me know if i can help in this PR and when i can start the task for V14 :)

vdewulf:
Module tested on v16.
From the Dashboard, I clicked on the bank account journal, 3 dots on the right corner, Views > Transactions, and there I was able to create manually a new transaction (I suppose I'll need to also test the import statement later on).
I was not able to create a new statement from the statement list (after clicking on the journal name).
One created (from a transaction, I created a linked Statement), I could not see and edit it (clicking on the line with the list of Statements does nothing).

The reconciliation screen is directly available, with the list of transactions on the left and the reconcile screen on the right.
Selection_003
By default, the proposed account is a 49 suspense account. But this is misleading because I thought it would use this account to reconcile if nothing is selected. But in this case, Odoo considers nothing is selected at all and refuse to reconcile.

There are 3 buttons:
- reconcile
- clean
- view move => Is it normal to see the moves even if no reconciliation has been done, yet?

Once reconcile button is clicked, the next not-reconciled transaction is displayed.

I tried to reconcile 400€ from bank account with a 300€ invoice (too much money was paid by customer). The reconcile widget shows 300€ from invoice and the remaining 100€ on the 49 suspense account.
When I click on Reconcile, this message is displayed:
"An error has occurred.
The move (BNK1/2023/00004) is not balanced.
The total of debits equals 400.00 € and the total of credits equals 300.00 €.
You might want to specify a default account on journal "BEXX" to automatically balance each move."
I don't understand why It's not considered as balanced.

This raise the following question: there is no more "partner" to be choosen. What's the impact of this?
I then added a manual line, choosing the partner and the suspense account for 100€. Even with this change, the reconcile was not possible.

I tried a partial invoice reconciliation, and it works perfectly.
I also created a statement with a label corresponding to an open invoice (and selected the Partner). The automatic matching was done.

From the menu, I went to Accounting > Actions > Reconcile and I got an error (cf. txt in attachment).
[pr500_account_reconcile_action_reconcile_error.txt](https://github.com/OCA/account-reconcile/files/10361579/pr500_account_reconcile_action_reconcile_error.txt)

But from the Dashboard, if statements are to be reconciled, the button "1 item to reconcile" is working perfectly.

This is it for my current testing.
Thanks a lot fot his work @etobella and don't hesitate to tell me if you need specific tests/scenarios.

vdewulf:
> Good question. If it is account reconcile you need to search for the reconciled lines and unreconcile them. If it is bank reconcile, you probably need to do some manual moves in order to fix that cry
> 
> I think the same happens with previous versions or EE

On this topic, in older versions (like v12), from the bank statement, you can see all the transactions. When reconciled, the line is grey instead of black and if the "account_cancel" module is installed, you can click on the rounded arrow on the right of the line to cancel the reconciliation and posting of entries.
Selection_005


This is a really useful feature.
But I suppose this needs a new module "account_cancel" (must be kept on EE now), right?

etobella:
Probably it is kept on EE, but I am not sure :thinking:

etobella:
Thanks for the comments @vdewulf I will check the errors ASAP

etobella:
@vdewulf I am unable to reproduce the error of access on my local environment or runboat. Where did you tested it?

vdewulf:
> @vdewulf I am unable to reproduce the error of access on my local environment or runboat. Where did you tested it?

@etobella it's on our test environement at Coop IT Easy, maybe some error there!

vdewulf:
@etobella I just tested your improvement on runboat, it's working, thanks for your quick reply!

>IMP] account_reconcile_oca: Hide button on bank.statement.lines if i…

…t is not balanced properly (no suspense account)

legalsylvain:
Very quick test. Step to reproduce.
- go to the runboat 
- Go to accouning > Dashboard.
- In "Bank" Journal, click on "Reconcile x items"

<details><summary>KeyError: 'can_reconcile'</summary>Traceback (most recent call last):
  File "/opt/odoo/odoo/api.py", line 983, in get
    cache_value = field_cache[record._ids[0]]
KeyError: 6

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/fields.py", line 1138, in __get__
    value = env.cache.get(record, self)
  File "/opt/odoo/odoo/api.py", line 990, in get
    raise CacheMiss(record, field)
odoo.exceptions.CacheMiss: 'account.bank.statement.line(6,).can_reconcile'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/api.py", line 983, in get
    cache_value = field_cache[record._ids[0]]
KeyError: 6

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/fields.py", line 1138, in __get__
    value = env.cache.get(record, self)
  File "/opt/odoo/odoo/api.py", line 990, in get
    raise CacheMiss(record, field)
odoo.exceptions.CacheMiss: 'account.bank.statement.line(6,).reconcile_data_info'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/http.py", line 1579, in _serve_db
    return service_model.retrying(self._serve_ir_http, self.env)
  File "/opt/odoo/odoo/service/model.py", line 134, in retrying
    result = func()
  File "/opt/odoo/odoo/http.py", line 1608, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "/opt/odoo/odoo/http.py", line 1805, in dispatch
    result = self.request.registry['ir.http']._dispatch(endpoint)
  File "/opt/odoo/odoo/addons/base/models/ir_http.py", line 144, in _dispatch
    result = endpoint(**request.params)
  File "/opt/odoo/odoo/http.py", line 698, in route_wrapper
    result = endpoint(self, *args, **params_ok)
  File "/opt/odoo/addons/web/controllers/dataset.py", line 42, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/opt/odoo/addons/web/controllers/dataset.py", line 33, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/opt/odoo/odoo/api.py", line 461, in call_kw
    result = _call_kw_multi(method, model, args, kwargs)
  File "/opt/odoo/odoo/api.py", line 448, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/opt/odoo/odoo/models.py", line 2994, in read
    return self._read_format(fnames=fields, load=load)
  File "/opt/odoo/odoo/models.py", line 3132, in _read_format
    vals[name] = convert(record[name], record, use_name_get)
  File "/opt/odoo/odoo/models.py", line 5851, in __getitem__
    return self._fields[key].__get__(self, type(self))
  File "/opt/odoo/odoo/fields.py", line 1187, in __get__
    self.compute_value(recs)
  File "/opt/odoo/odoo/fields.py", line 1346, in compute_value
    records._compute_field_value(self)
  File "/opt/odoo/odoo/models.py", line 4198, in _compute_field_value
    field.compute(self)
  File "/opt/odoo/addons/base_sparse_field/models/fields.py", line 52, in _compute_sparse
    values = record[self.sparse]
  File "/opt/odoo/odoo/models.py", line 5851, in __getitem__
    return self._fields[key].__get__(self, type(self))
  File "/opt/odoo/odoo/fields.py", line 1187, in __get__
    self.compute_value(recs)
  File "/opt/odoo/odoo/fields.py", line 1346, in compute_value
    records._compute_field_value(self)
  File "/opt/odoo/odoo/models.py", line 4196, in _compute_field_value
    getattr(self, field.compute)()
  File "/mnt/data/odoo-addons-dir/account_reconcile_oca/models/account_bank_statement_line.py", line 251, in _compute_reconcile_data_info
    record.can_reconcile = record.reconcile_data_info["can_reconcile"]
KeyError: 'can_reconcile'

The above server error caused the following client error:
null</details>


Kind regards, and thanks for this work !

alexis-via:
I'm currently working on bank statement import on v16 and on the reconcile interface (big thanks to Eric Tobella for this great first version), and I have the feeling that, in addition to the many details of feature/user interface, there are 2 big topics that need to be adressed, so that we have some consistency among all the OCA module related to bank statements:

**1) double interface of bank statements**
Up to v15, we had two user interface for bank statements:
- the "traditional" interface : tree+form views defined in XML in the "account" module
- the "reconcile" interface available in "account_reconciliation_widget" which consisted of 3k lines of JS code.
Each interface had its specific feature/possibilities and users switched between the 2 interfaces.

With the new module account_reconcile_oca for v16 written by Eric, there is only 300 lines of JS code and a significant part of the "reconcile" interface is defined in XML, so easy to modify interface without advanced JS skills. In v16, the "account" module doesn't provide the tree+form XML view for account.bank.statement.line ; it only provides the tree view of account.bank.statement (form view is not provided).

In v16, should we stop using the "traditional" interface for bank statements and have all the features in the "reconcile" interface ?
My opinion: I think it would be nice to only use one interface, the "reconcile" interface. But it has several implications. For example:
- account_statement_import_file would depend on account_reconcile_oca ? I would not need the "glue" module account_statement_import_file_reconciliation_widget any more.
- we need to be able to edit and delete a bank statement line from the "reconcile" interface.
- we need to be able to access a detailed form view of the bank statement line ?

**2) do we still need the account.bank.statement object ?**
The new "reconcile" interface can work with account.bank.statement.line that are not attached to any account.bank.statement. In v16, when you delete an account.bank.statement, its lines are not deleted (no more ondelete="cascade" !). Since v10, in akretion's usability module for accounting, I added an option to hide the fields balance_start and balance_end_real of account.bank.statement and disable the check of balance_end_real and I always enable that option. Many bank statement file don't provide the end balance (including the OFX file, which doesn't provide the end balance but TODAY's balance) anyway. For me, you have to compare the balance of your bank account that you can see on your online banking with the balance of the default_account_id of your bank journal and it's the only serious and reliable way to do it.

In the end, if you don't use balance_start nor balance_end_real, the other fields of account.bank.statement are useless. The only useful feature provided by account.bank.statement is the ability to attach the bank statement file ! :) But we can find another way to do it: for example, we could add a M2O to ir.attachment on account.bank.statement.line that would be set by account_statement_import_file.

The impact of this is mainly in the code of account_statement_import_file and account_statement_import_online.

In addition to these 2 big topics, there are some details that I haven't seen in the current PR:

a) how do you manually set the partner on a bank statement line (if it hasn't been auto-set by the bank statement import modules) ?

b) how do you manually reconcile 2 account.move.lines in v16 ? The module account_reconciliation_widget provided that feature under the menu "Accounting > Accounting > Journal Items", select the lines to reconcile, then "Action > Reconcile". In fact, I don't think it is the role of the module account_reconcile_oca to provide this feature: I would profer to have it in a separate module. If there is not already an OCA initiative to provide this feature, I propose to do it myself. I propose to call the module "account_manual_reconcile" (or "account_move_line_manual_reconcile") under OCA/account-reconcile. The interface would be similar to the wizard account.move.line.reconcile that existed up to Odoo v11, cf https://github.com/odoo/odoo/blob/11.0/addons/account/wizard/account_reconcile_view.xml

c) when you made a mistake while reconciling a bank statement line, you need to be able to cancel/redo (already discussed by @vdewulf)

etobella:
> **1) double interface of bank statements** Up to v15, we had to user interface for bank statements:

I prefer to keep a single one. We might add all the necessary hooks and improvements there in order to make it easier for users
 
> **2) do we still need the account.bank.statement object ?** The new "reconcile" interface can work with account.bank.statement.line that are not attached to any account.bank.statement. In v16, when you delete an account.bank.statement, its lines are not deleted (no more ondelete="cascade" !). Since v10, in akretion's usability module for accounting, I added an option to hide the fields balance_start and balance_end_real of account.bank.statement and disable the check of balance_end_real and I always enable that option. Many bank statement file don't provide the end balance (including the OFX file, which doesn't provide the end balance but TODAY's balance) anyway. For me, you have to compare the balance of your bank account that you can see on your online banking with the balance of the default_account_id of your bank journal and it's the only serious and reliable way to do it.

I think we might need to keep it just as an aggregation entity. At the end we receive statements.
 
> a) how do you manually set the partner on a bank statement line (if it hasn't been auto-set by the bank statement import modules) ?

I see your point, I think the best would be to select the statement line (first line) and edit the partner there (similar to odoo)

> b) how do you manually reconcile 2 account.move.lines in v16 ? The module account_reconciliation_widget provided that feature under the menu "Accounting > Accounting > Journal Items", select the lines to reconcile, then "Action > Reconcile". In fact, I don't think it is the role of the module account_reconcile_oca to provide this feature: I would profer to have it in a separate module. If there is not already an OCA initiative to provide this feature, I propose to do it myself. I propose to call the module "account_manual_reconcile" (or "account_move_line_manual_reconcile") under OCA/account-reconcile. The interface would be similar to the wizard account.move.line.reconcile that existed up to Odoo v11, cf https://github.com/odoo/odoo/blob/11.0/addons/account/wizard/account_reconcile_view.xml

There is a menu Invocing / Accounting / Actions / Reconcile that finds all the possible reconciliable lines.

> c) when you made a mistake while reconciling a bank statement line, you need to be able to cancel/redo (already discussed by @vdewulf)

We probably need to add something there, but it might be a problem in some cases (we added the option to not modify lines, just create a new one. On that case, I am not sure if this will work...)

legalsylvain:
(partial thought, I am not an expert in accounting)

> 1) double interface of bank statements

:+1: to have a single interface.

> 2) do we still need the account.bank.statement object
> [...]
> The only useful feature provided by account.bank.statement is the ability to attach the bank statement file ! :) But we can find another way to do it: for example, we could add a M2O to ir.attachment on account.bank.statement.line that would be set by account_statement_import_file.

well, the model `account.bank.statement` exists in odoo CE, so i don't see the interest to not use it, and to add a M2O of ir.attachment. But maybe I missed something.

Thanks to all for all this great work and reflections.

rvalyi:
one more remark: if many OCA modules start depending on account_reconcile_oca, probably they won't play nice with Odoo EE out of the box, or let's say it might require some extra work. Personally I find it "fair enough", but this is a double edged sword as some captive partners may stop using them.

etobella:
It might happen, but if we do the modules undependant we might need glue modules with account_reconcile_oca. Also, they will not work properly with EE, so I would prefer to make them related :thinking:

rvalyi:
in general I would say it would be fair enough that OCA modules depend on account_revoncile_oca and that the glue modules burden would be for the ones ""opting"" for Odoo EE.

alexis-via:
@rvalyi Thanks for pointing out this argument related to Odoo EE, I didn't have it in mind.

After some thoughts, I propose the following plan:
1) we move the views of account.bank.statement.line in the OCA module account_usability https://github.com/OCA/account-financial-tools/tree/16.0/account_usability
2) account_reconcile_oca depends on account_usability
3) account_statement_import_base depends on account_usability
4) we have a glue module "account_statement_import_file_reconcile_oca" between account_reconcile_oca and account_statement_import_file (same as in v14 with the glue module account_statement_import_file_reconciliation_widget).
With this plan, we have 0 additional module compared to v14. If you agree with the plan, I'll do the PR on account_usability to add the views of account.bank.statement.line.

Are you OK with my plan?

pedrobaeza:
No, I don't want such overhead with a generic module adding a lot of stuff and more with the name `usability`. If anything is moved, it should be to a specific module `account_statement_base` or similar.

alexis-via:
Yes, my plan is stupid because Odoo EE users cannot install account_usability.

So here is v2 of my plan:
1) create a new module account_statement_base (name proposed by Pedro) that adds the views of account.bank.statement.line
2) account_reconcile_oca depends on account_statement_base
3) account_statement_import_base depends on account_statement_base
4) we have a glue module "account_statement_import_file_reconcile_oca" between account_reconcile_oca and account_statement_import_file (same as in v14 with the glue module account_statement_import_file_reconciliation_widget).
With this plan, we have only 1 additional module compared to v14. If you agree with the plan, I'll do the PR on OCA/account-reconcile to add the module account_statement_base with the views of account.bank.statement.line.

With this plan, both Odoo CE and Odoo EE users should be happy.

pedrobaeza:
I don't have any interest in making enterprise users happy. Enterprise introduces incompatibilities everywhere with OCA, so it's the time for people to decide between both. Why community people have to bother and get complicated on enterprise? If there's enterprise people interested, they should be the one adding the layers.

What I would do is to rename this module to account_statement_oca and be the base for the rest.

legalsylvain:
> No, I don't want such overhead with a generic module adding a lot of stuff and more with the name usability. If anything is moved, it should be to a specific module account_statement_base or similar.

hi. As the contributor of `account_usability` module, I'd like to add my opinion on that topic. 
- The objective of this module is to add all the menu entries that are missing in odoo CE. (and maybe also in Odoo EE sometimes). Also it adds some simple views, if it doesn't exist. 
- It replace the previous situation where there was a lot of minimal modules that was adding just an entry. The module was named `account_menu` before V16. [See 12.0 PR.](https://github.com/OCA/account-financial-tools/pull/855)
- So for me the module must not contain a "lot of stuff, more and more". No extra "features". Just xml view / actions / menu and some trivial fixes. (like the one that fixes the [ugly hack](https://github.com/odoo/odoo/blob/16.0/addons/account/models/res_users.py#L38) of Odoo that makes the accountant group invisible in Odoo CE)
- Maybe the name is incorrect. (usability is maybe too generic). This module **should be kept simple**, and very easy to migrate.

> Yes, my plan is stupid because Odoo EE users cannot install account_usability.

I don't get it. Does it make sense to install the account_reconcile_oca module on an EE version anyway?

legalsylvain:
Regarding the other topic (see below), I answer here, but I think we should move that talk in a separate thread to avoid poluting this great PR. @oca/board : could you create an issue in a dedicated place ?

> I don't have any interest in making enterprise users happy. Enterprise introduces incompatibilities everywhere with OCA, so it's the time for people to decide between both. Why community people have to bother and get complicated on enterprise? If there's enterprise people interested, they should be the one adding the layers.

Factual things :
- OCA members doens't have all access to EE. OCA members have all access to CE.
- OCA infrastructure CI run only on CE.
- Ensure OCA modules works properly with both odoo CE and Odoo EE versions requires _sometimes_ extra work, especially when the module concerns a privatized functionality. (accounting, web, etc...)
- Sometimes compatibility is functionally impossible because the designs are different.

-> As a result, OCA can not guarantee that OCA modules works properly with EE.

For a while now I see problematic questions to OCA contributors like: 
- Will your module work with Odoo EE ? ([Ref](https://github.com/OCA/account-fiscal-rule/pull/327#discussion_r1062468409))
- Is the code of your module different enough from the Odoo EE code ? ([Ref](https://github.com/OCA/web/pull/2351#issuecomment-1335315781)). (I'm not talking about copyright infringement here, but about redevelopment of a privatized feature.)

For OCA members that don't have access to EE Code, these questions don't make sense, and they have no way to answer them.

**Proposal A**
- Stop asking to OCA members questions regarding EE compatilities. Otherwise, the OCA will send a bad signal to full Open Source players, who will gradually stop contributing.

- If OCA modules are incompatible with Odoo EE.
    - solution 1 (minimal) : When identified, add a text in the ROADMAP.rst section. ("_this module has incompatibilities if it is installed on an enterprise version_")
    - solution 2 : the changes required to ensure compatibility are trivial and doesn't add complexity : Add a patch to make the module compatible
    - Solution 3 : The changes  required to ensure compatibility are NOT trivial and / or add complexity : Develop a dedicated module `module_ee`  to make the module working with OCA module and EE, and exclude the module from the CI. (as it is not testable).
**All that work should be done by EE users**. 

- add a disclaimer in the README.rst template that mention "_This module has been tested on a continuous integration server with an Odoo CE version. It is not guaranteed that the module is not incompatible with the EE version._"

**Proposal B**
- require contributors to make modules compatible with the EE version
- For that purpose, change OCA CI to make the tests running on CE and EE
- assume that people that doesn't have access to EE will stop contributing
- Finally, remove little by little the word "OpenSource" from the communication of the OCA (website, etc...)

For those who know me, I am obviously more in favor of proposal A, and the proposal B is little ironic, but @pedrobaeza is right in my opinion and a direction must be taken and communication should be explicit on this subject.

pedrobaeza:
Very interesting topic indeed, @legalsylvain, but I don't want to pollute this thread more, and should be handled in other place, but we are clearly inline and we (as OS contributors) are shooting our own foot in a lot of occasions.

Coming again to this pull request, I believe more now that we should rename this module to `account_statement_oca` and remain as is and being the base for any other thing (like statement import, if it proceeds). It can be moved to OCA/account-financial-tools, or finally create a repository OCA/account-statement that gathers all the account-reconcile and bank-statement-* repos.

alexis-via:
My PR on this PR to add a dependency on account_statement_base: https://github.com/tegin/account-reconcile/pull/1 (EDIT: replaced by PR https://github.com/tegin/account-reconcile/pull/2) 

Independently from the Odoo CE/EE debate, in this particular case, I think it would be overkill to oblige account_statement_import_base to depend on account_reconcile_oca just for some views of account.bank.statement.line. So I really think that having a small module account_statement_base that contains the missing view is the best option.

legalsylvain:
> Independently from the Odoo CE/EE debate, in this particular case, I think it would be overkill to oblige account_statement_import_base to depend on account_reconcile_oca just for some views of account.bank.statement.line. So I really think that having a small module account_statement_base that contains the missing view is the best option.

Yes it seems that in that case, the "solution 2" mentionned in my previous comment (trivial patch) is possible.

alexis-via:
@legalsylvain yes, I confirm we are in the trivial scenario here.

AntoniRomera:
How is this module going, can I help in any way?

pedrobaeza:
You can test it and give your review.

account_reconcile_oca/static/src/js/reconcile_controller.esm.js
```diff
@@ -0,0 +1,99 @@
...
+export class ReconcileController extends KanbanController {
+    async setup() {
+        super.setup();
+        this.state = useState({
+            selectedRecordId: null,
+        });
+        useSubEnv({
+            parentController: this,
+            exposeController: this.exposeController.bind(this),
+        });
```
AntoniRomera:
```suggestion
        });
        this.env.exposeController(this);
```
You should call "this.env.exposeController(this);" because when you enter in the reconciliation menu the default view is the kanban view that uses this controller and not the form one, so if you don't call it you will get an error that this.form_controller is undefined.

account_reconcile_oca/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_account_account_reconcile,account.account.reconcile,model_account_account_reconcile,account.group_account_user,1,1,0,0
+access_account_account_reconcile_data,account.account.reconcile,model_account_account_reconcile_data,account.group_account_user,1,1,0,0
```
AntoniRomera: It must be possible to delete and create the records of "account.account.reconcile.data", otherwise the clear button gives an error.
```suggestion
access_account_account_reconcile_data,account.account.reconcile,model_account_account_reconcile_data,account.group_account_user,1,1,1,1
```

AntoniRomera (changes requested):
After a quick functional review without going into too much detail there are these two things, When it comes to reconciling it seems to do it correctly, at the expense of changes proposed by @alexis-via 

Tested on a local machine, runbot gives a 500.

florian-dacosta:
After some tests, here are the issues I met : 

1) Tried on local instance and in runboat and I have the "can_reconcile" keyerror Sylvain Legal talked about here : https://github.com/OCA/account-reconcile/pull/500#issuecomment-1377037050

I am surprised nobody else talked about this error, it seems I have it, if I click on the button "Clean Data" on a line which can't be reconciled by default.
On runboat, you have the error on the "Last Year Interests" bank statement line.

From what I see, if a statement line can't be reconciled and then we click on "clean data", here : https://github.com/tegin/account-reconcile/blob/16.0-add-account_reconcile_oca/account_reconcile_oca/models/account_bank_statement_line.py#L370
It will write False on can_reconcile field, which is a sparse field. So it will trigger the inverse method to write on the serialized field.
Since it wants to write is `False`, Odoo will actually remove the key from the serialized field : https://github.com/OCA/OCB/blob/16.0/addons/base_sparse_field/models/fields.py#L69

I don't understand why this line is usefull : https://github.com/tegin/account-reconcile/blob/16.0-add-account_reconcile_oca/account_reconcile_oca/models/account_bank_statement_line.py#L370
Since the previous line already update the serialized field `reconcile_data` which contain the `can_reconcile` data already.
Removing this line fixes it for me.

Anyway, I guess a best approach to fix the bug would be to avoid accessing the can_reconcile data the way it is done now : `self.reconcile_data_info["can_reconcile"]` but rather check like this : self.reconcile_data_info.get("can_reconcile", False)
It seems best to consider that the value is False if the key is missing in the serialized field, since it seems it is the way Odoo handle the empty/false value when writing directly on the sparse field.

2) I have created a reconcile model of type "Rule to suggest counterpart entry" and I have check the box "Auto-validate" (`auto_reconcile`).
My bank statement line which match this rule has an automatic counter part suggestion, but it is not reconciled automatically, I still have to click on the button reconcile.

Big thanks for this big work

etobella:
> After some tests, here are the issues I met :
> 
> 1. Tried on local instance and in runboat and I have the "can_reconcile" keyerror Sylvain Legal talked about here : [[ADD] account_reconcile_oca #500 (comment)](https://github.com/OCA/account-reconcile/pull/500#issuecomment-1377037050)

It should be fixed now :smile: 

> 2. I have created a reconcile model of type "Rule to suggest counterpart entry" and I have check the box "Auto-validate" (`auto_reconcile`).

I have done a change to do the match automatically on statement line creation. It should not work when we set the value to true and the line was already created. It seems logic to me

account_reconcile_oca/security/ir.model.access.csv
etobella: True!

account_reconcile_oca/static/src/js/reconcile_controller.esm.js
etobella: It is called on the ReconcileFormController :thinking:

AntoniRomera: Yep, but when you enter the view to reconcile, we are in kanban and does not jump only the function in ReconcileFormController, I tried it on my machine locally, if you want I'll try it again and pass you the video.

florian-dacosta:
Thanks @etobella, it seems to work fine now indeed.
Some other remarks : 
1) It is possible to put manual values for account, amount, label for the liquidity line. But when we reconcile, these values are ignored on the account move lines (which is normal). I guess these manual fields should be readonly or invisible for the liquidity line, to avoid a lot of confusion for the user, what do you think ?
Capture d’écran du 2023-02-20 12-38-19

2) If we change the partner of the statement line, we have to put the manual partner for the liquidity lines, but the partner is not updated on the left part of the screen. Maybe it is a minor issue, I don't know if it is easy to solve or not ?
Capture d’écran du 2023-02-20 12-36-02

3) On manual vals tab, if I set a manual account, then remove it, it crashes with error : 
```
  File "/opt/odoo/addons/web/controllers/dataset.py", line 42, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/opt/odoo/addons/web/controllers/dataset.py", line 33, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/opt/odoo/odoo/api.py", line 461, in call_kw
    result = _call_kw_multi(method, model, args, kwargs)
  File "/opt/odoo/odoo/api.py", line 448, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/opt/odoo/odoo/models.py", line 6489, in onchange
    record._onchange_eval(name, field_onchange[name], result)
  File "/opt/odoo/odoo/models.py", line 6200, in _onchange_eval
    method_res = method(self)
  File "/mnt/data/odoo-addons-dir/account_reconcile_oca/models/account_bank_statement_line.py", line 225, in _onchange_manual_reconcile_vals
    "account_id": self.manual_account_id.name_get()[0],
IndexError: list index out of range
```
It comes from https://github.com/tegin/account-reconcile/blob/16.0-add-account_reconcile_oca/account_reconcile_oca/models/account_bank_statement_line.py#L225
I guess suspense account should be put back in this case.

etobella:
Thanks @florian-dacosta 

Comments attended

florian-dacosta:
Thanks @etobella !

1. About the readonly manual fields for the liquidity line, I think we should also put `manual_amount` and `manual_name` readonly since it has no effect on the account move line during the reconciliation. (same domain as the manual_account_id)

2. About the partner, when it is updated on the bank statement line, it still does not appear on the left side of the screen (the kanban part). It will only appear if you leave the reconciliation widget and enter once again, because then the view is refreshed.
(Still not very important IMO but nice if easy to implement)

3. fixed indeed, I was thinking to put back the suspense account in case the manual account is set to False. Here it goes to "undefined", I guess it is ok too, I don't have opinion about this, the important beeing to avoid the odoo crash which is done !

4. When you set a partner on the liquidity line, it updates the partner of the bank statement line, but I think it should also update the `manual_partner_id` of the suspense line.
Indeed, it is the default behavior, if you open for the first time a bank statement line which as the `partner_id` set, with the widget, the partner_id is set on the suspense account line.
Also if you set the partner on the statement line and click on the "clean" button, the partner will be set.

For me, the only important missing feature missing the the possibility to revert a reconciliation, in case of mistake.
About this, you said : 

> We probably need to add something there, but it might be a problem in some cases (we added the option to not modify lines, just create a new one. On that case, I am not sure if this will work...)

Are you talking about the journal option "Reconcile Mode" = "Keep suspense accounts" ?
Maybe we could start implementing it only for the "Reconcile Mode" = "Edit move" option ?
I am not sure if I missing something about this part, I imagine a revert button that would do something like :
```
account.move.button_draft
account.move.line_ids.unlink
Recreates the lines with account.bank.statement.line._prepare_move_line_default_vals
account.move.action_post
```
I could make a PR about that if you are interested

etobella:
> 1. About the readonly manual fields for the liquidity line, I think we should also put `manual_amount` and `manual_name` readonly since it has no effect on the account move line during the reconciliation. (same domain as the manual_account_id)

About the name, you are right, but it is different for amount, as we might want to do a partial reconcile :smile: 

> 2. About the partner, when it is updated on the bank statement line, it still does not appear on the left side of the screen (the kanban part). It will only appear if you leave the reconciliation widget and enter once again, because then the view is refreshed.
>    (Still not very important IMO but nice if easy to implement)

The change is applied on a change of view. For example, if you select another record, the change is applied. I don't know if there is a better way for this :sad:

> 3. fixed indeed, I was thinking to put back the suspense account in case the manual account is set to False. Here it goes to "undefined", I guess it is ok too, I don't have opinion about this, the important beeing to avoid the odoo crash which is done !
> 4. When you set a partner on the liquidity line, it updates the partner of the bank statement line, but I think it should also update the `manual_partner_id` of the suspense line.
>    Indeed, it is the default behavior, if you open for the first time a bank statement line which as the `partner_id` set, with the widget, the partner_id is set on the suspense account line.
>    Also if you set the partner on the statement line and click on the "clean" button, the partner will be set.

I am not sure about this. I mean, I may modify the suspense line for some reasons, so I am not sure if we should rewrite it

> For me, the only important missing feature missing the the possibility to revert a reconciliation, in case of mistake. About this, you said :
> 
> > We probably need to add something there, but it might be a problem in some cases (we added the option to not modify lines, just create a new one. On that case, I am not sure if this will work...)
> 
> Are you talking about the journal option "Reconcile Mode" = "Keep suspense accounts" ? Maybe we could start implementing it only for the "Reconcile Mode" = "Edit move" option ? I am not sure if I missing something about this part, I imagine a revert button that would do something like :
> 
> ```
> account.move.button_draft
> account.move.line_ids.unlink
> Recreates the lines with account.bank.statement.line._prepare_move_line_default_vals
> account.move.action_post
> ```
> 
> I could make a PR about that if you are interested

I will add the functionality! Don't worry :smile:

etobella:
@florian-dacosta Unreconcile has been implemented. Until some work is missing in order to add all necessary tests. However, I hope it is almost ready :smile:

AntoniRomera:
https://jam.dev/c/b9b10f8c-7335-49f0-aac7-aef7c23781b3

alexis-via:
I propose 2 small fixes/improvements: https://github.com/tegin/account-reconcile/pull/3

alexis-via:
Thanks for this great work on the reconcile interface.

I'm testing the following scenario:
- For partner P1, I have an open customer invoice of 1000 €
- I have a bank statement line for P1 of 990 €
Goal : I consider that this payment of 990 € fully pays the invoice of 1000 € ; I have a write-off of 10 € to send to an expense account (this scenario regularly happens when the customer is out-of-sepa and an intermediary bank took a fee directly on the amount of the payment)
In the reconcile interface, when I select the invoice of 1000 € in the "reconcile" tab, it immediately proposes to treat it as a partial payment:

reconcile_oca_partial_payment

The problem is that there is no way to say "hey, I don't want to consider it as a partial payment, but as a full payment with a write-off" !

The workaround is to click on the bin corresponding to the move line of the invoice, go to the "manual operation" tab, enter a counter-part of 10 €, press enter, then go back to the "reconcile" tab and select the invoice of 1000 €. It is long and not convenient, and you have calculate the amount of the write-off by hand.

reconcile_oca_full_writeoff

etobella:
Thanks! I will review it, as your comments make all the sense to me. I will try to see how to make it possible with little effort

etobella:
@alexis-via Comments attended!

Thanks for your review :smile: 

@AntoniRomera Bug fixed

mileo (commented):
Hi @etobella, nice work!

Could you check this return: 

https://github.com/tegin/account-reconcile/blob/16.0-add-account_reconcile_oca/account_reconcile_oca/models/account_bank_statement_line.py#L357

Fresh install with demo data:

image

etobella:
Sorry! Should be fixed now

zuher83:
Hi @etobella thanks, nice work
The tax option included in the reconciliation models does not work properly. It does not generate accounting entries related to the tax. It right when the amount is without tax but if you choose a tax included the result is not right.

screen-reconciliation-widget
screen-reconciliation-model

alexis-via:
@etobella I tested it: it works very well, great work ! Thank you !

etobella:
One question, should we need to keep the "To Check" Functionality?

etobella:
@zuher83 Can you test it again?

alexis-via:
I have a number of different info/thoughts/remarks/suggestions, some of them following my work on OCA/bank-statement-import today:

1) I re-implemented the notification system directly in the module account_statement_import_file using ir.actions.client/display_notification, cf https://github.com/OCA/bank-statement-import/pull/575
It's not as powerful as the previous notification system that was present in account_reconciliation_widget (no more link to the auto-deleted duplicate statement lines), but I think it's good enough. The advantage is that it is simple and doesn't have any impact on other modules.

2) In account_statement_import_file, we used to attach the imported file to the bank statement for diagnostic purposes. But, in v16, there is no more form view of the bank statement, so we can't easily access the attachment. And, in v16, account.bank.statement doesn't inherit from mail.thread (but it has a new M2M field "attachment_ids"). We could restore the form view of bank statement in account_statement_base... or maybe we should find a new idea to store the bank statement file in Odoo (but I haven't found this new idea yet !).

3) There is no way to import account.bank.statement.line via CSV/XLS using the standard import feature of Odoo via "Favorites > Import records". In the special reconcile view of account_reconcile_oca, we don't have the menu "Favorites"... is it on purpose ?

4) In v16, when you delete a bank statement, it doesn't delete the lines (there is no more ondelete=cascade on the field 'statement_id' on statement line). What do you think about it ? Should we change it ? Maybe it's linked to the fact that there is no more "state" field on bank statements and bank statement lines are directly in "posted" state, so we can't directly delete the statement lines after import anyway...

5) Vocabulary: something I have always disliked in the "reconcile" interface: the string used on the "Reconcile" button. When you process a bank statement line with a manual counterpart, you have to click on the "Reconcile" button... but you are not doing any "reconciliation" ! I think the "Reconcile" button should be named "Validate", "Process" or "Confirm" (on invoices, the label of the button is "Confirm"). I know it has been like that for so many years... but this re-implementation may be a good opportunity to improve the vocabulary!

zuher83:
> @zuher83 Can you test it again?

@etobella Now it's fixed.

alexis-via:
An important thing missing: in the "Manual operations" tab, we need an Analytic distribution field (like the Analytic Account field we had in v16). I tried to add it in this PR: https://github.com/tegin/account-reconcile/pull/4

etobella:
@alexis-via I did the analytics too, sorry, I didn't see your PR :cry:

etobella:
> 3. There is no way to import account.bank.statement.line via CSV/XLS using the standard import feature of Odoo via "Favorites > Import records". In the special reconcile view of account_reconcile_oca, we don't have the menu "Favorites"... is it on purpose ?

I might try to find a way to do that, but it is really necessary? I mean, it is something that Odoo does not has on EE, and I am not sure if it used :thinking: 

> 4. In v16, when you delete a bank statement, it doesn't delete the lines (there is no more ondelete=cascade on the field 'statement_id' on statement line). What do you think about it ? Should we change it ? Maybe it's linked to the fact that there is no more "state" field on bank statements and bank statement lines are directly in "posted" state, so we can't directly delete the statement lines after import anyway...

Well, that is something that odoo changed, I don't think it is a good idea to restore that logic, as statement is only an aggregation of statement lines. I like this new approach (much simpler IMO)

> 5. Vocabulary: something I have always disliked in the "reconcile" interface: the string used on the "Reconcile" button. When you process a bank statement line with a manual counterpart, you have to click on the "Reconcile" button... but you are not doing any "reconciliation" ! I think the "Reconcile" button should be named "Validate", "Process" or "Confirm" (on invoices, the label of the button is "Confirm"). I know it has been like that for so many years... but this re-implementation may be a good opportunity to improve the vocabulary!

Done!

etobella:
I made some JS improvements in order to improve the UX of the widgets. I hope they are good enough now :smile: 

Once we are sure that we can merge it, I will squash several commits in order to make it more logic :laughing: Keeping the history until now it was a good approach IMO :smile:

AntoniRomera:
### Functional Review
Captura de pantalla de 2023-03-13 09-02-24

Steps (tested in runbot):
1. Change the account field
2. And switch back to 121000
3. Add lines
4. Try to "Reconcile"

etobella:
@AntoniRomera Can you try again

AntoniRomera:
I try again in runbot, when I change the account I still get the same error, sorry I don't have more information, I don't get the complete message and in the browser log there is nothing.

etobella:
OK! Don't worry, I will review that again ASAP :wink:

etobella:
@AntoniRomera I was able to test your issue. Account should never be changed :sob: Now it is made readonly by default, so the error is avoided! Thanks!!!

AntoniRomera:
Okay, fine

AntoniRomera:
Ready to merge?

ljsalvatierra-factorlibre:
Hi, i am wondering the OCA position on this `[IMP] account: Optional bank statements`

- https://github.com/odoo/odoo/pull/99092

I need help to understand the motivations behind it and if OCA have any intentions of implementing it back.

I believe there is no intention on going back but i would like to know why if possible.

etobella:
Odoo has not removed them, just made them optional. It makes sense to me to keep as it is. It makes sense to have a statement when you are importing data from files. Otherwise, it makes no sense IMO. If you want it required, you can always propose a module.

It is probably ready to merge, but some more reviews are welcome

gerard-vacas:
About what @ljsalvatierra-factorlibre is commenting, in the current v16, after this development, we can find the statement as an object wich has statement lines attached and also every statement line has a statement assigned, but after all, statements are not usable as in previous versions (although they are accessible through other multiple click ways that take you to a not user friendly view).

What we have been discussing is that having the opportunity of doing some of the next operations on statements would be interesting:
- To be able to access to it's (previous verions look like) form view to be able to:
  - Define a reference for them.
  - Be able to delete any line at the beggining of an odoo implantation, to delete lines that might be conciled in the previous system and would stand there forever and, in consequence, to be able to modify the ending balance. There could be other reasons to do so, too.
  - Access to the reconcile interface for it's own lines using a domain, through it's form view, so if you have multiple pending reconciliations from other statements, they will not always be shown. It's true that it already exists a filter for statement in the reconcliation interface, but nowadays is useless as they have no reference and even if they have one, it's not very intuitive having to know it. Take in consideration that some companies might have 1k+ operations in a week for an account.

We also have doubts about being able to create statements. As we consider it might be useful in some cases, it could not be indispensable if you base their use on imports/synchronization.

As you said, it could be better to consider another module like account_bank_statement_usability or something like this, but i wanted to add this here to hear your opinions.

---

About the usability of this module, in a reconcile POV, it LGTM, thanks for the great job

etobella:
It is more suitable for another module like `account_statement_base` as this module expect to create all necessary widgets for reconcile, while the other one is trying to reintroduce all info from statements.

pedrobaeza:
Going to test the module, if I click on the journal dashboard on "RECONCILE N ITEMS", I get this error:

```
  File "/mnt/data/odoo-addons-dir/account_reconcile_oca/models/account_bank_statement_line.py", line 321, in _compute_reconcile_data_info
    record.reconcile_data_info = record._default_reconcile_data()
  File "/mnt/data/odoo-addons-dir/account_reconcile_oca/models/account_bank_statement_line.py", line 414, in _default_reconcile_data
    *self._reconcile_data_by_model(
  File "/mnt/data/odoo-addons-dir/account_reconcile_oca/models/account_bank_statement_line.py", line 348, in _reconcile_data_by_model
    for line in reconcile_model._get_write_off_move_lines_dict(
  File "/opt/odoo/addons/account/models/account_reconcile_model.py", line 523, in _get_write_off_move_lines_dict
    if currency.is_zero(balance):
UnboundLocalError: local variable 'balance' referenced before assignment
```

Please squash commits as well for proceeding with the merge when everything else is OK.

ljsalvatierra-factorlibre:
> Going to test the module, if I click on the journal dashboard on "RECONCILE N ITEMS", I get this error:
> 
> ```
>   File "/mnt/data/odoo-addons-dir/account_reconcile_oca/models/account_bank_statement_line.py", line 321, in _compute_reconcile_data_info
>     record.reconcile_data_info = record._default_reconcile_data()
>   File "/mnt/data/odoo-addons-dir/account_reconcile_oca/models/account_bank_statement_line.py", line 414, in _default_reconcile_data
>     *self._reconcile_data_by_model(
>   File "/mnt/data/odoo-addons-dir/account_reconcile_oca/models/account_bank_statement_line.py", line 348, in _reconcile_data_by_model
>     for line in reconcile_model._get_write_off_move_lines_dict(
>   File "/opt/odoo/addons/account/models/account_reconcile_model.py", line 523, in _get_write_off_move_lines_dict
>     if currency.is_zero(balance):
> UnboundLocalError: local variable 'balance' referenced before assignment
> ```
> 
> Please squash commits as well for proceeding with the merge when everything else is OK.

Hi, i opened this issue last week

- https://github.com/odoo/odoo/issues/116064

etobella:
> Going to test the module, if I click on the journal dashboard on "RECONCILE N ITEMS", I get this error:
> 
> ```
>   File "/mnt/data/odoo-addons-dir/account_reconcile_oca/models/account_bank_statement_line.py", line 321, in _compute_reconcile_data_info
>     record.reconcile_data_info = record._default_reconcile_data()
>   File "/mnt/data/odoo-addons-dir/account_reconcile_oca/models/account_bank_statement_line.py", line 414, in _default_reconcile_data
>     *self._reconcile_data_by_model(
>   File "/mnt/data/odoo-addons-dir/account_reconcile_oca/models/account_bank_statement_line.py", line 348, in _reconcile_data_by_model
>     for line in reconcile_model._get_write_off_move_lines_dict(
>   File "/opt/odoo/addons/account/models/account_reconcile_model.py", line 523, in _get_write_off_move_lines_dict
>     if currency.is_zero(balance):
> UnboundLocalError: local variable 'balance' referenced before assignment
> ```
> 
> Please squash commits as well for proceeding with the merge when everything else is OK.

I fixed it yesterday. The error comes from an odoo function that is not well defined. It should work now.

etobella:
Squashed all commits in 4.

It should work now :smile:

zuher83 (approved):
Great job 🙏😉

account_reconcile_oca/README.rst
```diff
@@ -0,0 +1,96 @@
...
+Bank reconcile
+~~~~~~~~~~~~~~
+
+Access `Invoicing / Dashboard` with a user with Full Acounting capabilities.
+Select reconcile on the journal of your choice.
+
+Account reconcile
+~~~~~~~~~~~~~~~~~
+
+Access `Invoicing / Accounting / Actions / Reconcile`
```
LoisRForgeFlow, line 47: In this view I see an empty list:

image

However from the banks journal it is fine:

image

LoisRForgeFlow (approved):
Functional review :+1:

This module is more than ready to be merged IMO. I just have a small non-blocking comment below.

Amazing job @etobella

account_reconcile_oca/README.rst
etobella: The first one is empty as you don't have any accounts move lines to reconcile, Both widgets are similar but different, as the reconcile accounts is not able to change the move lines as the account.bank.statement

AaronHForgeFlow (commented):
Thank you @etobella !!!

zuher83:
Thanks, @etobella 🙏🎉

PR #514, opened 2023-01-18
[16.0][ADD] module account_statement_base

alexis-via:
This module contains the views of account.bank.statement.line, which have been removed from the account module in Odoo v16

Related to https://github.com/OCA/account-reconcile/pull/500

### DIFF
--- a/account_statement_base/views/account_bank_statement_line.xml
+++ b/account_statement_base/views/account_bank_statement_line.xml
@@ -1,0 +1,149 @@
+<?xml version="1.0" ?>
+<!--
+  Copyright 2022-2023 CreuBlanca
+  Copyright 2023 Akretion France (http://www.akretion.com/)
+  @author: Alexis de Lattre <alexis.delattre@akretion.com>
+  Licence LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0).
+-->
+<odoo>
+
+<record id="account_bank_statement_line_form" model="ir.ui.view">
+    <field name="model">account.bank.statement.line</field>
+    <field name="arch" type="xml">
+        <form>
+            <group name="main">
+                <field name="company_id" invisible="1" />
+                <field name="currency_id" invisible="1" />
+                <field name="country_code" invisible="1" />
+                <field name="state" invisible="1" />
+                <field name="suitable_journal_ids" invisible="1" />
+                <field name="date" />
+                <field name="payment_ref" required="1" />
+                <field name="partner_id" />
+                <field name="amount" />
+                <field
+                        name="foreign_currency_id"
+                        domain="[('id', '!=', currency_id)]"
+                        groups="base.group_multi_currency"
+                    />
+                <field name="amount_currency" groups="base.group_multi_currency" />
+                <field name="partner_bank_id" />
+                <field
+                        name="journal_id"
+                        invisible="context.get('default_journal_id')"
+                    />
+                <field name="move_id" />
+                <field
+                        name="statement_id"
+                        domain="[('journal_id', '=', journal_id)]"
+                        invisible="not context.get('account_bank_statement_line_main_view')"
+                    />
+            </group>
+            <notebook>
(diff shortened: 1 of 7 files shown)


account_statement_base/views/account_bank_statement_line.xml
```diff
@@ -0,0 +1,158 @@
...
+            <filter
+                    name="reconciled"
+                    string="Reconciled"
+                    domain="[('is_reconciled', '=', True)]"
+                />
+            <filter
+                    name="not_reconciled"
+                    string="Not Reconciled"
+                    domain="[('is_reconciled', '=', False)]"
+                />
```
rvalyi, line 118: I think the v15 search view has a few more handy filters: what about adding narration and amount? https://github.com/odoo/odoo/blob/15.0/addons/account/views/account_bank_statement_views.xml#L279

account_statement_base/views/account_bank_statement_line.xml
```diff
@@ -0,0 +1,158 @@
...
+                        <field name="transaction_type" />
+                        <field name="partner_name" />
+                        <field name="account_number" />
+                        <field name="is_reconciled" />
+                        <field name="payment_ids" widget="many2many_tags" />
+                    </group>
+                </page>
+            </notebook>
+            <footer invisible="not context.get('add_bank_statement_line')">
+                <button name="save" type="object" string="Save" class="btn-primary" />
```
rvalyi: as there is no save method on the object, it seems you should use a Javascript "special" attribute, something like (ripped from v15)
```
                        <button string="Confirm" class="btn-primary" special="save" data-hotkey="v"/>
                        <button string="Cancel" class="btn-secondary" special="cancel" data-hotkey="z"/>
```

see failing test https://github.com/OCA/account-reconcile/actions/runs/3952659874/jobs/6768008598#step:8:64

rvalyi (changes requested):
kudos for the work, small suggestions here...

account_statement_base/views/account_bank_statement_line.xml
rvalyi: on a second thought, as the account.bank.statement.line form is not meant to be edited in a bank statement anymore, shouldn't you just remove this footer?

alexis-via: You're right. I did it yesterday on my laptop, but I forgot to push it to github before going to bed, sorry.

legalsylvain:
Regarding the name, account_bank_statement_line_view could be more explicit, dont you think ?or did you planned to increase the scope of this module ?

alexis-via:
@legalsylvain For the moment, it only contains view, but maybe we'll add more things in it in the future. The goal of this module is to have the common code/views between reconcile interface and statement import. I used the name suggested by Pedro.

account_statement_base/views/account_bank_statement_line.xml
alexis-via: I added search on "amount". Search on "narration" is already included in the first filter, see "filter_domain"

rvalyi (approved):

florian-dacosta (approved):
LGTM code review

AntoniRomera (approved):

bguillot (approved):

Gelo-fl:
Hi @dreispt,

This module is merged, however the translations are locked:
[account-reconcile-16.0-account_statement_base](https://translation.odoo-community.org/projects/account-reconcile-16-0/account-reconcile-16-0-account_statement_base/)

Can I know why?

Thanks in advance,

Gelo-fl:
> Hi @dreispt,
> 
> This module is merged, however the translations are locked: [account-reconcile-16.0-account_statement_base](https://translation.odoo-community.org/projects/account-reconcile-16-0/account-reconcile-16-0-account_statement_base/)
> 
> Can I know why?
> 
> Thanks in advance,

Please,
Could someone please unblock the translation of this module?

pedrobaeza:
Unblocked

PR #515, opened 2023-01-31
[15.0][FIX] account_reconcile_widget: analytic tags formatting

pedrobaeza:
Forward-port of #418

Before this commit. analytic tags are not correctly formatted when sent to the server after reconciliation, that lead to an error being raised.

Steps to reproduce the problem:

- Create a reconciliation model of type "Manually create a write-off on clicked button".
- Show analytic on the configuration and show the optional column "Analytic tags".
- Put any analytic tag.

When using such reconciliation model on the reconciliation, the analytic tag is not well formatted and you can't complete the reconciliation.


pedrobaeza:
Tested on runboat to check if still working the patch in v15, and it's correct, so merging as well to finish the patch chain:

/ocabot merge patch

PR #444, opened 2022-02-18
[14.0][FIX] account_reconciliation_widget: in case of manual operation with…

jonerikceberio:
… analytic account, create the analytic entry if the move is posted

PR #521, opened 2023-02-09
[15.0][FIX] account_reconciliation_widget: Analytic in manual operation

pedrobaeza:
Forward-port of #444

In case of a manual operation with analytic account/tags, create the analytic entry(ies) if the move is posted.

@Tecnativa


pedrobaeza:
@omar7r @jonerikceberio is it OK this on v15?

pedrobaeza:
Merging for now, as not apparent changes:

/ocabot merge patch

Issue #522, opened 2023-02-14
[15.0] [account_reconciliation_widget] Foreign Currencies in Bank Statement Line is not recognize


singwangho:
## Module
account_reconciliation_widget

## Describe the bug
If the bank account is configured to multi-currency (i.e. the currency for the account is not set), when using reconcile, the widget's "bank statement line" erroneously returns the foreign currency as company's currency.  Reconciliation is not possible.

## To Reproduce
**Odoo15**:

The behavior is reproduced with demo data.  Steps to reproduce the behavior:
1. Load demo data
2. Load the modules: om_accountant_account and account_reconciliation_widget)
3. Add new bank account
4. Add a new bank statement with line information: Label="Receive Euro 50", Amount Currency=50.0; Foreign Currency=EUR; Amount=54.0
_5. For clarity, I have changed currency symbol of USD from "$" to "US$"._
6. Click "Post" and then "Reconcile'
7. Note in the statement line amount it show US$50.00 instead of EUR50.  Note when rollover the 'i' the Amount shown is also incorrect.

**Expected behavior**
For the statement line amount, I expect the line to show line's currenct which is EUR50.  And that it is reconciling against EUR50 instead of US50.

**Additional context**
N/A

singwangho:
Hi, I did some digging and found that this module "account_reconciliation_widget" doesn't read in AccountBankStatementLine.**foreign_currency_id** at all.

So generally it doesn't really support individual line's foreign currencies...

Below edits _models/reconciliation_widget.py_ should give the minimum support for the 'foreign_currency_id' field in statement line.

### Suggested FIX
diff account_reconciliation_widget/models/reconciliation_widget.py
1021,1023c1021,1023
\<         if st_line.amount_currency and st_line.currency_id:
\<             amount = st_line.amount_currency
\<             amount_currency = st_line.amount
\---
\>         if st_line.amount_currency and st_line.foreign_currency_id:
\>             amount = st_line.amount
\>             amount_currency = st_line.amount_currency
1025c1025
\<                 self.env, abs(amount_currency), currency_obj=statement_currency
\---
\>                 self.env, abs(amount_currency), currency_obj=st_line.foreign_currency_id

whotopia:
I have the exact same issue for Odoo 14 CE.  
I am trying to do a 7000 GBP transfer from GBP account to CHF Account.   
To do this in Odoo 14, I perform 2 transfer operations. One for each account via the transfers account.  
[...]

PR #437, opened 2022-01-26
[15.0][MIG] account_reconciliation_widget: Migration

kelvzxu:

PR #531, opened 2023-03-03
[15.0][FIX] account_reconciliation_widget: foreign currency in bank statement line handling

singwangho:
In reference to an earlier bug report: #522 

This is a bug fix which covers a rare case where multiple currencies are used in the same bank account.

**Bank Statement Line Error Demonstration**
image
**Before FIX - reconciliation is not possible at this stage due to wrong currency recognized by widget**
image
**After FIX**
image

### DIFF
--- a/account_reconciliation_widget/models/reconciliation_widget.py
+++ b/account_reconciliation_widget/models/reconciliation_widget.py
@@ -1023,11 +1023,11 @@ class AccountReconciliation(models.AbstractModel):
         st_line_currency = (
             st_line.foreign_currency_id or st_line.currency_id or statement_currency
         )
-        if st_line.amount_currency and st_line.foreign_currency_id:
+        if st_line.amount_currency and (st_line_currency != statement_currency):
             amount = st_line.amount
             amount_currency = st_line.amount_currency
             amount_currency_str = formatLang(
-                self.env, abs(amount_currency), currency_obj=st_line.foreign_currency_id
+                self.env, abs(amount_currency), currency_obj=st_line_currency
             )
         else:
             amount = st_line.amount
--- a/account_reconciliation_widget/models/reconciliation_widget.py
+++ b/account_reconciliation_widget/models/reconciliation_widget.py
@@ -1020,11 +1020,14 @@ class AccountReconciliation(models.AbstractModel):
         statement_currency = (
             st_line.journal_id.currency_id or st_line.journal_id.company_id.currency_id
         )
-        if st_line.amount_currency and st_line.currency_id:
-            amount = st_line.amount_currency
-            amount_currency = st_line.amount
+        st_line_currency = (
+            st_line.foreign_currency_id or st_line.currency_id or statement_currency
+        )
+        if st_line.amount_currency and st_line.foreign_currency_id:
+            amount = st_line.amount
+            amount_currency = st_line.amount_currency
             amount_currency_str = formatLang(
-                self.env, abs(amount_currency), currency_obj=statement_currency
+                self.env, abs(amount_currency), currency_obj=st_line.foreign_currency_id
             )
         else:
             amount = st_line.amount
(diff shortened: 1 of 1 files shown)


account_reconciliation_widget/models/reconciliation_widget.py
```diff
@@ -1020,11 +1020,14 @@ def _get_statement_line(self, st_line):
         statement_currency = (
             st_line.journal_id.currency_id or st_line.journal_id.company_id.currency_id
         )
-        if st_line.amount_currency and st_line.currency_id:
-            amount = st_line.amount_currency
-            amount_currency = st_line.amount
+        st_line_currency = (
+            st_line.foreign_currency_id or st_line.currency_id or statement_currency
```
pedrobaeza, line 1024: Why `st_line.currency_id` would be different from `statement_currency`?

pedrobaeza: This question is not answered

singwangho: In older versions of Odoo (e.g. `13.0`), `st_line.currency_id` is actually used as to allow different currency for individual bank statement line!  `st_line.foreign_currency_id` does not exists.  So it maybe possible that from migration `st_line.currency_id` is different from `statement_currency`.

**However** from `14.0` onwards, `st_line.foreign_currency_id` is introduced and `st_line.currency_id` is re-defined as "Journal Currency" according to the main code (`14.0/addons/account/models/account_bank_statement.py`).  So `st_line.currency_id` should be identical to `statement_currency`.

I've modified the code to be more resilient against `st_line.currency_id` being mis-used as a foreign currency arising from legacy issues.

I do recommend applying this FIX to `14.0` as well.

Thanks for reviewing!  Btw, I don't know why pre-commit failed...

pedrobaeza: Then, according to your comment, the problem is in the migration of your DB with such mixed data, not something reproducible in current version. You should fix your DB instead of changing the module.

singwangho: Hi, please allow me to clarify.  Above screenshots of the bug and fix are all taken from 15.0 environment.

This is not a DB issue because:
1. The bug is reproducible on fresh demo-data;
2. This is not an isolated case as a similar issue was reported in [CybroOdoo/CybroAddons#246](https://github.com/CybroOdoo/CybroAddons/issues/246) and a similar proposed fix [CybroOdoo/CybroAddons#177](https://github.com/CybroOdoo/CybroAddons/issues/177) -- thanks to [whotopia](https://github.com/whotopia) for letting me know he encountered the similar problem.

I probably confused you when I was trying to explain why `st_line.currency_id` should be same as `statement_currency` and went on to explain why the bug happened.

Anyways let me try again.  This commit aims to add basic support for `foreign_currency_id` so multi-currency bank statements can be re-conciliated with the widget.

**_`st_line.currency_id` should be same as `statement_currency`_**
Below is code from Odoo 15.0's main repository where `st_line.currency_id` is defined, `currency_id` is the Journal Currency which is statement's currency.
image


**_What caused this bug exists?_**
This widget code was copied from older Odoo versions.  When Odoo introduced `foreign_currency_id` in 14.0 --- this widget did not get properly updated.

As evident from this image, you can see pre-14.0 versions, `currency_id` is used for handling multi-currency.
image

singwangho: Seems like the problem happened during #437 [15.0][MIG] account_reconciliation_widget:Migration to 15.0:

In #437 [15.0][MIG] account_reconciliation_widget:Migration to 15.0:

The commit "https://github.com/OCA/account-reconcile/pull/437/commits/62cd65ca8fe22670c62b3546c38bdfa270e78a4e Fix foreign_currency_id vs currency_id on statement" actually fixes the issue.

But then the commit is over-written in https://github.com/OCA/account-reconcile/pull/437/commits/7f6d1a2ec6d0361479a53d03676208c197458a9c ... not sure why.

Odoo does use `foreign_currency_id` in `15.0` evidently from:
https://github.com/odoo/odoo/blob/15.0/addons/account/models/account_bank_statement.py see Line 549

pedrobaeza (approved):
Thanks for the explanations. I confirm the bug introduced on the migration and the proposed fix is correct. Let's merge it for fixing the problem:

/ocabot merge patch

PR #539, opened 2023-04-03
[16][FIX] account_reconcile_oca: Fix account_account_reconcile id to avoi…

florian-dacosta:
…d big int

I have 2 issues with current implementation, the account_account_reconcile query fails because the returned id is too big (big int).
Also, I have a case of an account with id 1 containing account move lines with no partner which give 0 as id which cause an access error on the reconcile widget.
From what I understand, we just want a unique ID and it seems this make the job.

What do you think @etobella ?
The database where I have the issue is quite small, with max partner id = 97519 and max account id 1125

### DIFF
--- a/account_reconcile_oca/models/account_account_reconcile.py
+++ b/account_reconcile_oca/models/account_account_reconcile.py
@@ -36,13 +36,7 @@ class AccountAccountReconcile(models.Model):
     def _select(self):
         return """
             SELECT
-                CAST(
-                    (
-                        coalesce(aml.partner_id, 0) + a.id
-                    )*(
-                        COALESCE(aml.partner_id, 0)+a.id - 1
-                    )/2 + COALESCE(aml.partner_id, 0) AS INTEGER
-                ) as id,
+                min(aml.id) as id,
                 MAX(a.name) as name,
                 aml.partner_id,
                 a.id as account_id,


etobella:
The problem with that is that navigation will not work properly

The way I proposed might have problems with overfloat... I didn't notice that. Maybe I need to make it work somehow with two parameters

florian-dacosta:
> The problem with that is that navigation will not work properly

What do you mean ? I did not see issues, but I did not test a lot.

account_reconcile_oca/models/account_account_reconcile.py
```diff
@@ -36,13 +36,7 @@ def _table_query(self):
...
         return """
             SELECT
-                CAST(
-                    (
-                        coalesce(aml.partner_id, 0) + a.id
-                    )*(
-                        COALESCE(aml.partner_id, 0)+a.id - 1
-                    )/2 + COALESCE(aml.partner_id, 0) AS INTEGER
-                ) as id,
+                ROW_NUMBER() OVER (ORDER BY 1) as id,
```
etobella: Why don't we use min(aml.id) :thinking: Locally worked perfectly, If you have a bigger Database to test, that would be great :smile:

etobella:
The row number is not immutable enough, if you click on the partner, when you return, the row number might have changed :cry: Also, the id is used for storing the configuration on a transient model, if it is not immutable, something wrong might happen:

Imagine you have 2 reconcile items, select the second one, Set some data, but don't reconcile, go to the first one and reconcile it, the second one previous data will not be restored :cry:

account_reconcile_oca/models/account_account_reconcile.py
florian-dacosta: It seems to work like a charm! Not sure why I did not think of it, thanks !

florian-dacosta:
Thanks for the clarification, indeed we need a fix id instead of a random one.

etobella (approved):
Fast tracking

/ocabot merge patch

PR #540, opened 2023-04-04
[16][account_reconcile_oca] take payment ref as counterpart bank line ref

florian-dacosta:
Today, the reference by default on the bank counterpart is the number of the accounting entry, which is no help for accountant to know what it is about.
Once the statement line is reconciled, when searching the account move lines from me, or in the reports (general ledger...) you may want to see the information that you have on your bank statement line.

what do you think @etobella ?

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -183,7 +183,7 @@ class AccountBankStatementLine(models.Model):
                     and self.partner_id.name_get()[0]
                     or False,
                     "date": fields.Date.to_string(self.date),
-                    "name": self.name,
+                    "name": self.payment_ref or self.name,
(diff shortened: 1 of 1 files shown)


account_reconcile_oca/models/account_bank_statement_line.py
```diff
@@ -183,7 +183,7 @@ def _recompute_suspense_line(self, data, reconcile_auxiliary_id, manual_referenc
                     and self.partner_id.name_get()[0]
                     or False,
                     "date": fields.Date.to_string(self.date),
-                    "name": self.name,
+                    "name": self.payment_ref,
```
etobella: self.payment_ref or self.name, but you are right :smile:

etobella (approved):

kluna1998 (approved):

PR #542, opened 2023-04-08
[14.0][FIX]account_reconcile_payment_order - wrong match logic

luc-demeyer:
The code tries to match the positive account.payment.order,total_company_currency with a negative st_line.amount hence resulting in no match at all.


pedrobaeza:
Can you please add a unit test for assuring the behavior?

PR #544, opened 2023-05-11
[15.0][FIX] account_reconciliation_widget: Handle blue lines

pedrobaeza:
When coming from previous versions of Odoo, you may have payments directly done against the bank account. On the reconciliation widget, they are represented as blue lines.

One possibility is to replace in all these pending entries the bank account by the outstanding payment/receipt account, but this means to modify past accounting that may be locked.

So this commit is restoring the ability to reconcile against these blue lines, although this is a deprecated thing.

Things done:

- Repair the JS widget for informing correctly about the blue lines to reconcile.
- Cleanup.
- When having such lines to reconcile, the temporary statement line is removed, and the payment one is linked.
- When reverting reconciliation of the statement lines linked to payments, the entry is not removed, just removed the link, and a new entry is created for the statement line.

@Tecnativa TT43018

### DIFF
--- a/account_reconciliation_widget/models/account_bank_statement.py
+++ b/account_reconciliation_widget/models/account_bank_statement.py
@@ -97,10 +97,6 @@ class AccountBankStatementLine(models.Model):
         new_aml_dicts = new_aml_dicts or []
 
         aml_obj = self.env["account.move.line"]
-
-        company_currency = self.journal_id.company_id.currency_id
-        statement_currency = self.journal_id.currency_id or company_currency
-
         counterpart_moves = self.env["account.move"]
 
         # Check and prepare received data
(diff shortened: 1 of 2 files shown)


pedrobaeza:
@ozono you renamed the JS variables in https://github.com/OCA/account-reconcile/commit/aa43f5881b75615239bcb28e222df0e38319e3b4#diff-c9b44ca63aa607e18bb87e802f8e19f38534e43525703aba4078af1178e03252R953

Do you remember why?

ozono:
Well, at this point, I can't remember the specific reason... perhaps we thought it clarified what the variable was used for, or simply, it was a misunderstanding.

ozono (approved):
Without doing a full test, code seems ok. Thx.

sergio-teruel (approved):

PR #552, opened 2023-06-13
[16.0][FIX] account_mass_reconcile : fix write-off with today's date

florian-dacosta:


bguillot (approved):

florian-dacosta:
Hello @AaronHForgeFlow @ljsalvatierra-factorlibre 

Seeing you are working with `account_mass_reconcile` module, would you mind doing a review here ?
It fixes the actual/today write off date mode.
Also, display an error message when trying to check the history of an empty reconciliation run. I did it because else it was displaying a traceback error.

account_mass_reconcile/models/mass_reconcile_history.py
```diff
@@ -52,6 +52,8 @@ def _open_move_lines(self):
         :return: action to open the move lines
         """
         move_line_ids = self.mapped("reconcile_ids.reconciled_line_ids").ids
+        if not move_line_ids:
+            raise exceptions.UserError(_("No Reconciled items this time."))
```
ljsalvatierra-factorlibre: I wouldn't rise an exception.

ljsalvatierra-factorlibre (changes requested):

AaronHForgeFlow:
Ok for the fallback value on the date of the write-off. I am not sure about showing a message when there is no reconciled items. If the list is empty, it means there are no reconciled items, that is correct I think.

florian-dacosta:
@AaronHForgeFlow @ljsalvatierra-factorlibre Well... I guess you are right. I can't remember nor reproduce the issue I had with the empty reconciled items list.

I just removed this part, only fixing the reconciliation with write off at today's date here.
If I run in the other issue, I'll make a dedicated PR with a way to reproduce it on runbot.

Thanks for the reviews

account_mass_reconcile/models/base_reconciliation.py
```diff
@@ -172,7 +172,7 @@ def create_write_off(self, lines, amount, amount_curr, same_curr):
 
         move = self.env["account.move"].create(
             {
-                "date": lines.env.context.get("date_p"),
+                "date": lines.env.context.get("date_p") or fields.Date.today(),
```
ljsalvatierra-factorlibre, line 180:
```python
"date": lines.env.context.get("date_p", fields.Date.context_today(self)),
```

Please add a unit test for the use case when the date is not given.

ljsalvatierra-factorlibre (changes requested):

account_mass_reconcile/models/base_reconciliation.py
florian-dacosta: Done

AaronHForgeFlow (approved):

ljsalvatierra-factorlibre (approved):

PR #553, opened 2023-06-15
[16.0][FIX] account_reconcile_oca: account name can be translatable

marielejeune:
By default, account_account name is a char field. But if l10n_multilang is installed, it is translatable, hence the DB column is now a jsonb.
We must handle both cases in the select query.

### DIFF
--- a/account_reconcile_oca/models/account_account_reconcile.py
+++ b/account_reconcile_oca/models/account_account_reconcile.py
@@ -34,10 +34,18 @@ class AccountAccountReconcile(models.Model):
         )
 
     def _select(self):
-        return """
+        account_account_name_field = self.env["ir.model.fields"].search(
+            [("model", "=", "account.account"), ("name", "=", "name")]
+        )
+        account_name = (
+            f"a.name ->> '{self.env.user.lang}'"
+            if account_account_name_field.translate
+            else "a.name"
+        )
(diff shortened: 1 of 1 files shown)


marielejeune:
Hi @etobella , here is a fix to be able to use your module when `l10n_multilang` is installed. I am however not sure if this is the right solution.
Account may not be all translated into the current user's language.
What's your opinion about it?

etobella (approved):
It looks fine to me :wink:

account_reconcile_oca/models/account_account_reconcile.py
```diff
@@ -34,10 +34,18 @@ def _table_query(self):
         )
 
     def _select(self):
-        return """
+        account_account_name_field = self.env["ir.model.fields"].search(
+            [("model", "=", "account.account"), ("name", "=", "name")]
+        )
+        account_name = (
+            f"a.name ->> '{self.env.user.lang}'"
```
pedrobaeza, line 41: What happens if the name is not translated in the JSON? It will give an error?

marielejeune: No error, the column will be empty.

pedrobaeza (approved):
/ocabot merge patch

PR #556, opened 2023-06-17
[14.0][IMP] account_reconciliation_widget: Skip business models synchronization on reconcile

pedrobaeza:
Same as #555 

There's no need to synchronize business models (payments and statement lines) when the reconcile is done, as the only value written in the journal item is `full_reconcile_id`.

This way, we speed up the reconciliation process through the widget.

@Tecnativa


chienandalu (approved):

carlosdauden (approved):

ozono (commented):
All OK for me

PR #560, opened 2023-06-23
[15.0][FIX] account_reconciliation_widget: Show correct currency

gdrius:
When reconciling bank statement lines in a foreign currency, it shows amounts in a company's currency while the symbol is shown of foreign currency. Wrong currency appears on counterpart lines and open balance line.

The wrong currency symbol only effects representation in the UI. When reconciliation is validated, it sets company's currency on account move lines.

The first screenshot how it looks without the fix:

screenshot2

The next screenshot shows how the journal entry looks after reconciliation:

screenshot3

And the last screenshot shows how the reconciliation lines looks after the fix. Journal entry is not effected by the fix.

screenshot4

No extra migration/update steps needed for existing data.
It is a fixup for PR #531

### DIFF
--- a/account_reconciliation_widget/models/reconciliation_widget.py
+++ b/account_reconciliation_widget/models/reconciliation_widget.py
@@ -1050,7 +1050,7 @@ class AccountReconciliation(models.AbstractModel):
             "date": format_date(self.env, st_line.date),
             "amount": amount,
             "amount_str": amount_str,  # Amount in the statement line currency
-            "currency_id": st_line_currency.id,  # Currency according to statement line
+            "currency_id": (st_line.currency_id or statement_currency).id,
             "partner_id": st_line.partner_id.id,
             "journal_id": st_line.journal_id.id,
             "statement_id": st_line.statement_id.id,


account_reconciliation_widget/models/reconciliation_widget.py
```diff
@@ -1050,7 +1050,7 @@ def _get_statement_line(self, st_line):
             "date": format_date(self.env, st_line.date),
             "amount": amount,
             "amount_str": amount_str,  # Amount in the statement line currency
-            "currency_id": st_line_currency.id,  # Currency according to statement line
+            "currency_id": (st_line.currency_id or statement_currency).id,
```
pedrobaeza, line 1053: Shouldn't this be directly `statement_currency.id`? @singwangho why did you change this?

gdrius: @pedrobaeza it is used to match currency of `amount_str` (at 1041 line).

pedrobaeza (approved):

account_reconciliation_widget/models/reconciliation_widget.py
singwangho: Hi, I can replicate @gdrius issue too.

I changed it to revert this back to "#437 Fix foreign_currency_id vs currency_id" which has following code
`"currency_id": st_line.foreign_currency_id.id or statement_currency.id,`

But I think @gdrius is correct that `currency_id` should match `amount_str`.

PR #570, opened 2023-07-31
[16.0][WIP] account_reconciliation_widget: Migration to 16.0

dsolanki-initos:


pedrobaeza:
Why not using the existing enterprise-like new reconciliation widget: https://github.com/OCA/account-reconcile/tree/16.0/account_reconcile_oca ?

etobella:
@dsolanki-initos We created a new module from scratch, so this module should be no longer necessary. Also, it is done in OWL, this one is done with a legacy system and we don't know when we will lose it. Can you check the new module?

AlexPForgeFlow:
Hi! I've seen that this module could be necessary or the account_reconcile_oca should be improved. When you try to reconcile not from bank statements but from accounts, not all options from the bank reconciliation are available.

From bank statements:

image

From chart of accounts:

* The first issue is that the system won't allow you to reconcile when you only have debits or credits. A blank screen appears:
image
This could be improved changing the query of #L77 of model account_account_reconcile.py.
image

*
image


In addition, there's an important feature allowing you to initiate the reconciliation widget when you have selected several journal items from the same account, making it possible to reconcile them (or write off).
image

josep-tecnativa:
Please, include these changes in the migration https://github.com/OCA/account-reconcile/pull/601/commits/d490df5e19eafcf68af6200dbee351ac79b58508

pedrobaeza:
Instead of keeping this old system, the missing features should be included in the new reconciliation widget.

PR #583, opened 2023-08-31
[14.0][IMP] account_reconciliation_widget: added option to filter all accont move lines

ValentinVinagre:
Added option to filter all accont move lines from the date indicated in 'account_bank_reconciliation_start' field.

### DIFF
--- a/account_reconciliation_widget/models/reconciliation_widget.py
+++ b/account_reconciliation_widget/models/reconciliation_widget.py
@@ -805,6 +805,22 @@ class AccountReconciliation(models.AbstractModel):
             ("balance", "!=", 0.0),
             ("parent_state", "=", "posted"),
         ]
+        if (
+            st_line.company_id.account_bank_reconciliation_start_all_aml
+            and st_line.company_id.account_bank_reconciliation_start
+        ):
+            domain_matching = expression.AND(
+                [
+                    domain_matching,
+                    [
+                        (
+                            "date",
+                            ">=",
+                            st_line.company_id.account_bank_reconciliation_start,
+                        )
+                    ],
+                ]
+            )
 
         domain = expression.OR([domain_reconciliation, domain_matching])
         if partner_id:
--- a/account_reconciliation_widget/models/res_company.py
+++ b/account_reconciliation_widget/models/res_company.py
@@ -12,3 +12,6 @@ class ResCompany(models.Model):
         "for some time and don't want to reconcile all the past payments with "
         "bank statements.",
     )
+    account_bank_reconciliation_start_all_aml = fields.Boolean(
+        string="Filter all account move lines"
+    )
--- a/account_reconciliation_widget/readme/CONTRIBUTORS.rst
+++ b/account_reconciliation_widget/readme/CONTRIBUTORS.rst
@@ -1,1 +1,6 @@
-* Tecnativa - Pedro M. Baeza
+* `Tecnativa <https://www.tecnativa.com>`__:
+
+  * Pedro M. Baeza
+* `Sygel <https://www.sygel.es>`__:
+
+  * Valentin Vinagre
--- a/account_reconciliation_widget/wizards/res_config_settings.py
+++ b/account_reconciliation_widget/wizards/res_config_settings.py
@@ -11,3 +11,6 @@ class ResConfigSettings(models.TransientModel):
     account_bank_reconciliation_start = fields.Date(
         related="company_id.account_bank_reconciliation_start", readonly=False
     )
+    account_bank_reconciliation_start_all_aml = fields.Boolean(
+        related="company_id.account_bank_reconciliation_start_all_aml", readonly=False
+    )
--- a/account_reconciliation_widget/wizards/res_config_settings_views.xml
+++ b/account_reconciliation_widget/wizards/res_config_settings_views.xml
@@ -1,31 +1,35 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <!--
   Copyright 2022 Akretion France (http://www.akretion.com/)
+  Copyright 2023 Valentin Vinagre <valentin.vinagre@sygel.es>
   @author: Alexis de Lattre <alexis.delattre@akretion.com>
   License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 -->
 <odoo>
 
-<record id="res_config_settings_view_form" model="ir.ui.view">
-    <field name="model">res.config.settings</field>
-    <field name="inherit_id" ref="account.res_config_settings_view_form" />
-    <field name="arch" type="xml">
-        <xpath expr="//div[@id='bank_cash']" position="inside">
-            <div
+    <record id="res_config_settings_view_form" model="ir.ui.view">
+        <field name="model">res.config.settings</field>
+        <field name="inherit_id" ref="account.res_config_settings_view_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//div[@id='bank_cash']" position="inside">
+                <div
                     class="col-xs-12 col-md-6 o_setting_box"
                     id="account_bank_reconciliation_start"
                 >
-                <div class="o_setting_left_pane" />
-                <div class="o_setting_right_pane">
-                    <label for="account_bank_reconciliation_start" />
-                    <div class="text-muted">
-                    Date from which you started using bank statements in Odoo. Leave empty if you've been using bank statements in Odoo from the start.
+                    <div class="o_setting_left_pane" />
+                    <div class="o_setting_right_pane">
+                        <label for="account_bank_reconciliation_start" />
+                        <div class="text-muted">
+                        Date from which you started using bank statements in Odoo. Leave empty if you've been using bank statements in Odoo from the start.
+                        </div>
+                        <field name="account_bank_reconciliation_start" />
+                        <br />
+                        <field name="account_bank_reconciliation_start_all_aml" />
+                        <label for="account_bank_reconciliation_start_all_aml" />
                     </div>
-                    <field name="account_bank_reconciliation_start" />
                 </div>
-            </div>
-        </xpath>
-    </field>
-</record>
+            </xpath>
+        </field>
+    </record>
 
 </odoo>


ValentinVinagre:
@HaraldPanten

HaraldPanten (approved):
Functional review. Tested in local environment and runboat.

HaraldPanten:
@etobella that might be interesting for you 😉

pedrobaeza:
Not sure about this. Ignoring unreconciled entries doesn't seem the best strategy: they will still persist in open entries report, views, etc.

HaraldPanten:
> Not sure about this. Ignoring unreconciled entries doesn't seem the best strategy: they will still persist in open entries report, views, etc.

Hi @pedrobaeza , the objective of this improvement is only improving a filter in the reconciliations view (the filter already exists, we are not creating any extra feature). The usability is quite similar as it was in V12, for example, when the reconciliation view was in Odoo Community. It's not about "balancing" the ledger, it's just about filtering a view.

There are some situations where users are unable to reconcile account entry lines, such as when accounting is balanced (because payments have been entered manually introducing account moves), but users have not reconciled account move lines in several years.

In this case, the situation arises in which there are hundreds of thousands account move lines with pending reconciliation, so... when users start reconciling properly, this view is impossible to manage.

Do you know what I mean?

pedrobaeza:
My comments were just for the use case you are describing.

HaraldPanten:
> My comments were just for the use case you are describing.

Yeap, but showing hundreds of thousands of old account move lines makes impossible to start working properly, reconciling the items as users should do (but haven't done). It helps for "tiding the accounting" when you need to start working well.

Not everyone makes "their homework". You know...

As I said, the same feature was in previous versions and it was never a problem.

ValentinVinagre:
This is the original Odoo code in v.12. In case it can help to review the functionality.

**The field:**
https://github.com/odoo/odoo/blob/f7431b180834a73fe8d3aed290c275cc6f8dfa31/addons/account/models/company.py#L60-L62

**The filter:**
https://github.com/odoo/odoo/blob/f7431b180834a73fe8d3aed290c275cc6f8dfa31/addons/account/models/account_reconcile_model.py#L449-L451

etobella (approved):
If it was already present on odoo, I don't see it as a problem

pedrobaeza (changes requested):
You are mixing things. The filter in v12 is for payments (blue lines), and that filter is already present in this module. What you pretend is to filter everything, not only blue lines.

ValentinVinagre:
> 

Sorry bad link in the last comment.
What happened is that when the module was transferred to oca, the date filter criteria was changed (as you can see in the link). In OCA it currently only filters based on the "domain_reconciliation" domain and not both domains as it was in Odoo originally.
So, we recover that functionality while respecting the OCA functionality.

https://github.com/odoo/odoo/blob/f7431b180834a73fe8d3aed290c275cc6f8dfa31/addons/account/models/reconciliation_widget.py#L526-L527

etobella:
Well, After reading original code from odoo (v12) and the current code), I see we moved the filter from the total domain to only one part of it:

As you can see, the filter was applied to domain https://github.com/odoo/odoo/blob/12.0/addons/account/models/reconciliation_widget.py#L526-L528

On the following PR https://github.com/OCA/account-reconcile/pull/443, the code was moved changing the behavior.

IMO, we should move  the code as it was originally and keep part of the PR code: https://github.com/OCA/account-reconcile/pull/443/files#diff-00ca0391f0efd17a399a50f2dca6af6f06a32a2456f722c4f3a01b1edcdb4e91L771-L779

WDYT @pedrobaeza  or did you move this code for a reason I am unable to see?

HaraldPanten:
With our changes we keep the current functionality, but we offer the other option (the original one). So... we are not "breaking" anything nor offering "new bad praxis" to the users.

pedrobaeza:
That domain is used only for blue lines:

Selección_018

And the same for 13.0:

https://github.com/odoo/odoo/blob/13.0/addons/account/models/reconciliation_widget.py#L72-L81

etobella:
@pedrobaeza that's true, but that domain contains an OR https://github.com/OCA/account-reconcile/blob/14.0/account_reconciliation_widget/models/reconciliation_widget.py#L809

On #443, the condition was moved from all the domain to just the first part of that domain (the first condition of the OR). It should affect all of it.

HaraldPanten:
That's it, Enric 👍

pedrobaeza:
The intention since the beginning has been to apply to blue lines, although the code in some moments doesn't reflect it, and the fact that Odoo itself is not allowing this "masking" shows that it's a bad idea.

As said, you will hide these reconciliations from the reconciliation widget, but open items in `account_financial_report` will reflect them, the journal items view filtered by "Non reconciled" will show these elements, etc, so it's tricking only one part of the system.

HaraldPanten:
> The intention since the beginning has been to apply to blue lines, although the code in some moments doesn't reflect it, and the fact that Odoo itself is not allowing this "masking" shows that it's a bad idea.
> 
> As said, you will hide these reconciliations from the reconciliation widget, but open items in `account_financial_report` will reflect them, the journal items view filtered by "Non reconciled" will show these elements, etc, so it's tricking only one part of the system.

Hi Pedro, sorry but I don't agree. It was not since the beginning. At the beginning it was like a "copy" of the old reconciliation features in previous versions.

On the other hand, these entries will not appear in the General Ledger (`account_financial_report`) because users can set  date filters there (for example from 01/01/2023 to 31/12/2023). Even the Trial Balance is correct. As I said, it's not about "balancing the accounting" but hiding lines in the view.

How would you reconcile more than 300.000 items from 5 years ago that can't be related to other moves? Payments have been entered manually with the correct amounts.

As I said, the account's credit/debit are balanced and they're correct. We are not creating a new feature but restoring an old one.

For me, saying that "Odoo itself is not allowing this" is not a real strong reason. Everyday we're adding new features in OCA that are not allowed in Odoo. Even we restore old features that work better than new ones (done by Odoo). And we (OCA) choose our own way of working that is different from Odoo point of view.

I really don't understand why shouldn't this feature be accepted. Users can choose the way they preffer to work. Hiding all lines (even if they're not blue) is not a must, but an option.

If you don't agree these changes, could you suggest any option in order to satisfy both parts?

THX

pedrobaeza:
You will find the same problem on v16 and the rest of the upcoming versions.

The way to deal with this is to create an extra account (in the Spanish CoA, 430099), that is not reconcilable, and load the balances there. They go to the same lines in MIS reports. You move those balances from 430999 to 430000 for those that need to still be reconciled.

HaraldPanten:
> You will find the same problem on v16 and the rest of the upcoming versions.
> 
> The way to deal with this is to create an extra account (in the Spanish CoA, 430099), that is not reconcilable, and load the balances there. They go to the same lines in MIS reports. You move those balances from 430999 to 430000 for those that need to still be reconciled.

Hi Pedro,

We'll apply this improvement in newer versions as well as we've been doing in all the modules migrations we did. By the way, It'll be only needed in V15. In newer versions (16, 17...) I think it'll not be needed, because if we get the idea of the new reconciliation methodology, you know that we've filters in the search bar.

Moving the balance to another account is not a solution. As I told you we can not make this account movements as we mustn't change the debit/credit of the accounts. They're correctly balanced and we must keep them as they are. It's a huge amount of data that can't be managed, even technically.

On the other hand, Enric found some changes in a previous PR. The condition was already there but it was removed when making visible payment/debit orders in reconciliation.

We are restoring the original functionality in a better way than it was before, so I can't understand your point of view.

> @pedrobaeza that's true, but that domain contains an OR https://github.com/OCA/account-reconcile/blob/14.0/account_reconciliation_widget/models/reconciliation_widget.py#L809
>
> On https://github.com/OCA/account-reconcile/pull/443, the condition was moved from all the domain to just the first part of that domain (the first condition of the OR). It should affect all of it.

Regards.

pedrobaeza:
On OCB, this has been forbidden since v11, so it's not a restoration at all. I still don't want to introduce this option in the same module, as existing users (our customers) will use (and abuse) this new option. Please add a new module with this option, and add hooks if you need in this module.

etobella:
IMO @pedrobaeza you removed an existent functionality on the PR I marked by moving the code (probably as a side effect)... If we put back the code (not the filter), it should work as usual and keep the functionality implemented on that PR. I don't get the problem in restoring a functionality that existed previously and was removed on a PR.

HaraldPanten:
> On OCB, this has been forbidden since v11, so it's not a restoration at all. I still don't want to introduce this option in the same module, as existing users (our customers) will use (and abuse) this new option. Please add a new module with this option, and add hooks if you need in this module.

Pedro, I still insist that it'd be an **optional** feature. It wouldn't be a mandatory option. We put a checkbox in order to filter by all the items. You can decide to use it or not.

We can hide it with a special permission group if you think it would be better. On the other hand, it's "dangerous" to allow unexperienced users managing global configuration 😝.

All we know several options that being given to users could create mass disasters, but we're not removing them.

pedrobaeza:
Putting that under an option doesn't make this as something valid to be added to a stable module. I'm very reluctant because this is a total bypass of good business practices, and I have direct cases with this that use the OCA module. Is it too much problem to add a new module `account_reconciliation_widget_limit_aml` or similar name?

HaraldPanten:
> Putting that under an option doesn't make this as something valid to be added to a stable module. I'm very reluctant because this is a total bypass of good business practices, and I have direct cases with this that use the OCA module. Is it too much problem to add a new module `account_reconciliation_widget_limit_aml` or similar name?

OK, we can do that and try to make the inheritance, but we have to know that it would not be blocked for not wasting our time.

THX.

HaraldPanten:
Closing

pedrobaeza:
Sorry for not being clear enough since the beginning. Adding as an optional module is perfect and I won't block it (I won't install it neither :stuck_out_tongue:). Or maybe yes, but only in a controlled case where the scope is very clear, not on my existing customers that have access to configuration, and discover later that they used it in an unintended way, and we have to deal with the consequences.

Do you understand my concerns?

HaraldPanten:
> Sorry for not being clear enough since the beginning. Adding as an optional module is perfect and I won't block it (I won't install it neither 😛). Or maybe yes, but only in a controlled case where the scope is very clear, not on my existing customers that have access to configuration, and discover later that they used it in an unintended way, and we have to deal with the consequences.
> 
> Do you understand my concerns?

Yes, it's OK.

This time was disapointing for us, but I know that you're trying to do your best for the Community.

As we say in Spain "_it happens in the best families_"  😂

Issue #579, opened 2023-08-23
[16.0] account_reconcile_oca reset (unreconcile fails)


nagybal:
## Module
account_reconcile_oca

## Describe the bug
If I hit reset on certain matches to be uncreconciled, the below Odoo Server Error appears:

```
RPC_ERROR
Odoo Server Error

Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 1584, in _serve_db
    return service_model.retrying(self._serve_ir_http, self.env)
  File "/usr/lib/python3/dist-packages/odoo/service/model.py", line 133, in retrying
    result = func()
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 1611, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 1815, in dispatch
    result = self.request.registry['ir.http']._dispatch(endpoint)
  File "/usr/lib/python3/dist-packages/odoo/addons/base/models/ir_http.py", line 154, in _dispatch
    result = endpoint(**request.params)
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 697, in route_wrapper
    result = endpoint(self, *args, **params_ok)
  File "/usr/lib/python3/dist-packages/odoo/addons/web/controllers/dataset.py", line 46, in call_button
    action = self._call_kw(model, method, args, kwargs)
  File "/usr/lib/python3/dist-packages/odoo/addons/web/controllers/dataset.py", line 33, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/usr/lib/python3/dist-packages/odoo/api.py", line 461, in call_kw
    result = _call_kw_multi(method, model, args, kwargs)
  File "/usr/lib/python3/dist-packages/odoo/api.py", line 448, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/usr/lib/python3/dist-packages/odoo/addons/account_reconcile_oca/models/account_bank_statement_line.py", line 564, in unreconcile_bank_line
    return getattr(self, "_unreconcile_bank_line_%s" % self.reconcile_mode)(
AttributeError: 'account.bank.statement.line' object has no attribute '_unreconcile_bank_line_False'

The above server error caused the following client error:
RPC_ERROR: Odoo Server Error
[...]

PR #584, opened 2023-09-04
[16.0][IMP] account_reconcile_oca: verify data is defined

etobella:
Should fix https://github.com/OCA/account-reconcile/issues/579#issuecomment-1705588449

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -388,7 +388,11 @@ class AccountBankStatementLine(models.Model):
         foreign_currency = (
             self.currency_id != self.company_id.currency_id
             or self.foreign_currency_id
-            or any(line["currency_id"] != line["line_currency_id"] for line in data)
+            or any(
+                line.get("line_currency_id")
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Can't you prepare data so `line_currency_id` is always filled?

etobella:
The problem might come from old data. If you had this module installed and you reconciled a line previous to the last version, data will not be there. The only option I find is to do this :thinking: migrating the lines might be  hard and expensive for just this... WDYT?

pedrobaeza:
But in any place should be filling `line` dictionary, so I'm talking to fill that value always in that place instead of having to be defensive each time you access data.

etobella:
I understand it, and we are doing that. The problem comes from old data that might have been generated before the inclussion of such data. If you reset and old line, the json might not contain the data :sob:

pedrobaeza (approved):
OK if there's no other solution. In general, storing intermediate data instead of preparing them on the fly is not a good design choice, but not knowing the details, I can't totally judge.

etobella:
Mmmh, I get your point @pedrobaeza I will create a second PR fixing it in another way, let's see what's better :smile:

PR #585, opened 2023-09-05
[16.0][IMP] account_reconcile_oca: only store info on unreconciled items

etobella:
#584

The idea is to clear the data after reconciliation (we need to keep until it is reconciled in order to ensure that we can navigate without loosing everything).
Still missing a migration script in order to get out the old data.

WDYT @pedrobaeza ?


pedrobaeza:
Now I understand more the reason for storing some "transient" information: you want to resume a half-baked reconciliation. That's not the usual case, but it can be useful. But ignoring that information if not needed is something reasonable.

pedrobaeza (approved):

PR #594, opened 2023-09-25
[16.0][IMP] account_reconcile_oca: display partner_name

astirpe:
On reconciliation widget (ReconcileDataWidget), the `t-esc="reconcile_line.partner_id[1]"` does not always show the partner name. That happens when the partner is not yet in the database (contacts). If the partner name is not displayed, we have no clue how to reconcile the line.

With this proposed PR, the Counterpart in the Partner field is shown, without the need of having the contact stored in the database.

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -189,7 +189,7 @@ class AccountBankStatementLine(models.Model):
                     "account_id": self.journal_id.suspense_account_id.name_get()[0],
                     "partner_id": self.partner_id
                     and self.partner_id.name_get()[0]
-                    or False,
+                    or (False, self.partner_name),
                     "date": fields.Date.to_string(self.date),
                     "name": self.payment_ref or self.name,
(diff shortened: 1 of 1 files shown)


ThijsvOers (approved):
Tested this locally since I got an error due to #595. The fix works and I approve it

etobella (approved):

account_reconcile_oca/models/account_bank_statement_line.py
```diff
@@ -717,3 +717,17 @@ def action_to_check(self):
...
+    ):
+        vals = super()._get_reconcile_line(
+            line,
+            kind,
+            is_counterpart=is_counterpart,
+            max_amount=max_amount,
+            from_unreconcile=from_unreconcile,
+        )
+        if vals["partner_id"] is False:
+            vals["partner_id"] = (False, self.partner_name)
```
pedrobaeza, line 732: Is this reconcile line a dictionary for feeding the widget and not one for creating a record? If so, it's better to add `vals["partner_name"]` and show it in the UI if the other value doesn't exist.

Issue #592, opened 2023-09-20
[16.0] account_reconcile_oca: error when open transactions menu in runboat DB


rinaldifirdaus:
## Module
account_reconcile_oca

## Describe the bug
In the following runboat database, error occurs when opening transactions menu from invoicing dashboard.
http://oca-account-reconcile-16-0-28d467153ecb.runboat.odoo-community.org/web#action=274&model=account.journal&view_type=kanban&cids=1&menu_id=121

error message:
```
Traceback (most recent call last):
  File "/opt/odoo/odoo/api.py", line 984, in get
    cache_value = field_cache[record._ids[0]]
KeyError: 8

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/fields.py", line 1160, in __get__
    value = env.cache.get(record, self)
  File "/opt/odoo/odoo/api.py", line 991, in get
    raise CacheMiss(record, field)
odoo.exceptions.CacheMiss: 'account.bank.statement.line(8,).can_reconcile'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/api.py", line 984, in get
    cache_value = field_cache[record._ids[0]]
KeyError: 8

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/fields.py", line 1160, in __get__
    value = env.cache.get(record, self)
  File "/opt/odoo/odoo/api.py", line 991, in get
    raise CacheMiss(record, field)
odoo.exceptions.CacheMiss: 'account.bank.statement.line(8,).reconcile_data_info'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/http.py", line 1584, in _serve_db
    return service_model.retrying(self._serve_ir_http, self.env)
  File "/opt/odoo/odoo/service/model.py", line 133, in retrying
    result = func()
  File "/opt/odoo/odoo/http.py", line 1611, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "/opt/odoo/odoo/http.py", line 1815, in dispatch
    result = self.request.registry['ir.http']._dispatch(endpoint)
  File "/opt/odoo/odoo/addons/base/models/ir_http.py", line 154, in _dispatch
    result = endpoint(**request.params)
  File "/opt/odoo/odoo/http.py", line 697, in route_wrapper
[...]

PR #580, opened 2023-08-26
[IMP] account_reconcile_oca: Include taxes from models

etobella:
Fixes #576

PR #595, opened 2023-09-26
[16.0][FIX] account_reconcile_oca: UnboundLocalError: local variable …

astirpe:
…'balance' referenced before assignment

Fixes https://github.com/OCA/account-reconcile/issues/592, by applying the patch proposed here: https://github.com/odoo/odoo/pull/123626

This is a quick&dirty fix, waiting for better ideas on how to fix it.

### DIFF
--- a/account_reconcile_oca/models/__init__.py
+++ b/account_reconcile_oca/models/__init__.py
@@ -2,3 +2,4 @@ from . import account_reconcile_abstract
 from . import account_journal
 from . import account_bank_statement_line
 from . import account_account_reconcile
+from . import account_reconcile_model
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -359,7 +359,7 @@ class AccountBankStatementLine(models.Model):
                 continue
             new_data.append(line_data)
             liquidity_amount += line_data["amount"]
-        for line in reconcile_model._get_write_off_move_lines_dict(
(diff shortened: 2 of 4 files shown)


etobella:
Can you squash your commits?

ThijsvOers (approved):
Tested this and it works. Please merge this fix as soon as possible since other modules are depending on this fix (and they are failing now). TIA

etobella:
I don't feel comfortable with this change. Which is the definition of the model that is giving you this error?

astirpe:
@etobella this is something you fixed when porting the module to V16: https://github.com/OCA/account-reconcile/pull/500#issuecomment-1484668648

But then the bug re-appeard after your PR: https://github.com/OCA/account-reconcile/pull/580

So the problem is standard odoo method `_get_write_off_move_lines_dict()` that is bugged. Model is "account.reconcile.model". The fix I applied is this one: https://github.com/odoo/odoo/pull/123626/files . Not sure I answered your question?

pedrobaeza:
Don't make a verbatim copy of the method, but overwrite `_get_write_off_move_lines_dict` and add/patch what you need.

astirpe:
This is a quick&dirty fix and i don't think I will have time to change it. If anyone will take over and propose a better solution I will be happy. Shall I close this one?

ghost:
Hi,
If Odoo merges this FIX: https://github.com/odoo/odoo/pull/154551, the error will be fixed and it won't be necessary to ovewrite any funcion.

astirpe:
Thank you @ACheung-FactorLibre

PR #598, opened 2023-10-04
[15.0][FIX] account_reconciliation_widget: m2x record keyboard  navigation

chienandalu:
We couldn't  use our direction keys to go down the record list on a many2one field as the navigation events are propagated from the field it self and go up to the widget where a navigation through the reconcile keys is expected.

Both behaviors are preserved.

Peek 04-10-2023 15-45

cc @Tecnativa TT45219


pedrobaeza (approved):
Thanks! Tested on runboat

victoralmau (approved):

sergio-teruel:
Great!!!

PR #603, opened 2023-10-20
[14.0][FIX] account_reconciliation_widget: Avoid unbalanced foreign currency reconciliations + exclude own amls in reconcile models propositions

pedrobaeza:
When having a statement line in a foreign currency, each resulting journal item computes the debit/credit amount from the foreign currency rate, and then rounding the result to company currency digits.

There's a chance in this process that the journal entry final balance is not 0 summing the rounded balances in company currency.

For fixing this, there can be several strategies, like creating an extra journal item for the difference, but I have opted for removing the difference in the latest counterpart aml, so no extra noise and no need of account decision algorithm for this extra move.

As currency amounts are correct and are the ones used in reconciliation, there won't be any problem adjusting this amount.

@Tecnativa TT45568

### DIFF
--- a/account_reconciliation_widget/models/account_bank_statement.py
+++ b/account_reconciliation_widget/models/account_bank_statement.py
@@ -1,5 +1,6 @@
 from odoo import _, fields, models
 from odoo.exceptions import UserError
+from odoo.tools import float_is_zero
 
 
 class AccountBankStatement(models.Model):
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Added an extra commit for a discovered situation testing the other patch: the journal item of a bank statement line can be proposed to itself in the reconciliation widget. This patch also requires a patch in Odoo: https://github.com/odoo/odoo/pull/139419

chienandalu (approved):

victoralmau (approved):

PR #607, opened 2023-11-02
[15.0][FIX] account_reconciliation_widget: only stop up down key events

chienandalu:
After the previous commit (311a0fa) users couldn't use the tab key to navigate over manual item fields.

cc @Tecnativa TT45219


pedrobaeza:
Shouldn't this be an allowed list instead. I think I can press other keys like left and right cursors, or characters, and they should be passed upstream?

chienandalu:
Well, this is only for these two navigation cases which are the ones that overlap with the general navigation events of the widget. So it should be read like: "Inside the widget when you're in a relational field and press up or down, do the usual stuff with relational fields, otherwise let's navigate the reconciliation items".

pedrobaeza (approved):
OK, let's go with this one:

/ocabot merge patch

Issue #605, opened 2023-10-27
[16] An error pops up when i select a transaction to reconcile


malekben:
## Module
Account reconcile oca
## Describe the bug
An error pops up when i select an item to reconcile
## To Reproduce
**Affected versions**:
v16
1. Go to accounting dashboard and click on reconcile items in a given journal
2. Try to select an item other then selected by default
3. An error pops up
UncaughtPromiseError > TypeError
Promesse non interceptée > Cannot read properties of undefined (reading 'model')
TypeError: Cannot read properties of undefined (reading 'model')
    at ReconcileController.selectRecord (http://localhost:8070/web/assets/804-53d7241/web.assets_backend.min.js:14261:94)
    at ReconcileController.openRecord (http://localhost:8070/web/assets/804-53d7241/web.assets_backend.min.js:14264:31)
    at ReconcileKanbanRecord.onGlobalClick (http://localhost:8070/web/assets/804-53d7241/web.assets_backend.min.js:4244:410)
    at Object.mainEventHandler (http://localhost:8070/web/assets/190-ab7d9b4/web.assets_common.min.js:1997:77)
    at HTMLDivElement.listener (http://localhost:8070/web/assets/190-ab7d9b4/web.assets_common.min.js:1240:15)

**Expected behavior**
It should change the interface to the selected item.

norlinhenrik:
@malekben @etobella 
Are you able to reproduce the error? I am not.
I made a new bank journal.
I made 2 invoices of the same product, to 2 different partners. 
I made a manual bank transaction of the same amount.
Reconcile suggests one of the invoices. I can delete the suggestion and select the other invoice.

PR #610, opened 2023-11-09
[16.0][IMP] account_reconcile_oca: Some fixes and improvements

etobella:
This PR fixes the bug #605  and allows to access the reconciliation widget easily when no lines are defined (but we want to create them manually)

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -351,6 +351,53 @@ class AccountBankStatementLine(models.Model):
         for record in self:
             record.reconcile_data = record.reconcile_data_info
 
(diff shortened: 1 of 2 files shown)


account_reconcile_oca/models/account_bank_statement_line.py
```diff
@@ -351,6 +351,53 @@ def _inverse_reconcile_data_info(self):
         for record in self:
             record.reconcile_data = record.reconcile_data_info
 
+    def _apply_lines_for_bank_widget(self, residual_amount_currency, partner, model):
+        """
+        This function is quite similar to the one from account.reconcile.model but adding
+        some functionalities for taxes
```
pedrobaeza, line 357: Why do you need this?

PR #619, opened 2024-01-02
[14.0][FIX] account_reconciliation_widget: Don't link counterpart to the statement line

pedrobaeza:
Backport of #617 

Steps to reproduce the problem:

- Issue an invoice.
- Add a bank statement for paying such invoice.
- Reconcile the statement with the invoice. It gets paid.
- Revert reconciliation from the statement.
- Go again to reconcile the statement line.

Expected behavior: The invoice is available to be reconciled again.

Current behavior: The invoice is not available.

That's because the move line of the AR/AP account was linked to the statement line in the reconciliation process through the field `statement_line_id`. That field is used for linking the generated move lines to the generating statement line, not for other things.

This is probably a bad migration from v13 original code to v14, where the reconciliation and the datamodel changed.

It includes a migration script that mitigates a bit the problem on past reconciled invoices, removing such link. It doesn't cover other AR/AP lines, but at least we keep data consistent for these ones.

@Tecnativa TT46644


chienandalu (approved):

victoralmau (approved):

pedrobaeza:
Unrelated CI error with `account_mass_reconcile_as_job`, so let's merge this manually (as we have version bump) to fix the conflicting behavior.

PR #629, opened 2024-02-03
[17.0][MIG] account_reconcile_oca

norlinhenrik:
I have done the migration tasks. I am very bad in javascript. When installing I got this error. Is there anyone else who can fix it?
account_reconcile_oca


norlinhenrik:
I have disabled the chatter and tried to fix the other js. But when I open bank transactions, I get 
**Cannot read properties of undefined (reading 'root')**

norlinhenrik:
@etobella Would you like to help with this migration? I can pay for a couple of hours.

etobella:
Not an easy task for several reasons:

- Odoo removed some logic for reconcile models and needs to be restored... I would recommend a different module in order to avoid license problems
- There was an error of definiion on the views using list when should be kanban....
- I have issues showing the new widgets. Almost ready

norlinhenrik:
Thank you for these important insights.

> Odoo removed some logic for reconcile models and needs to be restored... I would recommend a different module in order to avoid license problems

Is this python or javascript or both?

> There was an error of definition on the views using list when should be kanban....

Where? I cannot find it.

> I have issues showing the new widgets. Almost ready

Nice!

etobella:
image


Now I will prepare the missing functions moved from CE to EE

etobella:
Superseed by #630

PR #631, opened 2024-02-09
[WIP][16.0][IMP] account_reconcile_oca: Create exchange diff move when concile c…

aritzolea:
…urrencies don't match

### DIFF
--- a/account_reconcile_oca/models/account_partial_reconcile.py
+++ b/account_reconcile_oca/models/account_partial_reconcile.py
@@ -1,0 +1,15 @@
+# Copyright 2024 FactorLibre - Aritz Olea
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import api, models
+
+
+class AccountPartialReconcile(models.Model):
+
+    _inherit = "account.partial.reconcile"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        if self.env.context.get("no_foreign_currency", False):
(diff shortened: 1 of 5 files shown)


account_reconcile_oca/models/account_partial_reconcile.py
```diff
@@ -0,0 +1,15 @@
...
+
+class AccountPartialReconcile(models.Model):
+
+    _inherit = "account.partial.reconcile"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        if self.env.context.get("no_foreign_currency", False):
+            for vals in vals_list:
+                vals["debit_amount_currency"] = 0
```
etobella, line 17: Why not credit? It seems something weird to do it with this context. I would recommend to give some more context to your issue, how it was raised and so on...

etobella:
After reviewing Odoo Code for reconcile I do not agree in this approach. this fields are computed automatically on reconcile function....

Can you explain the reason of this PR?

SoniaViciana:
I haven't tried the PR yet, but want to solve this:

image

image

etobella (changes requested):
But removing the logic of foreign currency should not be an option. 

There is probably a bug in the code, but applying this patch is not the solution

aritzolea:
Closing PR after checking the bug is coming indeed from Core Odoo 16. We are trying to get a solution but applied to core directly.

etobella:
Thanks @aritzolea for checking

PR #637, opened 2024-03-11
[16.0][IMP] [account_statement_base account_reconcile_oca] make statements listed and editable

lk-eska:
Currently there is no way to list or edit bank statement lines. this PR brings this feature to account_statement_base account_reconcile_oca modules.

### DIFF
--- a/account_reconcile_oca/views/account_bank_statement_line.xml
+++ b/account_reconcile_oca/views/account_bank_statement_line.xml
@@ -306,11 +306,12 @@
         <field
             name="context"
         >{'default_journal_id': active_id, 'search_default_not_reconciled': True, 'view_ref': 'account_reconcile_oca.bank_statement_line_form_reconcile_view'}</field>
-        <field name="view_mode">tree</field>
+        <field name="view_mode">kanban,tree</field>
         <field
             name="view_ids"
             eval="[(5, 0, 0),
-            (0, 0, {'view_mode': 'tree', 'view_id': ref('bank_statement_line_reconcile_view')})]"
+            (0, 0, {'view_mode': 'kanban', 'view_id': ref('bank_statement_line_reconcile_view')}),
+            (0, 0, {'view_mode': 'tree', 'view_id': ref('account_statement_base.account_bank_statement_line_tree')})]"
         />
         <field name="help" type="html">
             <p class="o_view_nocontent_smiling_face">
@@ -325,11 +326,12 @@
         <field
             name="context"
         >{'default_journal_id': active_id, 'view_ref': 'account_reconcile_oca.bank_statement_line_form_reconcile_view'}</field>
-        <field name="view_mode">tree</field>
+        <field name="view_mode">tree,kanban</field>
         <field
             name="view_ids"
             eval="[(5, 0, 0),
-            (0, 0, {'view_mode': 'tree', 'view_id': ref('bank_statement_line_reconcile_view')})]"
+            (0, 0, {'view_mode': 'tree', 'view_id': ref('account_statement_base.account_bank_statement_line_tree')}),
+            (0, 0, {'view_mode': 'kanban', 'view_id': ref('bank_statement_line_reconcile_view')})]"
         />
         <field name="help" type="html">
             <p class="o_view_nocontent_smiling_face">
--- a/account_statement_base/views/account_bank_statement_line.xml
+++ b/account_statement_base/views/account_bank_statement_line.xml
@@ -60,21 +60,51 @@
 <record id="account_bank_statement_line_tree" model="ir.ui.view">
     <field name="model">account.bank.statement.line</field>
     <field name="arch" type="xml">
-        <tree decoration-muted="is_reconciled">
+        <tree editable="top" multi_edit="1" decoration-muted="is_reconciled">
             <field name="sequence" widget="handle" />
-            <field name="date" />
-            <field name="payment_ref" />
-            <field name="ref" optional="hide" />
-            <field name="partner_id" />
-            <field name="amount" />
+            <field name="date" attrs="{'readonly': [('is_reconciled', '=', True)]}" />
+            <field
+                    name="payment_ref"
+                    attrs="{'readonly': [('is_reconciled', '=', True)]}"
+                />
+            <field
+                    name="ref"
+                    optional="hide"
+                    attrs="{'readonly': [('is_reconciled', '=', True)]}"
+                />
+            <field
+                    name="partner_id"
+                    attrs="{'readonly': [('is_reconciled', '=', True)]}"
+                />
+            <field name="amount" attrs="{'readonly': [('is_reconciled', '=', True)]}" />
+            <field
+                    name="foreign_currency_id"
+                    optional="hidden"
+                    groups="base.group_multi_currency"
+                    attrs="{'readonly': [('is_reconciled', '=', True)]}"
+                />
             <field
                     name="amount_currency"
                     optional="hide"
                     groups="base.group_multi_currency"
+                    attrs="{'readonly': [('is_reconciled', '=', True)]}"
                 />
-            <field name="narration" optional="hide" string="Notes" />
-            <field name="journal_id" optional="hide" />
-            <field name="partner_bank_id" optional="hide" />
+            <field
+                    name="narration"
+                    optional="hide"
+                    string="Notes"
+                    attrs="{'readonly': [('is_reconciled', '=', True)]}"
+                />
+            <field
+                    name="journal_id"
+                    optional="hide"
+                    attrs="{'readonly': [('is_reconciled', '=', True)]}"
+                />
+            <field
+                    name="partner_bank_id"
+                    optional="hide"
+                    attrs="{'readonly': [('is_reconciled', '=', True)]}"
+                />
             <button
                     name="action_undo_reconciliation"
                     type="object"
@@ -85,9 +115,9 @@
             <field name="company_id" invisible="1" />
             <field name="is_reconciled" invisible="1" />
             <field name="currency_id" invisible="1" />
-            <field name="foreign_currency_id" invisible="1" />
             <field name="country_code" invisible="1" />
             <field name="state" invisible="1" />
+            <field name="suitable_journal_ids" invisible="1" />
         </tree>
     </field>
 </record>


account_reconcile_oca/views/account_bank_statement_line.xml
```diff
@@ -325,11 +326,12 @@
         <field
             name="context"
         >{'default_journal_id': active_id, 'view_ref': 'account_reconcile_oca.bank_statement_line_form_reconcile_view'}</field>
-        <field name="view_mode">tree</field>
+        <field name="view_mode">tree,kanban</field>
```
etobella, line 329:
```suggestion
        <field name="view_mode">kanban,tree</field>
```

etobella (commented):
Just a small comment and we can proceed

account_reconcile_oca/views/account_bank_statement_line.xml
```diff
@@ -306,11 +306,12 @@
         <field
             name="context"
         >{'default_journal_id': active_id, 'search_default_not_reconciled': True, 'view_ref': 'account_reconcile_oca.bank_statement_line_form_reconcile_view'}</field>
-        <field name="view_mode">tree</field>
+        <field name="view_mode">kanban,tree</field>
         <field
             name="view_ids"
             eval="[(5, 0, 0),
-            (0, 0, {'view_mode': 'tree', 'view_id': ref('bank_statement_line_reconcile_view')})]"
+            (0, 0, {'view_mode': 'tree', 'view_id': ref('account_statement_base.account_bank_statement_line_tree')}),
```
etobella: Change the order, first the kanban please :wink:

account_reconcile_oca/views/account_bank_statement_line.xml
```diff
@@ -325,11 +326,12 @@
         <field
             name="context"
         >{'default_journal_id': active_id, 'view_ref': 'account_reconcile_oca.bank_statement_line_form_reconcile_view'}</field>
-        <field name="view_mode">tree</field>
+        <field name="view_mode">tree,kanban</field>
         <field
             name="view_ids"
             eval="[(5, 0, 0),
-            (0, 0, {'view_mode': 'tree', 'view_id': ref('bank_statement_line_reconcile_view')})]"
+            (0, 0, {'view_mode': 'tree', 'view_id': ref('account_statement_base.account_bank_statement_line_tree')}),
```
etobella, line 333: Change the order, first the kanban please :wink:

etobella (commented):
The order is important, as the default loaded view is tree now

account_reconcile_oca/views/account_bank_statement_line.xml
lk-eska: I just sent a fix.

pedrobaeza:
I haven't checked yet, but statement lines shouldn't be editable once the statement is published (just in case this is favoring other behavior).

etobella:
@pedrobaeza I had the same concern, but I tested it on runboat and an error was raised when editing reconciled moves :smiley:

pedrobaeza:
Then isn't better to not allow to edit such reconciled entries?

lk-eska:
Odoo doesn't allow you to change reconciled entries so there is no such concern.

The real problem is not having edit/delete capability on the statements. How to change or delete them if you imported wrong? How to set foreign currency and foreign currency amount? Is it possible to do it somewhere else?

pedrobaeza:
> Odoo doesn't allow you to change reconciled entries so there is no such concern.

What I mean is UX. According to what I'm understanding, the interface will let you change data in reconciled lines, but at the saving time, it will rise an error. What I'm talking about is to directly avoid to change such data for the reconciled entries.

lk-eska:
How do we allow editing some of the lines conditionally? (on tree editable mode)  We can improve UX if this is possible. It works on enterprise similarly by the way.

etobella:
In 16 you should do: attrs="{'readonly': [('is_reconciled', '=', True)]}" on the view for each field

lk-eska:
> In 16 you should do: attrs="{'readonly': [('is_reconciled', '=', True)]}" on the view for each field

I added this to this PR.

etobella:
Can you squash your commits and a use a proper commit message?

Thanks :wink:

lk-eska:
There should be "Squash & Merge" option for accepting a PR?

pedrobaeza:
No, the bot can't squash, so it should be done manually.

etobella:
I checked and with this new tree view, we can unlink the items. It is right that we can unlink lines that are not reconciled, but with this we were able to unlink reconciled lines. Can you add a validation on the unlink in order avoid this?

HaraldPanten:
> I checked and with this new tree view, we can unlink the items. It is right that we can unlink lines that are not reconciled, but with this we were able to unlink reconciled lines. Can you add a validation on the unlink in order avoid this?

Hi @lk-eska @pedrobaeza I think that it should be fixed before merging this PR. Deleting reconciled items in the tree view should be forbidden. If you want, we can manage this PR keeping you as author.

Let us know, please.

lk-eska:
> No, the bot can't squash, so it should be done manually.

I squashed them into one commit.

lk-eska:
> > I checked and with this new tree view, we can unlink the items. It is right that we can unlink lines that are not reconciled, but with this we were able to unlink reconciled lines. Can you add a validation on the unlink in order avoid this?
> 
> Hi @lk-eska @pedrobaeza I think that it should be fixed before merging this PR. Deleting reconciled items in the tree view should be forbidden. If you want, we can manage this PR keeping you as author.
> 
> Let us know, please.

Deleting a statement line (unless it is reconciled with an invoice) is standard behavior in enterprise. This is not a bug but a feature and all enterprise users live with it without complaining. I think we are getting beyond the scope of this pull request. Not that I care about authorship but if you think this is required you can create your own PR fixing it once this gets merged.

etobella:
@lk-eska It is already nice to delete not reconciled lines, but right now,we are able to delete reconciled lines.

lk-eska:
I can't do it in runboat if it is reconciled with an invoice:

image

It can be deleted if it is reconciled manually and it makes sense.

etobella:
But only with reconciled items with invoices

You can see here what happens... Obviously, this is not right

https://github.com/OCA/account-reconcile/assets/28590170/845c836b-ee8a-4990-bbd9-255207f672d6

lk-eska:
> But only with reconciled items with invoices
> 
> You can see here what happens... Obviously, this is not right
> movie.mp4

It is deleting the generated account move so I see no problem here. Why bother unreconciling it when you can skip that step?

etobella:
I understand that a reconciled move should never be deleted, as it is posted.

So, IMO it is not a desired behavior.

HaraldPanten:
What about payments related to reconciled account moves like manual ones? For example like the ones that we create in the initial balance, or the ones coming from payrolls? Is that a desired behaviour?

lk-eska:
> What about payments related to reconciled account moves like manual ones? For example like the ones that we create in the initial balance, or the ones coming from payrolls? Is that a desired behaviour?

I haven't tested all possible use cases but I assume it allows deletion if it is not reconciled with another move. It should cover most of your questions. IMO, deleting is not a big deal if it is a standalone entry.

HaraldPanten:
> > What about payments related to reconciled account moves like manual ones? For example like the ones that we create in the initial balance, or the ones coming from payrolls? Is that a desired behaviour?
> 
> I haven't tested all possible use cases but I assume it allows deletion if it is not reconciled with another move. It should cover most of your questions. IMO, deleting is not a big deal if it is a standalone entry.

Hi, I'm starting to think that you're right. I've tested V15 and the functionality is the same as here (I think). In V15 you can delete reconciled bank statement lines as well.

pedrobaeza:
> In V15 you can delete reconciled bank statement lines as well.

AFAIK, that's not possible. What are the steps you do for that?

lk-eska:
I think there is a confusion when we say reconciled. Let me clarify:
1) reconciled with another entry such as invoice (deletion is not possible)
2) directly creating a new entry such as manual expense entry (deletion is possible)

HaraldPanten:
These are my steps in V15:

1. Create a bank statement with a bank statement line of -15$.
2. Validate the statement.
3. Go to the reconciliation view.
4. Make a manual operation an send these -15$ to any expenses account.
5. Go to the operations view (see screenshot).

Captura de Pantalla 2024-03-15 a las 11 53 48

6. Select the reconciled line in the tree view and try to delete it from the tree view.

Captura de Pantalla 2024-03-15 a las 11 54 00

8. Users can delete it without any restriction.

@pedrobaeza @etobella I think it's not correct, but it seems it has been working like that for several years. WDYT?

HaraldPanten:
> I think there is a confusion when we say reconciled. Let me clarify:
> 
> 1. reconciled with another entry such as invoice (deletion is not possible)
> 2. directly creating a new entry such as manual expense entry (deletion is possible)

Yep. That's it.

HaraldPanten (approved):
Manual operations can be deleted, but reconciled statement lines (with invoices or other payable/receivable accounts like customers/vendors in the starting balance) can't be deleted once they're reconciled.

IMHO, this pull request is functionally correct. LGTM.

pedrobaeza:
OK, analyzing it, it seems congruent with the new v16 approach:

- Statement lines can be created without statement.
- You can add a new statement line in any moment, so deleting the line it's not limiting you forcing to create a new statement.

Anyway, I think we should add a protection over lines belonging to an statement, as if you remove a line, your balances won't fit and you won't be able to add the line in the same statement. What do you think? I'm thinking in statements imported from banks, where the transaction list should be unalterable at that level.

HaraldPanten:
> OK, analyzing it, it seems congruent with the new v16 approach:
> 
> * Statement lines can be created without statement.
> * You can add a new statement line in any moment, so deleting the line it's not limiting you forcing to create a new statement.
> 
> Anyway, I think we should add a protection over lines belonging to an statement, as if you remove a line, your balances won't fit and you won't be able to add the line in the same statement. What do you think? I'm thinking in statements imported from banks, where the transaction list should be unalterable at that level.

I agree, but shouldn't we work in this new IMP/FIX in a new PR?

lk-eska:
Can we merge this now?

etobella:
Yes, I added a proper commit message and splitted them between both modules (Sorry for making you merge them, because they were two modules and I didn't notice that)

For the next time, remember to use proper commits message https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#L1114-L1167

As you can see, I kept your authorship and no modification on the code

If it is ok for you the applied changes, I can proceed with merge

HaraldPanten:
But I still see 2 different commits. Is that correct, @etobella ?

etobella:
Yes, two modules are modified.

lk-eska:
No problem @etobella ! You can merge it.

etobella (approved):
/ocabot merge patch

etobella:
Trying again

/ocabot merge patch

PR #643, opened 2024-04-10
[12.0][FIX] reconciles not being sorted

iTecan:
Original Code:

New code:

This also happens in 13.0, 14.0, 15.0


florian-dacosta (approved):

AaronHForgeFlow (approved):

pedrobaeza:
Uf, still v12...

/ocabot merge patch

PR #646, opened 2024-04-16
[14.0][IMP] account_reconciliation_widget: Performance opening reconciliation widget

pedrobaeza:
There was an unneeded mapped of the initial statements that fetches a lot of data from statement lines that are not going to be used later, so let's remove it and optimize a bit the initial opening time.

In a customer database, we have improved the opening time from 120 seconds to 15.

@Tecnativa TT48753

### DIFF
--- a/account_reconciliation_widget/models/reconciliation_widget.py
+++ b/account_reconciliation_widget/models/reconciliation_widget.py
@@ -308,12 +308,6 @@ class AccountReconciliation(models.AbstractModel):
         """
         if not bank_statement_line_ids:
             return {}
-        bank_statements = (
-            self.env["account.bank.statement.line"]
-            .browse(bank_statement_line_ids)
-            .mapped("statement_id")
-        )
-
         query = """
              SELECT line.id
              FROM account_bank_statement_line line
@@ -340,18 +334,17 @@ class AccountReconciliation(models.AbstractModel):
             [line["st_line"]["id"] for line in results["lines"]]
         )
         bank_statements_left = bank_statement_lines_left.mapped("statement_id")
-
+        data = bank_statements_left.read(["name", "journal_id"])
+        data = data and data[0] or {}
         results.update(
             {
                 "statement_id": len(bank_statements_left) == 1
-                and bank_statements_left.id
+                and bank_statements_left[0].id
                 or False,
                 "statement_name": len(bank_statements_left) == 1
-                and bank_statements_left.name
+                and data.get("name")
                 or False,
-                "journal_id": bank_statements
-                and bank_statements[0].journal_id.id
-                or False,
+                "journal_id": data and data.get("journal_id", [False])[0] or False,
                 "notifications": [],
             }
         )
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Unrelated error...

account_reconciliation_widget/models/reconciliation_widget.py
```diff
@@ -340,18 +334,17 @@ def get_bank_statement_data(self, bank_statement_line_ids, srch_domain=None):
...
         )
         bank_statements_left = bank_statement_lines_left.mapped("statement_id")
-
+        data = bank_statements_left.read(["name", "journal_id"])
+        data = data and data[0]
         results.update(
             {
                 "statement_id": len(bank_statements_left) == 1
-                and bank_statements_left.id
+                and bank_statements_left[0].id
```
chienandalu, line 342: Maybe better to use `fields.first`. If the recordset is empty this would fail.

account_reconciliation_widget/models/reconciliation_widget.py
```diff
@@ -340,18 +334,17 @@ def get_bank_statement_data(self, bank_statement_line_ids, srch_domain=None):
             [line["st_line"]["id"] for line in results["lines"]]
         )
         bank_statements_left = bank_statement_lines_left.mapped("statement_id")
-
+        data = bank_statements_left.read(["name", "journal_id"])
+        data = data and data[0]
```
chienandalu: In case of empty list it could be better to fallback to an empty dict so we can use `get` later.
```suggestion
        data = bank_statements_left.read(["name", "journal_id"]) or {}
```

account_reconciliation_widget/models/reconciliation_widget.py
```diff
@@ -340,18 +334,17 @@ def get_bank_statement_data(self, bank_statement_line_ids, srch_domain=None):
...
                 "statement_id": len(bank_statements_left) == 1
-                and bank_statements_left.id
+                and bank_statements_left[0].id
                 or False,
                 "statement_name": len(bank_statements_left) == 1
-                and bank_statements_left.name
-                or False,
-                "journal_id": bank_statements
-                and bank_statements[0].journal_id.id
+                and data["name"]
```
chienandalu: Like here
```suggestion
                and data.get("name")
```

account_reconciliation_widget/models/reconciliation_widget.py
```diff
@@ -340,18 +334,17 @@ def get_bank_statement_data(self, bank_statement_line_ids, srch_domain=None):
...
                 or False,
+                "journal_id": data and data["journal_id"][0],
```
chienandalu: Or here:
```suggestion
                "journal_id": data.get("journal_id", [False])[0],
```

account_reconciliation_widget/models/reconciliation_widget.py
pedrobaeza: It's not going to fail, as there's a check before, and `.id` doesn't need to be fetched from DB, so it's OK this way.

pedrobaeza:
@chienandalu changed except the first one, that is not needed.

chienandalu (approved):
:+1:

victoralmau (approved):

PR #649, opened 2024-04-24
[14.0][FIX] account_mass_reconcile: reconciles not being sorted

iTecan:
Original Code:

New code:

This also happens in 15.0


florian-dacosta (approved):

kaynnan (approved):
LGTM

PR #655, opened 2024-05-03
[14.0][IMP] account_reconciliation_widget: date_maturity and id sorted in DESC order, amount move to second ORDER BY

kaynnan:
cc @marcelsavegnago @douglascstd @WesleyOliveira98 @Matthwhy 

In Odoo version 12, the `account.reconciliation.widget` sorted the `date_maturity` in `DESC` order along with the `id`. However, in version 14, this behavior was changed to sort the date maturity in `ASC` order. 

This change was implemented to improve the user's visibility into how their transactions are presented. Sorting the date maturity and `id` in `DESC` order provides users with a clearer view of their transaction history, enhancing the overall user experience.

Before:

image

After:

image

### DIFF
--- a/account_reconciliation_widget/models/reconciliation_widget.py
+++ b/account_reconciliation_widget/models/reconciliation_widget.py
@@ -116,9 +116,9 @@ class AccountReconciliation(models.AbstractModel):
             SELECT "account_move_line".id, COUNT(*) OVER() FROM {from_clause}
             {where_str}
             ORDER BY ("account_move_line".debit -
-                      "account_move_line".credit) = {amount} DESC,
-                "account_move_line".date_maturity ASC,
-                "account_move_line".id ASC
+                    "account_move_line".credit) = {amount} DESC,
+                    "account_move_line".date_maturity DESC,
+                    "account_move_line".id DESC
             {limit_str}
         """.format(
                 from_clause=from_clause,


WesleyOliveira98 (approved):

pedrobaeza:
Can you please point to the v12 code where this happens? Check also the CI.

kaynnan:
@pedrobaeza It happens in this section, we can see that at the time there was no other group by for {amount}, but the others were DESC [get_move_lines_for_bank_statement_line V12](https://github.com/odoo/odoo/blob/12.0/addons/account/models/reconciliation_widget.py#L78)

account_reconciliation_widget/models/reconciliation_widget.py
```diff
@@ -115,10 +115,10 @@ def get_move_lines_for_bank_statement_line(
             """
             SELECT "account_move_line".id, COUNT(*) OVER() FROM {from_clause}
             {where_str}
-            ORDER BY ("account_move_line".debit -
-                      "account_move_line".credit) = {amount} DESC,
-                "account_move_line".date_maturity ASC,
-                "account_move_line".id ASC
+            ORDER BY "account_move_line".date_maturity DESC,
+                    ("account_move_line".debit -
+                    "account_move_line".credit) = {amount} DESC,
```
pedrobaeza, line 119: Don't put down this. It's very useful to have first the exact amount matches

kaynnan: I moved {amount} to the second line to have the correct behavior, if it remains in the first GROUP BY, the newest due dates are brought but even so there are still old dates present, so moving it to second position and changing the others to DESC, we have behavior similar to v12

kaynnan: How could we deal with this conflict without moving it to second position?

pedrobaeza: As said, I prefer to have first the exact matches, no matter the maturity date.

pedrobaeza: I don't know how to resolve it, but as is right now it's very useful and I don't want to lose this feature (in fact, I miss it a lot in v16).

kaynnan: I understand, at this point I will perform the tests again and commit again, perhaps it could have been some cache in my environment that caused this impression of conflict and made me move to second position

kaynnan: After making the adjustments, it had the same behavior, possibly some momentary cache made me have this impression and the need to change it to second position. Thx for review

pedrobaeza:
@kaynnan sorry, but I don't see this as an improvement, as you are putting in the first lines the newest lines, and the idea is to put in the first places the pending things that are most possible to be the matchings.

If you receive a payment, what is the most probable case? To pay the newest invoice? Or to pay the oldest one? I say the second, and that's why it's put as DESC.

kaynnan:
@pedrobaeza I understand, in that case I will close the PR, thanks for the reviews

PR #660, opened 2024-05-28
[16][FIX] account_move_line_reconcile_manual fix foreign currency reconciliation

florian-dacosta:
This PR fix the full reconciliation case for the cases when all entries have the same foreign currency. We compare the currency amount to detect if the reconciliation is full instead of the company currency.

After this PR there is one more problematic case : when we try to reconcile a line with foreign currency with a ligne with company currency.
During reconciliation, Odoo convert the line with company currency to foreign currency... and in the wizard, the company currency amounts are compared.
So in case of full reconciliation (with writeoff) it does not work. I have added the issue in the README.

@alexis-via


mourad-ehm (approved):

florian-dacosta:
@alexis-via

PR #677, opened 2024-07-31
[17.0][IMP] account_reconcile_oca: Use receivable/payable account on statements with partner as suspense account

etobella:
With this change, it sets the proper account automatically

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -239,13 +239,17 @@ class AccountBankStatementLine(models.Model):
                     }
                 )
             else:
+                account = self.journal_id.suspense_account_id
+                if self.partner_id and total_amount > 0:
+                    can_reconcile = True
+                    account = self.partner_id.property_account_receivable_id
+                elif self.partner_id and total_amount < 0:
+                    can_reconcile = True
+                    account = self.partner_id.property_account_payable_id
                 suspense_line = {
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Steps to reproduce the problem?

etobella:
It was not actually a problem. With the original solution, if I add a new statement.line related to a partner, the suspense account will be the suspense account of the Journal. With the new change, in this case will get the right receivable or payable account of the partner.

It is the behavior we had on 15 and Odoo is promoting on EE

pedrobaeza (approved):
OK, not sure the gain with this move, but it's not the usual case anyway.

etobella:
Well, it is how it worked and we lost that functionality from reconcile_widget to account_reconcile_oca. Someone noticed and I made the change because it wasn't too much :smile:

pedrobaeza:
Yeah, now I see:

Peek 2024-08-01 09-08

but it was so insignificant, as that old line doesn't serve for anything if you don't click or manually create a line. What I miss is to have the suspense line as grey to know that "you have to do an action".

JordiBForgeFlow:
I agree with @pedrobaeza. In enterprise the account stays as suspense account, but from a UI perspective it shows the receivable/payable. Weird.

But it's not a problem to show the suspense account because when reconciling with a line, it will be changed right? Perhaps the issue is when you don't want to reconcile with a line but keep the amount in the default partner receivable/payable account. Then you must remember this account.

pedrobaeza:
@etobella can you rebase?

PR #678, opened 2024-08-05
[16.0][IMP] account_statement_base: add possibility to navigate from statement lines to the associated journal entry

JordiBForgeFlow:
Through a button. Adds also the journal entry number as an optional field.

image

cc @etobella @pedrobaeza @alexis-via

### DIFF
--- a/account_statement_base/models/account_bank_statement_line.py
+++ b/account_statement_base/models/account_bank_statement_line.py
@@ -1,0 +1,21 @@
+# Copyright 2024 ForgeFlow
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class AccountBankStatementLine(models.Model):
+
+    _inherit = "account.bank.statement.line"
+
+    def action_open_journal_entry(self):
+        self.ensure_one()
+        if not self:
+            return {}
+        result = self.env["ir.actions.act_window"]._for_xml_id(
+            "account.action_move_line_form"
+        )
+        res = self.env.ref("account.view_move_form", False)
+        result["views"] = [(res and res.id or False, "form")]
+        result["res_id"] = self.move_id.id
+        return result
--- a/account_statement_base/views/account_bank_statement_line.xml
+++ b/account_statement_base/views/account_bank_statement_line.xml
@@ -63,6 +63,7 @@
         <tree editable="top" multi_edit="1" decoration-muted="is_reconciled">
             <field name="sequence" />
             <field name="date" attrs="{'readonly': [('is_reconciled', '=', True)]}" />
+            <field name="move_id" optional="hide" />
             <field
                     name="payment_ref"
                     attrs="{'readonly': [('is_reconciled', '=', True)]}"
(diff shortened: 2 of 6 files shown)


account_statement_base/models/account_bank_statement_line.py
```diff
@@ -0,0 +1,22 @@
...
+
+class AccountBankStatementLine(models.Model):
+
+    _inherit = "account.bank.statement.line"
+
+    def action_open_journal_entry(self):
+        self.ensure_one()
+        if not self:
+            return {}
+        action = self.env.ref("account.action_move_line_form").sudo().read()[0]
```
pedrobaeza: Please use the newer form to do this with `self.env["ir.actions.actions"]._for_xml_id()`

account_statement_base/models/account_bank_statement_line.py
```diff
@@ -0,0 +1,22 @@
...
+        action["view_mode"] = "form"
+        action["res_id"] = self.move_id.id
+        del action["view_id"]
+        del action["views"]
```
pedrobaeza: You are removing here the possible specific form view. This can have side effects.

account_statement_base/views/account_bank_statement_line.xml
```diff
@@ -112,12 +113,19 @@
                     icon="fa-undo"
                     attrs="{'invisible': [('is_reconciled', '=', False)]}"
                 />
+            <button
+                    name="action_open_journal_entry"
+                    type="object"
+                    string="Open Journal Entry"
```
pedrobaeza: I think there shouldn't be string. And please put title for proper screen readers (for blind people) usage.

account_statement_base/views/account_bank_statement_line.xml
```diff
@@ -112,12 +113,19 @@
...
+                    icon="fa-folder-open-o"
+                />
             <field name="company_id" invisible="1" />
             <field name="is_reconciled" invisible="1" />
             <field name="currency_id" invisible="1" />
             <field name="country_code" invisible="1" />
             <field name="state" invisible="1" />
             <field name="suitable_journal_ids" invisible="1" />
+
```
pedrobaeza: Don't add empty line here.

pedrobaeza (changes requested):

JordiBForgeFlow:
@pedrobaeza Thanks. I have attended your review comments

pedrobaeza:
You should squash by default the commits, as it doesn't add value. We have here in GitHub a differential view of the latest changes already.

JordiBForgeFlow:
Squashed. 

What I don't like about adding the button is that it shows always. We cannot make it optional, as it happens with normal fields.

pedrobaeza:
Really? You can make it invisible AFAIK.

etobella (approved):
I don't see a problem on leaving the button visible :thinking:

pedrobaeza:
Sorry, I wrote the reverse sentence: "you can make the button invisible with the `attrs` thing AFAIK". Can you try?

JordiBForgeFlow:
@pedrobaeza Can we merge this one?

pedrobaeza:
Please do it invisible when no move present.

JordiBForgeFlow:
> Please do it invisible when no move present.

I do not understand your suggestion. there's always a move.

pedrobaeza (approved):
Oh, sorry, my bad.

/ocabot merge minor

PR #679, opened 2024-08-07
[16.0][IMP] account_statement_base: Show action to create Statements

carlos-lopez-tecnativa:
Before this commit, when a user clicked the "New Transaction" button, an error was raised. After this commit, statement lines are displayed to create new records.

@Tecnativa TT50459

@pedrobaeza  @carolinafernandez-tecnativa could you please take a look

### DIFF
--- a/account_statement_base/models/account_journal_dashboard.py
+++ b/account_statement_base/models/account_journal_dashboard.py
@@ -1,0 +1,16 @@
+from odoo import models
+
+
+class AccountJournal(models.Model):
+    _inherit = "account.journal"
+
+    def create_cash_statement(self):
+        # Totally override this action for avoiding the standard
+        # message saying that you need to install the enterprise
+        # module. We do the equivalent thing instead.
+        self.ensure_one()
+        action = self.env["ir.actions.actions"]._for_xml_id(
+            "account_statement_base.account_bank_statement_line_action"
+        )
+        action["context"] = {"search_default_journal_id": self.id}
+        return action
--- a/account_statement_base/__init__.py
+++ b/account_statement_base/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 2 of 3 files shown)


account_statement_base/models/account_journal_dashboard.py
```diff
@@ -0,0 +1,15 @@
...
+    _inherit = "account.journal"
+
+    def create_cash_statement(self):
+        self.ensure_one()
+        action = self.env["ir.actions.actions"]._for_xml_id(
+            "account_statement_base.account_bank_statement_line_action"
+        )
+        action["context"] = {
+            "search_default_journal_id": self.id,
+        }
```
pedrobaeza: Without the final comma, everything goes to one line:

```suggestion
        action["context"] = {"search_default_journal_id": self.id}
```

account_statement_base/models/account_journal_dashboard.py
```diff
@@ -0,0 +1,15 @@
+from odoo import models
+
+
+class AccountJournal(models.Model):
+    _inherit = "account.journal"
+
+    def create_cash_statement(self):
+        self.ensure_one()
```
pedrobaeza, line 11: Please always document the why of the things:

```suggestion
        # Totally override this action for avoiding the standard
        # message saying that you need to install the enterprise
        # module. We do the equivalent thing instead.
        self.ensure_one()
```

pedrobaeza (commented):
There's no runboat to try though. Let's see in the next build.

account_statement_base/models/account_journal_dashboard.py
carlos-lopez-tecnativa: Thanks, I only added a similar comment in the commit message but not in the function itself. I will take that into account next time.

carolinafernandez-tecnativa (approved):
Functional review LGTM, Thanks

pedrobaeza (approved):
/ocabot merge patch

PR #680, opened 2024-08-08
[16.0] [IMP] account_mass_reconcile: add _filter in view

lfreeke:
Add the _filter in the view, so that user can use the field, The field is useful when you want to reconcile between certain dates for example.

### DIFF
--- a/account_mass_reconcile/__manifest__.py
+++ b/account_mass_reconcile/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     "name": "Account Mass Reconcile",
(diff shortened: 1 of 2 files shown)


flotho (approved):
Hi,
Thanks,
Easy code review

flotho (changes requested):
could you change your PR name and commit comment to follow guidelines  https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

thanks

PR #681, opened 2024-08-12
[ADD] account_in_payment

etobella:
This enables in-payment management. Just to be sure that all works well, some tests have been added (most part of the code)


pedrobaeza (changes requested):
I don't think this should be added here, or at least not unconditionally. This has been already added in payment orders in https://github.com/OCA/bank-payment/blob/4034b4bc799c788bdca09b0cf1bee429e1c39b24/account_payment_order/models/account_move.py#L235-L242.

etobella:
In payment orders it was added only for invoices payed with orders. Should be better if we add a parameter (disabled by default)?

pedrobaeza:
First of all, I don't think this should belong here. This module is for reconciling, not for setting something outside the reconciliation interface. You can add a module that enables this unconditionally and then the "setting" is to install it or not depending on your needs.

etobella:
The reason for doing it here was to add in a similar way that it is done by EE. But we can add a new module only for the two lines of code. Most of the changes are tests just for ensuring that everything works fine

pedrobaeza:
Well, the enterprise module mixes all the accounting things together in one module, so that's not a reference to mimic here in OCA. Do a separate module and as said, the decision is taken installing the module or not.

etobella:
@pedrobaeza Comments have been attended.

pedrobaeza (approved):

marcelsavegnago (approved):
Thanks @etobella

PR #683, opened 2024-08-16
[17.0][IMP] account_statement_base + account_reconcile: Forward port missing commits

etobella:
This PR adds several commits from this two modules

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -412,8 +412,9 @@ class AccountBankStatementLine(models.Model):
                 continue
             new_data.append(line_data)
             liquidity_amount += line_data["amount"]
+
         for line in reconcile_model._get_write_off_move_lines_dict(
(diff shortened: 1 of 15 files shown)


pedrobaeza:
Please check several commits that are not relevant, like the pre-commit one only modifying tr.po

etobella:
The only non-relevant commits that don't come from 16.0 are `[FIX] account_mass_reconcile: Fix pre-commit issue` and `[FIX] account_reconcile_oca: Remove warning`. First one is necessary in order to avoid an error on pre-commit but it is not related to any of this modules. Actually, 17.0 branch pre-commit is failing https://github.com/OCA/account-reconcile/actions/runs/10366738074/job/28696666245... Should I create an specific PR for that?
The second one was removing a warning on pre-commit, but I can remove it.

pedrobaeza (approved):
OK then

cvinh (approved):

pedrobaeza:
@etobella there are conflicts. Can you solve then?

PR #684, opened 2024-08-20
[IMP] account_reconcile_oca: Split overflow

etobella:
Right now we have:

[Grabación de pantalla desde 20-08-24 15:13:30.webm](https://github.com/user-attachments/assets/1d35297d-c51d-4223-bb5e-fc26daf8f683)

But we propose:

[Grabación de pantalla desde 20-08-24 15:10:43.webm](https://github.com/user-attachments/assets/f76011f7-ddf7-4e69-b5e9-eaed27e886c3)

### DIFF
--- a/account_reconcile_oca/static/src/scss/reconcile.scss
+++ b/account_reconcile_oca/static/src/scss/reconcile.scss
@@ -11,6 +11,10 @@
                 opacity: 100;
             }
         }
+        .row {
+            // We need to add this in order to make remove horizontal scroll
+            margin: 0;
+        }
         margin: 0 0 0;
         min-width: fit-content;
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Yes, I agree, although I think the ideal one is to "fix" the upper right part, not only the buttons, so while you add the lines, you see the remaining amount and so on. The only risk is that there's a lot of lines to reconcile against and this fixed part avoids to scroll for more.

luisDIXMIT (approved):
Tested on runboat, LGTM!!

pedrobaeza:
/ocabot merge patch

Please fw-port it to 17

PR #690, opened 2024-09-09
[16.0][FIX] account_reconcile_oca: Fix wrong display on reconciled moves

etobella:
Right now we see this

image

And we expect

image

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -442,7 +442,7 @@ class AccountBankStatementLine(models.Model):
             return
         self.partner_id = self.manual_partner_id
 
-    @api.depends("reconcile_data")
+    @api.depends("reconcile_data", "is_reconciled")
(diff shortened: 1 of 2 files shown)


luisDIXMIT (approved):
Tested on runboat, LGTM!

pedrobaeza:
Can you be a bit more specific in the screenshots. I don't see the change...

etobella:
When you approve a move, it gets the wrong information on 16.

On both examples, I reconciled an invoice and in the old option, the reconcile widget shows something weird with three lines but we should have only two. The data was right with 2 lines only

pedrobaeza:
Please explain that on the commit message.

etobella:
Fast tracking, as this is a bug that can confuse users

/ocabot merge patch

pedrobaeza:
Should this be forwarded-ported?

luisDIXMIT:
@pedrobaeza I have done the forward-port to 17.0 #691

PR #694, opened 2024-09-13
[16.0][IMP] account_reconcile_oca: Improve multicurrency management

### LINKED ISSUE #692: account_reconcile_oca : wrong reconciliation of account if foreign currency
Hello @etobella 

We meet some blocking issues when trying the statement reconciliation involving currencies different from the company currency.
Here is a first general issue, I'd like to have your  feedback. I'll create other issues that may be linked to this one, one  way or another.
I join a shell script (at the end on the issue) to easily configure what is needed to reproduce the issue on a demo database with only account_reconcile_oca installed.

Set up (run the script if you prefer) : 
- Active euro currency if not done already
- Create e Bank journal with currency = Euro
- Create a bank statement line in the bank journal in Euro

Then, goes to the reconcile UI, the currency amount of the suspense line is wrong, as it is in USD, should be in euro
1-first-screen-wrong-curr

 At the contrary, at this step, the accounting entry is ok, the suspense line is in EUR
2-accounting-entry-ok

 
 Then if I choose an account in manual operation tab, say the 121000 and the validate : 
3-manual-counterpart

4-reconciled-move-wrong-curr

The accounting entry is now wrong, as the currency of the counterpart is soudainly in company currency (USD).
I expect it to be in EUR, only the account should have changed for the initial accounting entry generated.

I'll gladly help on the matter although I am not (yet) confortable with the technical part of the module, let me know what you think.
Thanks for all your work on this matter.

<details>
<summary>Shell script : </summary>

```
eur = env.ref("base.EUR")
eur.write({'active': True})


eur_journal = env["account.journal"].create({
    "name": "Bank EUR",
    "type": "bank",
    "currency_id": eur.id,
})

eur_statement = env["account.bank.statement"].create({
    "name": "TEST 1 CNY",
    "journal_id": eur_journal.id,
    "line_ids": [(0, 0, {
        "date": "2024-09-10",
        "journal_id": eur_journal.id,
        "payment_ref": "VIR CUSTOMER",
        "amount": 100,
        "currency_id": eur.id,
    })],
})
```

</details>

### LINKED ISSUE #693: [16] account_reconcile_oca : bank account in foreign currency, rounding error for full reconcile
Hello @etobella 

Second issue, linked to https://github.com/OCA/account-reconcile/issues/692 I guess : 
It may be solved with a change on the linked issue, I post it to record it so we can check this case easily then.

Here is the use case (I'll join a shell script at the end allowing to reproduce the issue from a demo database with only account_reconcile_oca installed).

Set up (you can run the shell script)
- Active CNY currency
- Create a bank journal in a foreign currency (CNY) (set the currency on the journal)
- Create a rate for the CNY currency : 0.125989013758 USD per unit for 2024-09-10 and 0.126225969731 USD per unit for 2024-09-09
- Create a statement line on this CNY bank journal with amount 259200
- Create a customer invoice, with currency CNY, and total amount 259200

Then, go to the bank statement reconciliation, and you have this : 
1-before-reconcile

The suspense line does not have any currency visible, like in first issue. I think it is confusing already because we are in a account in CNY and the suspense line should reflect the bank line.

Then when I choose the invoice : 
2-inv-selection

The currency amount appear now, but it is not the right amount (259199.98 instead of 259200). I guess there is some kind of conversion toward usd and then backtoward CNY. 

Therefor the amount is strikethrough which is also confusing because the amount is of the invoice is the  same as the amount received.
And it let me think that the reconciliation will be partial

Once validated, I check the accounting entry : 
3-accounting-entry

I see that the counterpart is not in CNY currency, it has been switched to a full USD line. I expect to have a amount currency of -259200 for the line with account 121000. (same as first linked issue)

Finally on the reconciled invoice : 
4-invoice

I see a payment of 0.02 cny which I don't want.

Like I said it may be solved or partially solved by link issue, I still post it now to show all the problems I found for now.
And once again, thanks for the job on the matter!

<details>
<summary>Shell script : </summary>

```
usd = env.ref("base.USD")
cny = env.ref("base.CNY")
cny.write({'active': True})

env["res.currency.rate"].create({
    "name": "2024-09-10",
    "currency_id": cny.id,
    "inverse_company_rate": 0.125989013758
})
env["res.currency.rate"].create({
    "name": "2024-09-09",
    "currency_id": cny.id,
    "inverse_company_rate": 0.126225969731
})

cny_journal = env["account.journal"].create({
    "name": "Bank CNY",
    "type": "bank",
    "currency_id": cny.id,
})

cny_statement = env["account.bank.statement"].create({
    "name": "TEST 1 CNY",
    "journal_id": cny_journal.id,
    "line_ids": [(0, 0, {
        "date": "2024-09-10",
        "journal_id": cny_journal.id,
        "payment_ref": "VIR CUSTOMER",
        "amount": 259200,
        "currency_id": usd.id,
    })],
})


inv = self.env['account.move'].create({
    "move_type": "out_invoice",
    "partner_id": env.ref("base.res_partner_12").id,
    "invoice_date": "2024-09-09",
    "currency_id": cny.id,
    "invoice_line_ids": [(0, 0, {
        "name": "test",
        "price_unit": "259200",
        "tax_ids": [(6, 0, [])],
    })]
})
inv.action_post()
```

</details>

### LINKED ISSUE #695: [16] account_reconcile_oca : foreign currency in company currency account
Hello @etobella 

Here the use case is, for an journal in company currency (USD) when you have a line with foreign_currency_id and amount_currency set.

Set up  (run bash script at then end)
- Active CNY currency
- Create a rate for the CNY currency : 0.125989013758 USD per unit for 2024-09-10 and 0.126225969731 USD per unit for 2024-09-09
- create a statement line on default bank journal (in company currency USD) on the 2024-09-10, with amount 500, foreign currency : CNY and amount currency 3968.25
- create an customer invoice in CNY currency for an amount of 3968.25

Here screen shot from Odoo enterprise where all seems ok .
Default screen on the reconciliation UI, we see the counterpart in CNY because it is the foreign currency
1-enterprise-foreign-curr

When we select the invoice with same amount in CNY, we have a full reconciliation, with de exchange loss because Odoo rate is not the same as the one from the bank.
2-inv-selection

Then the invoice is fully reconciled : 
3-paid-invoice

Same steps with OCA module : 
In default screen of reconciliation UI, **we do not see the foreign currency amount** but the USD one. It does not help us so choose the right invoice : 
1-oca-foreign-curr

Then when selecting the invoice, we have **a suspense line instead of a income exchange line**, and **the amount in currency is wrong.**
3-reconcile-foreign

Then the reconciliation is impossible, because of unwanted suspense line.

Doing the same test, only changing the date of the invoice from 2024-09-10 to 2024-09-09, the exchange line become an expense, the result is even stranger : 
3reconciliation

The exchange line should be 90 cents instead of 94 and we should not have the last line, which is a suspense line.

<details>
<summary>Shell script : </summary>

```
usd = env.ref("base.USD")
cny = env.ref("base.CNY")
cny.write({'active': True})

env["res.currency.rate"].create({
    "name": "2024-09-10",
    "currency_id": cny.id,
    "inverse_company_rate": 0.125989013758
})
env["res.currency.rate"].create({
    "name": "2024-09-09",
    "currency_id": cny.id,
    "inverse_company_rate": 0.126225969731
})
bank_journal = env['account.journal'].search([('type', '=', 'bank'), ('currency_id', '=', False)], limit=1)

usd_statement = env["account.bank.statement"].create({
    "name": "TEST 1 foreign currency",
    "journal_id": bank_journal.id,
    "line_ids": [(0, 0, {
        "date": "2024-09-10",
        "journal_id": bank_journal.id,
        "payment_ref": "VIR CUSTOMER",
        "amount": 500,
        "currency_id": usd.id,
        "foreign_currency_id": cny.id,
        "amount_currency": 3968.25,
    })],
})


inv = self.env['account.move'].create({
    "move_type": "out_invoice",
    "partner_id": env.ref("base.res_partner_12").id,
    "invoice_date": "2024-09-10",
    "currency_id": cny.id,
    "invoice_line_ids": [(0, 0, {
        "name": "test",
        "price_unit": 3968.25,
        "tax_ids": [(6, 0, [])],
    })]
})
inv.action_post()
```

</details>

etobella:
Tries to fix #692 #693 
@florian-dacosta Can you check it please?

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -1113,8 +1113,8 @@ class AccountBankStatementLine(models.Model):
             "debit": amount if amount > 0 else 0.0,
             "kind": "other",
             "currency_id": self.company_id.currency_id.id,
-            "line_currency_id": self.company_id.currency_id.id,
-            "currency_amount": amount,
+            "line_currency_id": currency.id,
+            "currency_amount": 0,
         }
         reconcile_auxiliary_id += 1
         return reconcile_auxiliary_id, data
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -201,9 +201,17 @@ class AccountBankStatementLine(models.Model):
             new_data = []
             is_new_line = True
             pending_amount = 0.0
+            currency = self._get_reconcile_currency()
             for line in data:
                 if line["kind"] != "suspense":
-                    pending_amount += line["amount"]
+                    pending_amount += currency._convert(
+                        line["currency_amount"],
+                        self.env["res.currency"].browse(
+                            line.get("line_currency_id", currency.id)
+                        ),
+                        self.company_id,
+                        self.date,
+                    )
                 if self.add_account_move_line_id.id in line.get(
                     "counterpart_line_ids", []
                 ):
@@ -212,7 +220,11 @@ class AccountBankStatementLine(models.Model):
                     new_data.append(line)
             if is_new_line:
                 reconcile_auxiliary_id, lines = self._get_reconcile_line(
-                    self.add_account_move_line_id, "other", True, pending_amount
+                    self.add_account_move_line_id,
+                    "other",
+                    True,
+                    max_amount=pending_amount,
+                    move=True,
                 )
                 new_data += lines
             self.reconcile_data_info = self._recompute_suspense_line(
@@ -226,6 +238,7 @@ class AccountBankStatementLine(models.Model):
     def _recompute_suspense_line(self, data, reconcile_auxiliary_id, manual_reference):
         can_reconcile = True
         total_amount = 0
+        currency_amount = 0
         new_data = []
         suspense_line = False
         counterparts = []
@@ -240,10 +253,28 @@ class AccountBankStatementLine(models.Model):
             if line["kind"] != "suspense":
                 new_data.append(line)
                 total_amount += line["amount"]
+                if line.get("currency_amount"):
+                    currency_amount += (
+                        self.env["res.currency"]
+                        .browse(line["line_currency_id"])
+                        ._convert(
+                            line["currency_amount"],
+                            self._get_reconcile_currency(),
+                            self.company_id,
+                            self.date,
+                        )
+                    )
+                else:
+                    currency_amount += self.company_id.currency_id._convert(
+                        line["amount"],
+                        self._get_reconcile_currency(),
+                        self.company_id,
+                        self.date,
+                    )
             else:
                 suspense_line = line
         if not float_is_zero(
-            total_amount, precision_digits=self.currency_id.decimal_places
+            total_amount, precision_digits=self.company_id.currency_id.decimal_places
         ):
             can_reconcile = False
             if suspense_line:
@@ -255,6 +286,7 @@ class AccountBankStatementLine(models.Model):
                     }
                 )
             else:
+
                 suspense_line = {
                     "reference": "reconcile_auxiliary;%s" % reconcile_auxiliary_id,
                     "id": False,
@@ -269,8 +301,8 @@ class AccountBankStatementLine(models.Model):
                     "debit": -total_amount if total_amount < 0 else 0.0,
                     "kind": "suspense",
                     "currency_id": self.company_id.currency_id.id,
-                    "line_currency_id": self.company_id.currency_id.id,
-                    "currency_amount": -total_amount,
+                    "line_currency_id": self.currency_id.id,
+                    "currency_amount": -currency_amount,
                 }
                 reconcile_auxiliary_id += 1
             new_data.append(suspense_line)
@@ -375,7 +407,7 @@ class AccountBankStatementLine(models.Model):
         if self.manual_line_id.exists() and self.manual_line_id:
             self.manual_amount = self.manual_in_currency_id._convert(
                 self.manual_amount_in_currency,
-                self.company_id.currency_id,
+                self._get_reconcile_currency(),
                 self.company_id,
                 self.manual_line_id.date,
             )
@@ -529,7 +561,10 @@ class AccountBankStatementLine(models.Model):
         reconcile_auxiliary_id = 1
         for line in liquidity_lines:
             reconcile_auxiliary_id, lines = self._get_reconcile_line(
-                line, "liquidity", reconcile_auxiliary_id=reconcile_auxiliary_id
+                line,
+                "liquidity",
+                reconcile_auxiliary_id=reconcile_auxiliary_id,
+                move=True,
             )
             data += lines
         if not from_unreconcile:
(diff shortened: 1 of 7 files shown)


florian-dacosta:
@etobella Thanks for the PR!
Well, it fix the example I gave in the issue yes, thanks, but it does not fix all cases, like the one in the second issue.
Let me show you what I have with the same setup from https://github.com/OCA/account-reconcile/issues/693

Default screen when going to reconcile UI, we have a rounding issue
1-bfore-reconcile

After entering the 121000 account in manual tab : 
2-accounting-entry

etobella:
@florian-dacosta Round 2
image
I added a rounding factor in order to remove this possible error (the rounding should only be used on comparison)

florian-dacosta:
@etobella I still don't get why there is a rounding to find an amount that is written in the `amount` field of the `account.bank.statement.line`. Why the 259 200 needs to be computed  and not read for the line ?

Anyway, I've tested the later version
I still get this rounding issue before chosing the invoice.
And once validated, the state /payment on the invoice is not ok.

Go to reconcile UI - Not OK
1-rec-ui

Choose the invoice - OK
2-after-choosing-invoice

Reconciled invoice after validation - Not OK
3-invoice-partial

The bank accounting entry is missing the currency on the 121000 line
4-accounting-entry

etobella:
We have a rounding issue because the reconcile is done on the currency of the company. Also we cannot use that amount, because we might have previous lines.

The only option I can see is to do the reconcile in the amount of the statement line.

This is more work to do and I just wanted to add a patch, but seems we need to do a real refactoring (it is good to have so many tests)

PaulGoubert:
Thank you for your help with these issues, @etobella.
As previously mentioned by @florian-dacosta, we're in a rush on this topic due to our customer's situation—they've been expecting to use it properly for 6 months now. I hope you can tackle these topics and help us resolve the issues.

etobella:
@PaulGoubert @florian-dacosta Round 3

I have changed the flow in order to compare using the currency of the line. It has no sense to move from EUR to USD to EUR, so it should remove some of this errors.

etobella:
The only issue I found is that it is not showing properly when reconciling with the Foraign Exchange Loss

Before Reconciling
image
 
After reconciling
image

The foreign loss has been lost, but it is correctly reconciled 
image

etobella:
I applied a fix for this. I am not sure if it gets all the possible cases, but at least it is an approach

florian-dacosta:
@etobella Thanks for your work, it is really helpull !
About the rounding case, with last version of the code, here is what I have : 
1) First click on the reconcile button on journal 
1-start-reconciliation

2) It give me the reconciliation screen with the rounding error in the suspense account : 
2-before-begining-reconcile-ui

If I choose the counter part, all is fine, then when I reconcile, as you saif, all is fine, the invoice is reconciled correctly, the accounting entries are correct.

3) Right after, the view of the statement line in the reconciliation UI was correct. I am not sure how it work, but with a new tab/after restarting the server and going back on the statement line, I have really strange data : 
3-after-reconciled-view

For now the most annoying for me is the display in 1. Even if all goes fine after, it is very confusing for the user.



> I applied a fix for this. I am not sure if it gets all the possible cases, but at least it is an approach

I am not sure what you refer to ?
If it is about the last created issue (695), I have just updated it because I had made a mistake in the screen shot of the use case.
I have tested again right now, and it is still actual.

etobella:
1st point should be solved now.

florian-dacosta:
> 1st point should be solved now.

It is, thanks
I have made a PR on yours to fix different other issues, Take a look when you can

florian-dacosta:
@etobella Could you rebase please ?

account_reconcile_oca/models/account_bank_statement_line.py
```diff
@@ -291,7 +332,9 @@ def _check_line_changed(self, line):
             or self.manual_account_id.id != line["account_id"][0]
             or self.manual_name != line["name"]
             or (
-                self.manual_partner_id and self.manual_partner_id.name_get()[0] or False
+                self.manual_partner_id
+                and self.manual_partner_id.name_get()[0]
+                or [False, False]
```
yvaucher, line 337: `bool([False, False])` => `True`

Are you sure you want this?

florian-dacosta: I think we compare this result to [False, False] yes, but I will check again

PabloMartFL (approved):
Tested the PR on runboat and it resolves the problems in the issues:

- For the test, I created a company in EUR and created a Bank in EUR.
- The currency rate are the following:
- 
Captura desde 2024-10-04 12-23-36

The suspended balance apears correct and when you select the invoices that fully match with the statement, it doesnt correct the invoice amount.
Also, in the account move, the amount in currency is displayed correcly

https://www.awesomescreenshot.com/video/32216430?key=6538e7535b46717a2905cfcb08d9a052

account_reconcile_oca/models/account_bank_statement_line.py
florian-dacosta: If, from the reconciliation UI, for any bank statement line (which is not linked to any partner) you click on the counter part line, you can see that the onchange is called, and in the values of `reconcile_data_info` field, we have 
`[False, False]` in the `partner_id` key (in `data` list).
Then, if we compare here `False` with `[False, False]`, like it was before, it will be considered as a change where it is actually the same (there is no partner_id !).

That is why I did adapt here but I agree it is not nice and it could be an issue, since Odoo usually put `False` when there are not partner : https://github.com/OCA/account-reconcile/blob/16.0/account_reconcile_oca/models/account_reconcile_abstract.py#L69

But I don't know why the UI send `[False, False]` in the onchange, nor if it could be changed.
Any idea @etobella @yvaucher ?

florian-dacosta: Well, it actually comes from https://github.com/OCA/account-reconcile/blob/17.0/account_reconcile_oca/models/account_bank_statement_line.py#L891

I'll do a commit to propose a change about it.

florian-dacosta: Sorry for the noise here.
I'll make it short here. When there are not partner at all, in the data, the value  `[False, False]` is set here : 
https://github.com/OCA/account-reconcile/blob/16.0/account_reconcile_oca/models/account_bank_statement_line.py#L405
and here : 
https://github.com/OCA/account-reconcile/blob/16.0/account_reconcile_oca/models/account_bank_statement_line.py#L896

So when checking if the line has changed, and there is no partner at all, comparing the reconcile_data partner part (`[False, False]`)with `False` will always be considered different while no changed occured (no partner was set).

So I think we could leave this as is. (meaning comparing` [False, False]` with `line.get("partner_id")`

ljsalvatierra-factorlibre:
Hi @etobella can you do a rebase?

It has conflicts with https://github.com/OCA/account-reconcile/pull/758

account_reconcile_oca/models/account_bank_statement_line.py
```diff
@@ -561,16 +612,87 @@ def _default_reconcile_data(self, from_unreconcile=False):
...
+            if partial_lines:
+                for reconciled_line in (
+                    partial_lines.debit_move_id + partial_lines.credit_move_id - line
+                ):
+                    if (
+                        reconciled_line.move_id.journal_id
+                        == self.company_id.currency_exchange_journal_id
+                    ):
+                        reconcile_auxiliary_id, lines = self._get_reconcile_line(
+                            reconciled_line.move_id.line_ids - reconciled_line,
```
ljsalvatierra-factorlibre, line 650: Expected singleton.

```
>>> reconcile_auxiliary_id, lines = self._get_reconcile_line(
>>> new_vals = super()._get_reconcile_line(
>>> original_amount = amount = net_amount = line.debit - line.credit
>>> record.ensure_one()
```

ljsalvatierra-factorlibre (changes requested):

pedrobaeza (approved):
Tested on real cases and working.

/ocabot merge patch

PR #699, opened 2024-09-18
[16.0][IMP] account_reconcile_oca: Only fix partner on payment accounts

etobella:
Without the changes, if we have several Lines on the same account but different partners, we cannot reconcile them together. This is correct for Receivable and Liability accounts, but not on the others.

If we have this

image

We have 

image

But we expect

image


pedrobaeza (approved):
This is the solution to show the pending moves for accounts that are not AR/AP, so merging it:

/ocabot merge patch

PR #707, opened 2024-09-25
[16.0][FIX] account_statement_base: Allow to create new manual transactions in tree view

victoralmau:
Allow to create new manual transactions in tree view

Related to https://github.com/OCA/account-reconcile/pull/678

**Before**
antes

**After**
despues

Please @pedrobaeza and @carlosdauden can you review it?

@Tecnativa TT50970


aritzolea (approved):
I was just going to report the issue. Thank you for the fix!

pedrobaeza (approved):
/ocabot merge minor

Both the change that introduced this problem and this fix should be put in 17.0

PR #709, opened 2024-09-27
[16.0] [IMP] Show paid amount on move lines for the selected statement line

ByteMeAsap:
This change would allow users to see the amount paid for move lines through the selected statement line
Paid Amount On Move Lines


pedrobaeza (changes requested):
For me the important thing is the residual amount, not the paid amount, and this is adding more burden to the already crowded `account.move.line` model, which is critical for the general performance. Moreover, this is something sided to this module itself, as the field is added to the general model. I think you should develop an extra module for this feature to be optional for those wanting to use it.

PR #710, opened 2024-09-27
[16.0] [IMP] Show reconcile button even if there no entries for reconciliation through configuration

ByteMeAsap:
Adds a option in the configuration to show reconcile button even if there no entries for reconciliation

Configuration Settings

If the option is enabled , it will show the reconcile button even if there are no entries to reconcile

Reconcile Button

### DIFF
--- a/account_reconcile_oca/models/account_journal.py
+++ b/account_reconcile_oca/models/account_journal.py
@@ -31,3 +31,21 @@ class AccountJournal(models.Model):
         if self.get_journal_dashboard_datas()["number_to_reconcile"] > 0:
             return False
         return _("Well done! Everything has been reconciled")
+
+    def _fill_bank_cash_dashboard_data(self, dashboard_data):
+        """Populate all bank and cash journal's data dict with
+        relevant information for the kanban card."""
+        super()._fill_bank_cash_dashboard_data(dashboard_data)
+        bank_cash_journals = self.filtered(
+            lambda journal: journal.type in ("bank", "cash")
+        )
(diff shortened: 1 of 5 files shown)


pedrobaeza:
Instead of a configuration option, I think we should provide a permanent link to "Reconcile" in the 3 dots, and thus, show the button if there are entries to reconcile, and go through the dropdown if not.

ByteMeAsap:
@pedrobaeza , there is already a link for reconcile(the Transactions link does the same thing) in the 3 dots, the idea here is to minimise the number of clicks for the users

ByteMeAsap:
@pedrobaeza , there is already a link for reconcile(the Transactions link does the same thing) in the 3 dots, the idea here is to minimise the number of clicks for the users

pedrobaeza:
Then I'm not sure to make the base module more complex adding these kind of options. Maybe an extra module where the option is to not install it?

@ByteMeAsap what is the reasoning behind wanting to go to the reconcile interface without pending entries?

What do you think, @etobella ?

ByteMeAsap:
> @ByteMeAsap what is the reasoning behind wanting to go to the reconcile interface without pending entries?

@pedrobaeza , there are 2 reasons to have this, one being users can use this to create new transactions directly and the other one being allowing them to check already reconciled entries

pedrobaeza:
OK, what about showing a button "New transaction" instead when no entries pending?

Anyway, I would like to have the opinion of @etobella and @JordiBForgeFlow

PR #711, opened 2024-09-27
[16.0] [FIX] account_reconcile_oca: Allow counterpart entries for invoice matching reconciliation models

ByteMeAsap:
There is a problem with auto validation of invoices when we use the Payment Tolerance
Consider the following example:
Suppose the invoice matching model is setup with payment tolerance as shown in the picture:

Reconcile Model Setup

The current module would generate entries as follows:

Current Entries

Expected Behaviour is it should generate entries as follows:

Expected Behaviour

This PR would fix this scenario


brambuijs:
@etobella @ByteMeAsap This would be very usefull. In the Netherlands the 1 cent coin is not used anymore. Many businesses use 4,99 to soften the price of their products. When you pay with 5 euro's this 1 cent difference needs to be accounted for.

PR #659, opened 2024-05-24
[16.0][IMP] account_statement_base: link to account move lines from statements

sergiobstoj:
Link to Account Move Lines in Statements

PR #713, opened 2024-09-27
[17.0][IMP] account_statement_base: add smart button on view_bank_statement_form linking to journal items

sergiobstoj:
fw-port from https://github.com/OCA/account-reconcile/pull/659

@pedrobaeza


pedrobaeza (approved):
Thanks for fw-porting it. Fast-tracking it as straight-forward:

/ocabot merge minor

PR #715, opened 2024-09-30
[16.0] [IMP] account_reconcile_oca: adds reference field on reconcile form view

ByteMeAsap:
This PR adds the reference on the reconcile screen for the users to be able to see additional info for the bank statement lines without having to open the bank statement line move

Bank Statement Line Form VIew

Reference info on reconcile screen

### DIFF
--- a/account_reconcile_oca/views/account_bank_statement_line.xml
+++ b/account_reconcile_oca/views/account_bank_statement_line.xml
@@ -319,8 +319,11 @@
                         </group>
                     </page>
                     <page name="narration" string="Narration">
-                        <field name="payment_ref" />
-                        <field name="narration" />
+                        <group>
+                            <field name="ref" />
+                            <field name="payment_ref" string="Payment Ref" />
+                            <field name="narration" />
+                        </group>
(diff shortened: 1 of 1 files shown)


yvaucher:
The downside here to me is that the user may be confused on what the difference is between a Reference and a Label

Especially when Label is a reference.

I think that using better field labels could help to disambiguate the fields by being more explicit.

ByteMeAsap:
> The downside here to me is that the user may be confused on what the difference is between a Reference and a Label
> 
> Especially when Label is a reference.
> 
> I think that using better field labels could help to disambiguate the fields by being more explicit.

@yvaucher , can you suggest what the field labels should be?

yvaucher:
Just a thought, I didn't push the request changes button.

Would it be better to have the payment ref called "Payment ref" explicitly?

ByteMeAsap:
> Just a thought, I didn't push the request changes button.
> 
> Would it be better to have the payment ref called "Payment ref" explicitly?

@yvaucher , yes that seems logical to me

PR #720, opened 2024-10-01
[16.0] [IMP] account_reconcile_oca: pass default name for lines created through reconciliation models

ByteMeAsap:
This PR aims to pass a default name/label(same as the label on the bank statement line) for the new line that is created through reconciliation models if no label is passed through the reconciliation model

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -481,16 +481,19 @@ class AccountBankStatementLine(models.Model):
     def _reconcile_data_by_model(self, data, reconcile_model, reconcile_auxiliary_id):
         new_data = []
         liquidity_amount = 0.0
+        default_name = ""
         for line_data in data:
             if line_data["kind"] == "suspense":
                 continue
             new_data.append(line_data)
             liquidity_amount += line_data["amount"]
-
+            if line_data["kind"] == "liquidity":
+                default_name = line_data["name"]
         for line in reconcile_model._get_write_off_move_lines_dict(
             -liquidity_amount, self._retrieve_partner().id
         ):
             new_line = line.copy()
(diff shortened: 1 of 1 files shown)


gbrito (approved):

pedrobaeza:
Any screenshot of the before and after?

ByteMeAsap:
> Any screenshot of the before and after?

@pedrobaeza ,consider a reconciliation model 'Test' as can be seen in following screenshot.There is no 'Journal Item Label' passed.

Screenshot From 2024-12-18 19-32-05

So currently, it passes a blank label on the reconciliation line when the reconciliation model is selected, as can be seen in the following screenshot.

Screenshot From 2024-12-18 19-31-22

After this PR,the default label(same as the label on the bank statement line) would be passed on the reconciliation line as can be seen in following screenshot

Screenshot From 2024-12-18 19-31-53

pedrobaeza:
OK, understood. Thanks for the explanations. Please include that "Steps to reproduce the problem" in the commit message, omitting images, but explaining them by text.

ByteMeAsap:
@pedrobaeza ,done

ByteMeAsap:
> OK, understood. Thanks for the explanations. Please include that "Steps to reproduce the problem" in the commit message, omitting images, but explaining them by text.

@pedrobaeza ,done

pedrobaeza (approved):
Well, the commit message allows multiple lines, and the explanation should be put in consecutive lines, not everything in one line, but OK, I don't want to be more picky.

Please fw-port it to 17 using the expected commit message guidelines (with one first short line, and a detailed explanation in subsequent lines):

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

/ocabot merge patch

PR #721, opened 2024-10-01
[16.0] [FIX] account_reconcile_oca: Fix the layout of statement button on reconciliation screen

ByteMeAsap:
This Pr aims to fix the layout of statement button on reconciliation screen

Currently, the statement button cuts out some portion of the other parts as seen in following image:

Current Button

After this PR, the button would look as follows:

Button After Changes

### DIFF
--- a/account_reconcile_oca/static/src/scss/reconcile.scss
+++ b/account_reconcile_oca/static/src/scss/reconcile.scss
@@ -20,9 +20,9 @@
         width: 100%;
         .o_reconcile_create_statement {
             position: absolute;
-            height: 4px;
+            height: 0px;
             margin: 0;
-            padding: 2px 0 0 0;
+            padding: 0 0 0 0;
             border: 0;
             top: -14px;
             opacity: 0;


gbrito (approved):

account_reconcile_oca/views/account_bank_statement_line.xml
```diff
@@ -19,20 +19,16 @@
                 <field name="aggregate_name" />
                 <templates>
                     <t t-name="kanban-box">
-                        <div
```
flotho: personnaly I found the use of the div more visible on you scree.

flotho (approved):
small comment, otherwise LGTM

account_reconcile_oca/views/account_bank_statement_line.xml
ByteMeAsap: @flotho ,sorry didn't get you. Can you please elaborate?

ByteMeAsap:
Can we have this merged?

pedrobaeza (changes requested):
The fix is incorrect, as the label "STATEMENT" should be in the middle of the line as it was before. If it was overlapping the other text, then you should add height to that section (or compose it other way).

ByteMeAsap:
> The fix is incorrect, as the label "STATEMENT" should be in the middle of the line as it was before. If it was overlapping the other text, then you should add height to that section (or compose it other way).

@pedrobaeza , I have made changes, please review

pedrobaeza:
Can you please put an image of the new after?

ByteMeAsap:
After changes 

After Changes

pedrobaeza (approved):
OK, that looks better, thanks.

Please fw-port it to 17.

/ocabot merge patch

PR #722, opened 2024-10-02
[16.0][ADD] account_reconcile_payment_order

etobella:
@alexis-via I think you will enjoy this one :wink: 

Still some work to do on documentation, but we have something to start


etobella:
image
How it is seen

etobella:
I even added a post init hook in order to generate data.

@alexis-via WDYT about it?

pedrobaeza:
I told Alexis my take on this, and I don't think we should go this way.

My idea is to enable groups on the list view, and adding a button on the group headers to add the entire group lines in the current reconciliation. For payment orders, we will only need to group by the proper fields (and maybe add a related stored field in aml). This is also more versatile for other kind of mass additions.

etobella:
@pedrobaeza You are suggesting something like this?
image

pedrobaeza:
Well, more or less, but the group feature should be available on that "mini-view", and the button "ADD ALL" should be in each group. There are similar action buttons on v15 journal views (that are grouped by entry in this case):

imagen

etobella:
I checked your proposal and it might work partially, but there are some related issues:

It is not filtering properly. I mean, when you press the button, you are not getting data from all the childs, you are processing the data of the record.

Let's put an examples

- We group by Payment Order and Select one, it will work perfectly :+1: 
- We group by Payment Order, filter by a Partner and select one, it will select all the lines of the payment order. It will not get the proper data. :-1: 

Also, we will not be able to split data properly on a payment order if multiple dates exist.

So, IMO, it is not feasible using only standard odoo functionality :thinking: 

BTW, it looked as expected (I made the test using Journal, but you can see how it would look like):

image

pedrobaeza:
I understand the issue you have found, but I think we can provide a combination:

- You can stack groups, so that's give more granularity.
- If the group is missing any element because a filter, don't show the button.
- Have the button on the top also for including the "visible items" as you have it at the beginning.
- Include favorites for saving groups (or maybe provide them pre-fixed).

With all of this, we will have enough tools for doing reconciliation in batches. A small glue module providing the fields to group will be the only need on that matter.

pedrobaeza:
The feature added in #798 may help as an initial workaround for this use case: you pre-filter the entries (for example, searching for the payment order name "PAY0001"), and then clicking in "Add all" for not having to click one by one each line. You can remove extra lines that don't match, but you are reducing a lot the work.

The cases where there are several due dates is still not covered, but at least you have more tools to work on.

etobella:
I will not finish this.

PR #731, opened 2024-10-17
[16.0][FIX] account_reconcile_oca: Only filter by partner on receivable and payable lines

etobella:
Still testing, but should fix the manual reconciliation of non payable/receivable lines

@pedrobaeza


pedrobaeza (approved):
Tested and working.

pedrobaeza:
Let's fast-track it to fix the problem. If any additional problem found, it can be fixed in later PRs:

/ocabot merge patch

PR #733, opened 2024-10-17
[17.0][FIX] Allow reconciliation of general journal items

cvinh:
This PR fixes the query for allowing reconciliation of journal items

Steps to reproduce : 
- enter a journal entry with 2 journal items with accounts that are "allow reconciliation = True"
- clic on the menu Accoounting/Actions/Reconcile

Wrong behavior 
- no items are shown
 
Right behavior 
- items should be shown and allow their reconciliation 

cc @alexis-via that's what I was talking about at OXP

### DIFF
--- a/account_reconcile_oca/models/account_account_reconcile.py
+++ b/account_reconcile_oca/models/account_account_reconcile.py
@@ -91,7 +91,7 @@ class AccountAccountReconcile(models.Model):
         return """
             HAVING
                 SUM(aml.debit) > 0
-                AND SUM(aml.credit) > 0
+                OR SUM(aml.credit) > 0
         """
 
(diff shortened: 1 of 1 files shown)


cvinh:
I dont understand why test fails

cvinh:
Hello
Can you please reopen ?
This PR is needed to reconcile journal items without partner
Thanks

etobella:
The logic of reconciliation has changed to an active field. This PR will not work with master

cvinh:
> The logic of reconciliation has changed to an active field. This PR will not work with master

Thanks for replying
BTW maybe I've missed something, I don't find how to reconcile entries without partners except by hacking the query...

PR #736, opened 2024-10-23
[16.0] account_move_line_reconcile_manual : writeoff_journal_id initialisation

emiliesoutiras:
Currently in the interface, the selection list displays the first item in the list, but as the value is not initialized, it makes an alert empty value error when visually it's not.

### DIFF
--- a/account_move_line_reconcile_manual/wizards/account_move_line_reconcile_manual.py
+++ b/account_move_line_reconcile_manual/wizards/account_move_line_reconcile_manual.py
@@ -120,8 +120,7 @@ class AccountMoveLineReconcileManual(models.TransientModel):
                 journals = self.env["account.journal"].search(
                     [("type", "=", "general"), ("company_id", "=", wiz.company_id.id)]
                 )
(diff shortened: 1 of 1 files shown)


Kev-Roche (approved):
LGTM

florian-dacosta:
@emiliesoutiras Could you provide a screenshot of the current issue?
I don't get what is wrong, I did a test without the fix and I don't see the first journal displayed is the field, but I may have mis-understood the issue
manual-reconcile-write-off

emiliesoutiras:
The correct operation  seems to come from an Odoo [commit](https://github.com/odoo/odoo/commit/a3812be90017a921df3cfeb4af48a04a75caa315) dated October 28, 2024, so this modification is no longer necessary.

emiliesoutiras:
The correct operation seems to come from an Odoo [commit](https://github.com/odoo/odoo/commit/a3812be90017a921df3cfeb4af48a04a75caa315) dated October 28, 2024, so this modification is no longer necessary.

Issue #497, opened 2022-10-03
Migration to version 16.0


sbidoul:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0

# Modules to migrate

- [ ] account_mass_reconcile - By @sonhd91 - #517
- [x] account_move_reconcile_forbid_cancel - By @ACheung-FactorLibre - #568
- [x] account_move_reconcile_helper - By @mpascuall - #771
- [x] account_partner_reconcile - By @yibudak - #791
- [ ] account_reconcile_restrict_partner_mismatch - By @sbejaoui - #596
- [ ] account_reconciliation_widget
- [x] bank_statement_check_number - By @AlexPForgeFlow - #573

Missing module? Check https://github.com/OCA/maintainer-tools/wiki/%5BFAQ%5D-Missing-modules-in-migration-issue-list

TeoGoddet:
Regarding account_reconciliation_widget, it seems a lot of works need to be done. 
Maybe a full refactoring using the new backend functions and owl would be better.
The entreprise module switched to more standard views (Heavily customized kanban) 
Would be awesome if we could have the chatter in the widget

pedrobaeza:
Although that refactoring would be good, an initial step should be to adapt existing one for the few missing pieces in v16. The old JS engine is still there, so only bridges or little things like m2o widget should be needed, but not indepth analysis by my part yet.

The rest can come after this as a 2.0 version.

sonhd91:
Please add module `account_mass_reconcile` into module list
`account_mass_reconcile`: https://github.com/OCA/account-reconcile/pull/517

sonhd91:
`account_payment_mode_auto_reconcile` : #537

DMopen2s:
My company is in the process of implementing Odoo v16 community, and we need your help in migrating the account_reconciliation_widget module from v15 to v16, we have an economic budget for the developer to help us in this work, and the module would be released in the community version 16, please if there is someone who can help us, I would greatly appreciate it.
Greetings.

pedrobaeza:
Check #500 for the great replacement.

DMopen2s:
Thank you very much @pedrobaeza, I will review and if I can contribute something, I will inform you.

rrebollo:
@sbidoul I'm taking account_bank_statement_reopen_skip_undo_reconciliation at #737

mpascuall:
[...]

PR #737, opened 2024-11-06
[16.0] [MIG] account_bank_statement_reopen_skip_undo_reconciliation

rrebollo:
Upgrade to v 16.0


rrebollo:
It looks like this is not needed anymore, or at least it needs to be done in a way that might be better a fresh start.

There isn't a "Reset to New" feature for account.bank.statement anymore (CE) although at account.bank.statement.line there's now an action_undo_reconciliation method which looks like the natural evolution from v15 button_undo_reconciliation.

Some code from here (views) looks like it went to account_statement_base.

What do U think @alexis-via @MiquelRForgeFlow ? If I'm right maybe @sbidoul can update #497 adding account_bank_statement_reopen_skip_undo_reconciliation to the migration roadmap crossed out like maintainer tools wiki points out.

PR #596, opened 2023-09-29
[16.0][MIG] account_reconcile_restrict_partner_mismatch

sbejaoui:
based on 

- #565

PR #741, opened 2024-11-12
[18.0][MIG] account_reconcile_restrict_partner_mismatch: Migration to 18.0 

LucasTran380381:
Base on https://github.com/OCA/account-reconcile/pull/596


henrybackman (approved):
Can combine translation commits

LucasTran380381:
> Can combine translation commits

will squash

account_reconcile_restrict_partner_mismatch/models/account_move_line.py
```diff
@@ -0,0 +1,38 @@
...
+        """
+        self.ensure_one()
+        return bool(
+            self.company_id.restrict_partner_mismatch_on_reconcile
+            and not self.journal_id.no_restrict_partner_mismatch_on_reconcile
+            and self.account_id.account_type
+            in ("asset_receivable", "liability_payable")
+        )
+
+    def reconcile(self):
```
grindtildeath: This function is not called when using this module with OEE. You should instead hook on `account.move.line._reconcile_plan`

grindtildeath (changes requested):

account_reconcile_restrict_partner_mismatch/models/account_move_line.py
LucasTran380381: I changed method `_reconcile_plan` to fixed with Odoo EE

account_reconcile_restrict_partner_mismatch/models/account_move_line.py
```diff
@@ -0,0 +1,38 @@
...
+            and not self.journal_id.no_restrict_partner_mismatch_on_reconcile
+            and self.account_id.account_type
+            in ("asset_receivable", "liability_payable")
+        )
+
+    def _reconcile_plan(self, reconciliation_plan):
+        # to be consistent with parent method
+        if self:
+            partners = set()
+            for line in self:
```
grindtildeath: `_reconcile_plan` is an `api.model` function, you should loop over the `reconciliation_plan` var instead (cf docstring to know how it must be structured)

grindtildeath (changes requested):
Would be good to improve the test because having them green is obviously wrong right now.

Also, functionally speaking, I don't see a test validating two move lines with same partner getting reconciled, and it would be great to use different journals. (that is not strictly required for the mig but maybe add a comment in the roadmap)

account_reconcile_restrict_partner_mismatch/models/account_move_line.py
LucasTran380381: improved test-case and loop over reconciliation_plan as suggested

grindtildeath (approved):
Thanks you, that seems to work now :+1:

florian-dacosta (approved):
Code review + tests

bguillot:
Thanks !

PR #744, opened 2024-11-15
[16][FIX] account_reconcile_oca : keep aml label when empty on manual write-off reconcile model

florian-dacosta:
Before the PR : 
Create a reconciliation model with type "Button to generate counterpart entry" and one line with journal label item field as empty.
image

Then, reconcile a statement line using this reconcile model, the accounting entry will have an empty name.
image

After the PR : 
The accounting entry coming from the manual reconcile model will have the same name as the statement line.
image

In previous versions, the name from the statement line was kept.
After a quick check on enterprise, it is also taken from the statement line if empty on the reconciliation model.

@hparfr @etobella


hparfr (approved):

etobella (approved):
/ocabot merge patch

Can you por the change to 17 please?

PR #748, opened 2024-11-18
[17][FIX] account_reconcile_oca : keep partner line case of manual button reconcile

florian-dacosta:
Forward port https://github.com/OCA/account-reconcile/pull/747


etobella (approved):
/ocabot merge patch

PR #753, opened 2024-11-22
[16.0][ADD] account_reconcile_no_numerical_token

thomaspaulb:
By default, when Odoo encounters a bank statement line ref such as a34bc3xxx1, it will extract the numerical characters and concatenate them into a numerical token, eg. 3431 in this case. If then a move line is found that has for example cxxx343y1 as a ref, it will match with this one also: not on the text ref but via the numerical component.

Sometimes this is not desirable.

This module allows to configure a checkbox on a reconcile model to disable numerical token matching for this model.


pedrobaeza:
If this has to be with the modules, then it should be `account_reconcile_model_no_numerical_token`.

thomaspaulb:
I agree that it's not that widely useful since the whole model moves to EE in 17, but at least for some of our 16 instances that use the regular Odoo reconcilation this is useful, and I can imagine it could also be backported to earlier versions.

PR #758, opened 2024-11-26
[16.0][ADD] account_reconcile_analytic_tag: New module

victoralmau:
New module

ejemplo

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT51885


account_reconcile_analytic_tag/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2024 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Account reconcile analytic tag",
```
pedrobaeza:
```suggestion
    "name": "Analytic tags in account reconciliation",
```

account_reconcile_analytic_tag/models/account_bank_statement_line.py
```diff
@@ -0,0 +1,37 @@
+# Copyright 2024 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+from odoo import api, fields, models
+
+
+class AccountBankStatementLine(models.Model):
+    _inherit = "account.bank.statement.line"
+
+    analytic_tag_ids = fields.Many2many(
```
pedrobaeza: Shouldn't it be this name following current fields nomenclature?

```suggestion
    manual_analytic_tag_ids = fields.Many2many(
```

pedrobaeza (commented):
Tested and working. Just a few code remarks

account_reconcile_analytic_tag/models/account_bank_statement_line.py
victoralmau: The analytical distribution field does not have the manual prefix but if the name is consistent, it is already changed.

pedrobaeza (approved):

chienandalu (approved):
Functional tests :+1:

PR #761, opened 2024-11-27
[16.0][IMP] account_reconcile_analytic_tag: Add analytic tags to reconciliation model lines

victoralmau:
Add analytic tags to reconciliation model lines

Related to https://github.com/odoo/odoo/pull/188808

@Tecnativa TT51885

### DIFF
--- a/account_reconcile_analytic_tag/models/account_reconcile_model.py
+++ b/account_reconcile_analytic_tag/models/account_reconcile_model.py
@@ -1,0 +1,47 @@
+# Copyright 2024 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+from odoo import fields, models
+
+
+class AccountReconcileModel(models.Model):
+    _inherit = "account.reconcile.model"
+
+    # TODO: Remove if merging https://github.com/odoo/odoo/pull/188808
+    def _get_write_off_move_lines_dict(self, residual_balance, partner_id):
+        res = super()._get_write_off_move_lines_dict(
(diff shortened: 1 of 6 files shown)


pedrobaeza:
Isn't possible to do this without the hook?

victoralmau:
Changes done not to use the hook (although it is recommended that it exists).

account_reconcile_analytic_tag/models/account_reconcile_model.py
```diff
@@ -0,0 +1,49 @@
...
+    _inherit = "account.reconcile.model"
+
+    # TODO: Remove if merging https://github.com/odoo/odoo/pull/188808
+    def _get_write_off_move_lines_dict(self, residual_balance, partner_id):
+        res = super()._get_write_off_move_lines_dict(
+            residual_balance=residual_balance, partner_id=partner_id
+        )
+        if len(res) == 0:
+            return res
+        key = 0
```
pedrobaeza: For usual nomenclature (`key` is used in dictionaries), this should be:

```suggestion
        index = 0
```

Anyway, look for the next suggestion for saving some lines.

account_reconcile_analytic_tag/models/account_reconcile_model.py
```diff
@@ -0,0 +1,49 @@
...
+        currency = self.company_id.currency_id
+        for line in self.line_ids:
```
pedrobaeza:
```suggestion
        for index, line in enumerate(self.line_ids):
```

pedrobaeza (approved):
/ocabot merge minor

PR #762, opened 2024-12-10
[16.0][IMP] account_reconcile_oca: Add auto-reconcile compatibility (Example: Rule to match invoices/bills)

victoralmau:
Add auto-reconcile compatibility (Example: Rule to match invoices/bills)

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT52146

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -532,7 +532,14 @@ class AccountBankStatementLine(models.Model):
             res = (
                 self.env["account.reconcile.model"]
                 .search(
-                    [("rule_type", "in", ["invoice_matching", "writeoff_suggestion"])]
+                    [
+                        (
+                            "rule_type",
+                            "in",
+                            ["invoice_matching", "writeoff_suggestion"],
+                        ),
+                        ("company_id", "=", self.company_id.id),
+                    ]
                 )
                 ._apply_rules(self, self._retrieve_partner())
             )
@@ -555,6 +562,8 @@ class AccountBankStatementLine(models.Model):
                     )
                     amount -= sum(line.get("amount") for line in line_data)
                     data += line_data
+                if res.get("auto_reconcile"):
+                    self.reconcile_bank_line()
                 return self._recompute_suspense_line(
                     data,
                     reconcile_auxiliary_id,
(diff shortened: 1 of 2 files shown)


account_reconcile_oca/models/account_bank_statement_line.py
```diff
@@ -745,6 +754,7 @@ def create(self, mvals):
         models = self.env["account.reconcile.model"].search(
             [
                 ("rule_type", "in", ["invoice_matching", "writeoff_suggestion"]),
+                ("company_id", "in", result.mapped("company_id").ids),
```
pedrobaeza: With relation fields, no need to put `mapped`:

```suggestion
                ("company_id", "in", result.company_id.ids),
```

account_reconcile_oca/tests/test_bank_account_reconcile.py
```diff
@@ -10,6 +10,16 @@ class TestReconciliationWidget(TestAccountReconciliationCommon):
     @classmethod
     def setUpClass(cls, chart_template_ref=None):
         super().setUpClass(chart_template_ref=chart_template_ref)
+        # Auto-disable reconciliation model created automatically with
+        # generate_account_reconcile_model() to avoid side effects in tests
+        cls.invoice_matching_model = cls.env["account.reconcile.model"].search(
```
pedrobaeza: If it can be more than one, the variable name is not correct

account_reconcile_oca/tests/test_bank_account_reconcile.py
```diff
@@ -999,6 +999,7 @@ def test_partner_name_with_parent(self):
         )
 
         self.invoice_matching_model.active = True
+        self.invoice_matching_model.match_text_location_label = False
```
pedrobaeza: Why a separated commit just for this?

victoralmau: It is already unified (with squash) in one commit, it was separated because it fixed an existing test error related to the change explained in the commit message.

account_reconcile_oca/tests/test_bank_account_reconcile.py
victoralmau: I have changed the variable name to be consistent. As a curiosity to indicate that there will be one model per company created in https://github.com/odoo/odoo/blob/dbd2f89b4fe88488d3b02e475fae6ac8878694ae/addons/account/models/chart_template.py#L1085.

pedrobaeza (approved):
Let's fix the branch:

/ocabot merge patch

chienandalu (approved):

PR #730, opened 2024-10-16
[16.0][IMP] account_reconcile_oca: Open reconcile view when press the name on journal name

etobella:
Similar to Odoo EE functionality

With the change, we open directly reconcile view when pressing the open_action button

PR #763, opened 2024-12-12
[16.0][IMP] account_reconcile: ux improvements

chienandalu:
- [x] Superseds: #730 
- [x] Minor improvement on account_statement_base to add date grouping.

- [x] Depends on fix from https://github.com/OCA/account-reconcile/pull/762

fyi @etobella 

cc @Tecnativa TT52123

### DIFF
--- a/account_statement_base/views/account_bank_statement_line.xml
+++ b/account_statement_base/views/account_bank_statement_line.xml
@@ -181,6 +181,11 @@
                         string="Transaction Type"
                         context="{'group_by': 'transaction_type'}"
                     />
+                <filter
+                        name="date_groupby"
+                        string="Date"
+                        context="{'group_by': 'date'}"
+                    />
             </group>
         </search>
     </field>
--- a/account_reconcile_oca/models/account_journal.py
+++ b/account_reconcile_oca/models/account_journal.py
@@ -31,3 +31,12 @@ class AccountJournal(models.Model):
         if self.get_journal_dashboard_datas()["number_to_reconcile"] > 0:
             return False
         return _("Well done! Everything has been reconciled")
+
+    def open_action(self):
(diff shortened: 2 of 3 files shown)


pedrobaeza:
Please rebase.

chienandalu:
Done

account_statement_base/views/account_bank_statement_line.xml
```diff
@@ -181,6 +181,11 @@
                         string="Transaction Type"
                         context="{'group_by': 'transaction_type'}"
                     />
+                <filter
```
pedrobaeza, line 184:
```suggestion
                    <filter
```

chienandalu: Well this is indeed aligned with the above tags formatting... I didn't want to reformat so much to avoid the noise

pedrobaeza:
OK, the whole indentation there is mangled, but let's continue:

/ocabot merge minor

pedrobaeza (approved):
Also already approved by others in the previous PR.

ARESOLTEC-CANARIAS:
I believe the recent change—where clicking the journal button now directly redirects to the movement reconciliation instead of accessing bank statements—is a step backward for the Odoo 16 workflow.

Based on my experience and what I observe with my clients, the reconciliation process is much clearer when starting with bank statements. This workflow allows for making necessary modifications and assignments to clients before proceeding with reconciliation, offering greater control and visibility in bank management.

Forcing a direct entry into the reconciliation view not only complicates the process—since there remains a "Reconcile" button later on—but also creates confusion by duplicating functionality. I recommend reconsidering this change or at least providing a configuration option to retain the previous, more intuitive behavior.

PR #766, opened 2024-12-17
[16.0][FIX] account_reconcile_analytic_tag: Avoid error when accessing the reconciliation view

victoralmau:
Avoid error when accessing the reconciliation view

```
res[index]["manual_analytic_tag_ids"] = [(6, 0, line.analytic_tag_ids.ids)]
IndexError: list index out of range
```

Steps to reproduce the error:
- Go to Accounting Dashboard
- Click on the Reconcile X items button of the bank journal

Please @chienandalu can you review it?

@Tecnativa

### DIFF
--- a/account_reconcile_analytic_tag/models/account_reconcile_model.py
+++ b/account_reconcile_analytic_tag/models/account_reconcile_model.py
@@ -14,7 +14,8 @@ class AccountReconcileModel(models.Model):
         if len(res) == 0:
             return res
         currency = self.company_id.currency_id
-        for index, line in enumerate(self.line_ids):
+        index = 0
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Which error? It's not said in the description.

pedrobaeza (approved):
The index shouldn't increase if the balance is zero.

/ocabot merge patch

victoralmau:
Example of the error (added to the description of the PR).
ejemplo

chienandalu (approved):

pedrobaeza:
The error is not put on the commit message. Please take it into account for the next time. You always have to put the error.

PR #769, opened 2024-12-17
[16.0][FIX] account_reconcile_oca: Fixed analytic distribution update

BernatObrador:
This fixes a bug that occurred when both account_reconcile_oca and account_reconcile_model_oca modules were installed. The issue arose when assigning an analytic account and making no further changes, causing the analytic account to not be set correctly. As a result, the reconciliation process failed to create the analytic entries properly. This line: ensures that if the analytic distribution has changed, it will be correctly updated.

cc https://github.com/APSL 164854

@miquelalzanillas @lbarry-apsl @mpascuall @peluko00 @javierobcn @ppyczko please review


pedrobaeza (approved):
Thanks for backporting it. We can fast-track it:

/ocabot merge patch

PR #772, opened 2024-12-19
[16.0][FIX] account_reconcile_oca: Fixed exchange move lines never input analytic

BernatObrador:
This fix resolves an issue where analytic distribution was not generated for exchange difference lines. This issue caused conflicts with the analytic policy, as exchange accounts that required analytics would prevent reconciliation when these lines were created. The fix ensures that analytics are correctly applied.

cc https://github.com/APSL 165008

@miquelalzanillas @lbarry-apsl @mpascuall @peluko00 @javierobcn @ppyczko please review

### DIFF
--- a/account_reconcile_oca/models/account_move_line.py
+++ b/account_reconcile_oca/models/account_move_line.py
@@ -1,6 +1,8 @@
 # Copyright 2023 Dixmit
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+import json
+
 from odoo import _, models
 from odoo.exceptions import ValidationError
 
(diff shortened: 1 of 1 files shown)


mpascuall (approved):
Code review and LGTM. @etobella What do you think about it?

pedrobaeza (commented):
This bunch of code for a fix like that seems a bad patch, and more involving SQL code, so you should look for other way to solve the problem. Anyway, the multi-currency stuff seems to be solved at #694

Did you try it?

BernatObrador:
Hi @pedrobaeza, just taking a look at the PR. It solves some bugs related to multi-currency, but it doesn't fix anything regarding analytics.

This PR fixes the issue where multi-currency exchange lines could not allocate analytics. Without these changes, these lines would never pick up any type of analytics.

There are still more problems with this, for example, if you have the account_analytic_required module and the policy for exchange difference accounts is set to "always," you'll never be able to reconcile, as it will always warn you that the account is not allocating analytics.

I could review if there's a better way to handle it. The issue is that, as it is currently treated, it's necessary to make a query to access the reconcile_data from the account.bank.statement.line, which is where the analytics data for that line is stored.

pedrobaeza:
But SQL should never be used, and as said, it seems weird that such a patch requires so many lines of code.

BernatObrador:
@pedrobaeza After reviewing the issue again, I realized it can be resolved by using distribution models, so this fix is unnecessary. I'm closing the PR. Thanks for your feedback!

PR #773, opened 2024-12-27
[16.0][IMP] account_reconcile_oca: Improve _compute_reconcile_data_info. When changing account_id on journal item, data displayed in reconcilitation menu now is updated

sergiobstoj:
**Current behaviour:**

When changing `account_id` on move line, the account displayed in reconciliation menu is not getting updated.

https://github.com/user-attachments/assets/5e68dc5c-58a0-417d-ae32-e42a5f7e6848

**Behaviour with this PR:**

`account_id` displayed in reconciliation menu is updated.

https://github.com/user-attachments/assets/10dbad5e-b2f7-489a-9d85-00665f9b5ce4

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -447,9 +447,36 @@ class AccountBankStatementLine(models.Model):
             return
         self.partner_id = self.manual_partner_id
 
+    def _fields_to_check(self):
+        return ["account_id", "partner_id", "date", "name", "debit", "credit"]
+
+    def _check_recompute_reconcile_data(self):
+        self.ensure_one()
+        if self.reconcile_data:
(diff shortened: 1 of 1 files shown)


pedrobaeza:
In fact, I see that account change as something incorrect, as you are changing the liquidity account.

sergiobstoj:
I improved new method `_check_recompute_reconcile_data` to make it more generic and handle some other fields that could be changed.

About the account, I understand @pedrobaeza that you say it should not allow the account change? We are having some cases where it was necessary to change it due to mistake on counterpart account.

pedrobaeza:
If you change the bank account, it's because you have added the transaction in the wrong journal, and just changing the bank account is not enough. You have to remove the transaction in that bank, and create it in the correct bank.

I still state that this base of this PR is not correct.

PR #774, opened 2024-12-28
[17.0][FIX] Allow matching of partner in not at the begining of the string

davidwul:
re.match will yield a partner only if the search string is at the begining of the string. The match partner functionality does not specify the string needs to be at the begining. Therefore search would be more suitable.

### DIFF
--- a/account_reconcile_model_oca/models/account_reconcile_model.py
+++ b/account_reconcile_model_oca/models/account_reconcile_model.py
@@ -478,14 +478,14 @@ class AccountReconcileModel(models.Model):
 
         for partner_mapping in self.partner_mapping_line_ids:
             match_payment_ref = (
-                re.match(partner_mapping.payment_ref_regex, ".*"+st_line.payment_ref)
+                re.match(partner_mapping.payment_ref_regex, ".*" + st_line.payment_ref)
                 if partner_mapping.payment_ref_regex
                 else True
             )
             match_narration = (
                 re.match(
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
You are introducing a regex in the field, so for matching something in the middle, you have to put `.*<text>`, so it's correct as it is.

pedrobaeza:
You mustn't change this. Simply define correctly the regex on your model putting that `.*` if you want to match a word in the middle. Closing this PR.

davidwul:
I don't agree with you, 
when regexp is expected, it's mentionned:
image
but for partner mapping, nothing is mentionned:
image
we should change this for usability purpose. My proposal is to change here

pedrobaeza:
@etobella what do you think?

etobella:
I do not agree with the suggestion you introduced @davidwul you can add it there directly on the regex field :thinking:

davidwul:
@etobella 
I fully understand the developper's position. It's perfectly usable like that. But from a non developper point of view, there is no indication a regexp is needed. The labels or hints does suggest it's a search. Either adding more info for the user or changing re.match to re.search would be welcome from the lambda user perspective.

etobella:
@davidwul I understand your concerns,  but in this case I would recommend to change the label of the field. First reason is that if someone used `^whatevertext I want`, your change will break the configuration. Also, removing the `^` is not a good idea, because I might want to force it to be at the start.

PR #777, opened 2025-01-14
[16.0][FIX] account_reconcile_oca: KeyError on data

thomaspaulb:
I'm not quite sure if this is the proper solution to the below problem:

```
RPC_ERROR

  File "/home/ubuntu/odoo/auto/addons/account_reconcile_oca/models/account_bank_statement_line.py", line 456, in _compute_reconcile_data_info
    record.reconcile_data_info = record._default_reconcile_data(
  File "/home/ubuntu/odoo/auto/addons/account_reconcile_oca/models/account_bank_statement_line.py", line 570, in _default_reconcile_data
    self.reconcile_bank_line()
  File "/home/ubuntu/odoo/auto/addons/account_reconcile_oca/models/account_bank_statement_line.py", line 595, in reconcile_bank_line
    self._prepare_reconcile_line_data(self.reconcile_data_info["data"])
KeyError: 'data'
```

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -629,7 +629,7 @@ class AccountBankStatementLine(models.Model):
                     )
                     amount -= sum(line.get("amount") for line in line_data)
                     data += line_data
-                if res.get("auto_reconcile"):
+                if res.get("auto_reconcile") and self.reconcile_data_info:
                     self.reconcile_bank_line()
                 return self._recompute_suspense_line(
                     data,


thomaspaulb:
@victoralmau Looks like this got introduced with #762

victoralmau:
Please, can you indicate the steps to reproduce the error?

thomaspaulb:
@victoralmau It happens in cases where `reconcile_data_info` is (still) an empty dictionary. Then, the account bank statement line is clicked in the reconciliation widget and you get this error, because the default value is not filled yet but it is already trying to reconcile.

In Odoo Shell I've noticed that it happens already when trying to read such bank statement lines. However on second read of the same record, the error does not happen anymore because at that moment the default is already written and the `data` key has become filled.

victoralmau (approved):

pedrobaeza:
But isn't that something that indicates that the problem is that the default should be set from the beginning instead of just silencing/bypassing the error?

victoralmau:
I have not found the use case that I can functionally reproduce (important to clarify), but the fix seems to be enough to avoid the error.

thomaspaulb:
If I get time I can try to reproduce in a unit test

alexis-via (approved):
I also had this crash

PR #782, opened 2025-01-20
[16.0][FIX] account_reconcile_oca: Add context key skip_account_move_synchronization to avoid side effects

### LINKED ISSUE #779: [16][account_reconcile_oca] Foreign currency bank journal issue when currency rate changes
There is a bug with foreign currency statement lines when writing the partner on the liquidity line, explanation and how to reproduce : 

1) Install account_reconcile_oca
At this stage with a demo database, the company currency is USD and there is just an old rate on EURO currency to convert it to USD.

2) Create a bank journal with EURO currency

3) Create a statement line in this EURO bank journal with a 1000 euros amount and go to the reconcile UI.

Image

4) Go to the EURO currency and add a rate the the date of the statement line (today's date if you did not put a specific date)
Put 1.1 in the field "USD per unit"

5) Go back on the reconcile UI, click on the liquidity line and add a partner.
Then, you have an inconsistency between the accounting entries (`account.move.line`), where the debit and credit were updated with the current up to date rate, and the lines in the reconcile UI are still with the old rate.

Image

Image

This happen because the write of the partner on the bank statement line triggers a synchronisation to the accounting entries with the default vals of the related accounting entry : https://github.com/OCA/OCB/blob/16.0/addons/account/models/account_bank_statement_line.py#L824
But there are not recompute of the  statement line reconcile data.

Then, of course, if you try to reconcile some entries, the validation will fail, etc...

I am not sure what is the best way to fix this.I guess  the liquidity lines should be updated in the reconcile data, after the partner update here https://github.com/OCA/account-reconcile/blob/16.0/account_reconcile_oca/models/account_bank_statement_line.py#L507 and the also recompute the suspense line. Or maybe clean all the reconcile data but the user could have started to do some stuff already...
Also, updating other fields of the statement line (like payment ref for instance) would also cause the same issue. 

Any idea on the best way to do it ?

@etobella 
@pedrobaeza you may also be interested since I saw you merged the PR with the multi-currency fixes...

victoralmau:
Add context key skip_account_move_synchronization to avoid side effects

We need to define skip_account_move_synchronization in the context to avoid that when writing the bank statement line, some values can be updated in the account.move.line making the information inconsistent.

Related to https://github.com/OCA/account-reconcile/issues/779

Fixes #779 

@Tecnativa TT52634

### DIFF
--- a/account_reconcile_oca/views/account_bank_statement_line.xml
+++ b/account_reconcile_oca/views/account_bank_statement_line.xml
@@ -335,7 +335,7 @@
         <field name="domain">[('journal_id', '=', active_id)]</field>
         <field
             name="context"
-        >{'default_journal_id': active_id, 'search_default_not_reconciled': True, 'view_ref': 'account_reconcile_oca.bank_statement_line_form_reconcile_view'}</field>
+        >{'default_journal_id': active_id, 'search_default_not_reconciled': True, 'view_ref': 'account_reconcile_oca.bank_statement_line_form_reconcile_view', 'skip_account_move_synchronization': True}</field>
         <field name="view_mode">kanban,tree</field>
         <field
             name="view_ids"


etobella (approved):

pedrobaeza (approved):
/ocabot merge patch

florian-dacosta:
Hello,

I appreciate a fix regarding this issue. Although, this PR could have been left open a bit more time no?
I am not as confident as you seem to be about this one.

In some case, the account move synchronisation is necessary, no ? If the field "amount", or "amount_currency" is updated for instance, an update should be made right ? 

Are you sure it is fine and can you explain why please ?

victoralmau:
I am not sure if anything should be changed in any case, anyway, if you are clear about any case explain it in an issue, suggest the appropriate solution (only for that specific use case) and add a test to confirm that the use case works.

florian-dacosta:
> I am not sure if anything should be changed in any case

Sorry, I did not get what you mean?


I just did some test with this PR merged, and there is issues with what I was saying.
Example : 
1) install module with demo data
2) Create a journal in a foreign currency (like euro)
3) Create a statement line for 1000 euros in this journal
4) Now, from dashboard, click on reconcile => Switch to list view => edit the amount on the previously created line and set 2000.
5) Switch back to kanban view and you can see the amount currency is 2000 euros but the credit/debit did not change.
Also the related accounting entries did not update

Other case
1) From dashboard, click on reconcile
2) select a line : change its partner (by clicking on the liquidity line / manual operation tab)
image

3) Change the account of the suspense line and then validate
image

The resulting accounting entries are wrong, the partner is not set : 
image

PR #783, opened 2025-01-20
[16][account_reconcile_oca] Revert PR 782

florian-dacosta:
This revert https://github.com/OCA/account-reconcile/pull/782

As you can see, it seems to bring bigger issues than the one it aims to solve.

Please have a look fast @victoralmau @pedrobaeza @etobella


pedrobaeza:
OK, reverting:

/ocabot merge patch

PR #787, opened 2025-01-22
[16][FIX] account_reconcile_oca :  Wrong amount currency compute in case of change in liquidity line

florian-dacosta:
@etobella There is a regression in multi-currency management since the following commit was merged : https://github.com/OCA/account-reconcile/pull/694/commits/f0ae2f00ac2ea5ffbb1d3e19c79d2ee9c37b0635

I don't know the use case for which this code was added, but you can reproduce the issue it creates easily : 

On demo database with account_reconcile_oca (company currency is USD)
1) Create a bank journal with EURO currency
2) Create a statement line in this journal with amount = 100 euros
image

3) Click on the liquidity line and update the partner
image

=> The USD amount is copied in the amount in currency.
This is because here : https://github.com/OCA/account-reconcile/blob/16.0/account_reconcile_oca/models/account_bank_statement_line.py#L448
The source and destination currency is EURO and the amount converted is USD !
But even if this was fine, I don't see any reason to update amounts for the liquidity line... It could lead to erros because the rate at this  time may be different than the one used at the time of the creation of the statement line.

There may be a better fix, but not knowing exactly the purpose of this part, I prefer to leave it.

By the way, I am always surprised that it is possible to modifiy the manual_amount/manual_amount_currency of the liquidity line, is there any reason or could we put it as readonly to avoid user mistakes ?

I have completed a test to show the failing case.

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -442,10 +442,11 @@ class AccountBankStatementLine(models.Model):
             "debit": self.manual_amount if self.manual_amount > 0 else 0.0,
             "analytic_distribution": self.analytic_distribution,
         }
-        if self.manual_line_id:
+        liquidity_lines, _suspense_lines, _other_lines = self._seek_for_lines()
+        if self.manual_line_id and self.manual_line_id.id not in liquidity_lines.ids:
             vals.update(
                 {
-                    "currency_amount": self.manual_line_id.currency_id._convert(
+                    "currency_amount": self.manual_currency_id._convert(
                         self.manual_amount,
                         self.manual_in_currency_id,
                         self.company_id,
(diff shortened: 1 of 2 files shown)


victoralmau (commented):
In runboat solves the explained error BUT.

When updating (onchange) the `manual_amount_in_currency` field, the `manual_amount` field is not updated (the same in reverse).

I think @etobella will be able to provide more context to all this.

florian-dacosta:
> When updating (onchange) the `manual_amount_in_currency` field, the `manual_amount` field is not updated (the same in reverse).

This is the case with and without the PR I think. I am well aware of such problems, but this is not a problem I am trying to solve.
I believe there still are a lot of issues linked to multi-currency if the user start to edit some amount or amount currency manually. 
I think this part require quite some refactore, and also to have detailed use-cases we want to solve for a start.
For now we can edit both amount and amount in currency, and modyfing one sometimes impact the other, or not....

In anyway, I think we can deal with these issues in another PR, if someone have the courage to dig into this!

victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
Please forward-port it to 17

PR #792, opened 2025-01-31
[16.0][FIX] account_reconcile_oca: Fix currency and amount when using reconciliation models

sergiobstoj:
When using manual reconcile model to reconcile, the amount in currency is not properly updated.

#### Steps to reproduce in runbot:
1. Create Journal with foreign currency (for example EUR) and bank type.
2. Create Reconciliation Model as writeoff_button type with some account in counterpart entries.
3. Go to Dashboard -> Get into journal created in step 1 -> Create.
4. Set any amount and any label and Save it.
5. Click on your statement and click on reconciliation model button.
6. Amount in currency has the wrong currency and wrong amount.

https://github.com/user-attachments/assets/1c404d26-6b8b-42dd-b45b-d22108e0152c

#### Behaviour with PR:
Reproducing the same steps, currency and amount are now correctly displayed.

https://github.com/user-attachments/assets/81f8d2a2-5e86-4502-91f4-4b8c9d53338f

PR #793, opened 2025-02-04
[17.0][FIX] account_reconcile_oca: Fix currency and amount when using reconciliation models

sergiobstoj:
Forward-port of #792

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -585,6 +585,13 @@ class AccountBankStatementLine(models.Model):
                     self.company_id,
                     self.date,
                 )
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
@sergiobstoj please solve the existing conflicts, and only 18.0 is pending then.

sergiobstoj:
@pedrobaeza conflicts solved!

pedrobaeza:
I see that the same code you have added is already present. Can you check if anything more is needed?

sergiobstoj:
@pedrobaeza I checked that 17.0 already contains all that I have added in 16.0. I think we can close this one

PR #794, opened 2025-02-04
[16.0][IMP] account_reconcile_oca: improve visibility of partner on bank statement line

alexis-via:


alexis-via:
BEFORE

before_rec_oca

AFTER

after_rec

florian-dacosta (approved):
It seems like a good idea to me

pedrobaeza:
It should be checked what happens with large partner names. If they are stripped, this change is not valid.

alexis-via:
No pb with long names:

long_name

pedrobaeza (approved):
Please fw-port it to 17

/ocabot merge patch

etobella (approved):

PR #798, opened 2025-02-11
[16.0][IMP] account_reconcile_oca: Allow to select all lines at once

etobella:
The idea is to add a button to add all lnes at once

image


pedrobaeza (approved):
Code and test in runboat. This alleviates a lot the times when working with payment orders, only requiring to pre-filter the wanted entries.

Let's fast-track it for forwarding port it, and to try it in real scenarios, as the risk is low.

/ocabot merge patch

PR #810, opened 2025-02-28
[ADD] documentation : reconciliation model after migration

victor-champonnois:
I've noted that migrated databases (in my case from v12 to v16) do not have reconciliation models set by default. Therefore the account_reconcile_oca module does not work well. We need to manually configure the reconciliation models. 
Maybe there is a way to solve this trough code, but I don't know how, so in the meantime I propose a documentation. 
Also maybe we're doing something wrong in our migration process, if so, we will be happy to fix it!

@vdewulf


flotho (approved):
easy and usefull

etobella (approved):

PR #814, opened 2025-03-20
[18.0][MIG] Account_move_line_reconcile_manual: Migration to 18.0

TB-Ph35:


florian-dacosta:
Waiting to decide on https://github.com/akretion/account-reconcile/pull/5
Port of https://github.com/OCA/account-reconcile/pull/934

account_move_line_reconcile_manual/models/account_reconcile_manual_model.py
```diff
@@ -0,0 +1,99 @@
...
+                }
+            )
+            income_distri = aadmo._get_distribution(
+                {
+                    "account_prefix": model.income_account_id.code,
+                    "company_id": model.company_id.id,
+                }
+            )
+            model.expense_analytic_distribution = expense_distri
+            model.income_analytic_distribution = income_distri
```
Reyes4711-S73, line 99:
```suggestion
            model.update(
                {
                    "expense_analytic_distribution": expense_distri,
                    "income_analytic_distribution": income_distri,
                }
            )
```

jbaudoux (commented):
@florian-dacosta If I have an invoice in foreign currency and a payment in company currency, I can't reconcile with write-off. The created write-off line makes the residual amount to 0 but the invoice currency residual amount is not 0 and the currency line is not considered as reconciled.

account_move_line_reconcile_manual/models/account_reconcile_manual_model.py
```diff
@@ -0,0 +1,99 @@
...
+        res = super().default_get(fields_list)
+        res["company_id"] = self.env.company.id
+        journals = self.env["account.journal"].search(
+            [("company_id", "=", res["company_id"]), ("type", "=", "general")]
+        )
+        if len(journals) == 1:
+            res["journal_id"] = journals.id
+        return res
+
+    @api.depends("expense_account_id", "income_account_id")
```
BhaveshHeliconia, line 82:
```suggestion
    @api.depends("expense_account_id", "income_account_id")
    def _compute_analytic_distribution(self):
        aadmo = self.env["account.analytic.distribution.model"]
        for model in self:
            expense_code = model.expense_account_id.code
            income_code = model.income_account_id.code

            model.expense_analytic_distribution = aadmo._get_distribution({
                "account_prefix": expense_code,
                "company_id": model.company_id.id,
            }) if expense_code else {}

            model.income_analytic_distribution = aadmo._get_distribution({
                "account_prefix": income_code,
                "company_id": model.company_id.id,
            }) if income_code else {}
```

BhaveshHeliconia (changes requested):

florian-dacosta:
> @florian-dacosta If I have an invoice in foreign currency and a payment in company currency, I can't reconcile with write-off. The created write-off line makes the residual amount to 0 but the invoice currency residual amount is not 0 and the currency line is not considered as reconciled.

@jbaudoux 
I may try to check this case, but these multi-currency reconciliation cases are not always well managed by Odoo (at least it was not at all in v16). 
We manage well the "same currency" case (company currency or foreign), but multi company currency cases are not necessarily working indeed. I'd need to take the time to check if Odoo improved it in this version allowing us to have something more reliable about these cases.

florian-dacosta:
@jbaudoux Like I said, before (v16) multiple currency reconciliation were quite buggy, not really supported by the module (see https://github.com/OCA/account-reconcile/pull/558)

But I just added a simple multi-currency test (that was not working in v16) and it is passing now.
Also I did 2/3 manual simple test with invoice in foreign currency, payment in company currency + write-off, and it did work.
Could you describe the steps to reproduce your use case exactly ?

jbaudoux:
> @jbaudoux Like I said, before (v16) multiple currency reconciliation were quite buggy, not really supported by the module (see #558)
> 
> But I just added a simple multi-currency test (that was not working in v16) and it is passing now. Also I did 2/3 manual simple test with invoice in foreign currency, payment in company currency + write-off, and it did work. Could you describe the steps to reproduce your use case exactly ?

Thanks for looking into this. You could complete your test by appending some extra checks:
Line1 residual amount is 0
Line1 currency residual amount is 0
Line1 reconciled is True

In your manual test, have you changed the currency rate before reconciliation so that the currency conversion of the write off doesn't match the currency residual?

florian-dacosta:
@jbaudoux I did refactore the tests to be able to add different cases more easily.
And adding some differente rates did show issues on reconciliation between foreign currency and company currency indeed.

Like I said, these cases are quite complicated, but in v18, I think there is a way to get around the issue.
I did a refactore here : https://github.com/akretion/account-reconcile/pull/7
I am interested to know if you still have issues with this one.

PR #815, opened 2025-03-27
[16.0][ADD] account_reconcile_manual_tax

hbrunn:
This module allows to set taxes (price included) on manual
reconciliation lines.

Usage
=====

To use this module, you need to:

1. Enter the reconciliation screen
2. Switch to manual operation
3. Set the field ``Taxes``. Note only price included taxes can be
   selected.


Stephanvandenberg (approved):
LGTM - will improve handeling bank transactions w. taxes (such as Mollie)

Stephanvandenberg:
Additional testing done. Works well for the "taxes included" tax codes. Both on sales and purchasing side. Use case could for example be any bank transaction for which you are allowed to ask back taxes.

PR #817, opened 2025-03-31
[18.0][MIG] account_mass_reconcile

amkarthik:
cc: @SodexisTeam


StefanRijnhart (changes requested):
It seems to me this module relied on the fact that entries with the same account always belonged to the same company. That will not always be the case in Odoo 18. The matching mechanism should be slightly refactored to ensure that entries from different companies are not matched.

PR #825, opened 2025-04-11
[FIX] account_reconcile_oca: Apply matching of partners on reconciliation models

etobella:
We forgot to get this information. :cry:

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -572,11 +572,15 @@ class AccountBankStatementLine(models.Model):
             liquidity_amount += line_data["amount"]
             if line_data["kind"] == "liquidity":
                 default_name = line_data["name"]
+        partner = (
+            reconcile_model._get_partner_from_mapping(self) or self._retrieve_partner()
+        )
         for line in reconcile_model._get_write_off_move_lines_dict(
-            -liquidity_amount, self._retrieve_partner().id
+            -liquidity_amount, partner.id
         ):
             new_line = line.copy()
             new_line["name"] = new_line.get("name") or default_name
+            new_line["partner_id"] = partner and partner.name_get()[0] or False
             amount = line.get("balance")
             if self.foreign_currency_id:
                 amount = self.foreign_currency_id.compute(


account_reconcile_oca/models/account_bank_statement_line.py
```diff
@@ -572,11 +572,15 @@ def _reconcile_data_by_model(self, data, reconcile_model, reconcile_auxiliary_id
...
+        partner = (
+            reconcile_model._get_partner_from_mapping(self) or self._retrieve_partner()
+        )
         for line in reconcile_model._get_write_off_move_lines_dict(
-            -liquidity_amount, self._retrieve_partner().id
+            -liquidity_amount, partner.id
         ):
             new_line = line.copy()
             new_line["name"] = new_line.get("name") or default_name
+            new_line["partner_id"] = partner and partner.name_get()[0] or False
```
pedrobaeza, line 583: `partner_id` contains a string? Isn't the key name misleading?

etobella: This vals are reused on the widget, not for the creation. it is transformed later. it is true that partner_id might not be the best name, but this is what we have right now.

pedrobaeza: You can add a previous refactoring commit renaming all its uses for not continuing spreading an incorrect name.

etobella:
I added a migration script in order to make remove the data in order to ensure that nothing wrong happens
d

account_reconcile_oca/__manifest__.py
```diff
@@ -5,7 +5,7 @@
     "name": "Account Reconcile Oca",
     "summary": """
         Reconcile addons for Odoo CE accounting""",
-    "version": "16.0.2.3.7",
+    "version": "16.0.3.0.0",
```
pedrobaeza: Why do you need this version bump and migration script?

pedrobaeza:
Ouch, I didn't anticipate that problem... Maybe it has been a bad idea to ask for this refactoring. With this, we are losing any pending operation. That's not OK. Can we just touch existing data to rename it or at least remove the offending field? Or to not crash when finding an incorrect field? Other doubt is that you are now placing a recordset, instead of an ID or a text. Is that OK when serializing?

etobella:
Well, we always passed the name_get because we wanted to avoid the orm connection from the widget (it is handled by the model itself) I do prefer to keep it this way for safety reasons.

We have 3 options:
1. Keep the old name
2. Change the name and we try to create a migration script that is smarter (and slower)
3. Change the name and keep this migration script.

If we try to make something to make it work with both names in order to avoid the migration script is uglier IMO. So, it is not an option for me.

Which option do you prefer?

pedrobaeza:
So sorry for the noise, but seeing the consequences, please undo and keep 1.

etobella:
Don't worry :smile:

pedrobaeza (approved):
/ocabot merge patch

PR #833, opened 2025-04-19
[16.0][FIX] account_reconcile_oca: When doing manual reconciliations, allowto set the amount to reconcile in the invoice currency.

JordiBForgeFlow:
Before this fix, when attempting to reconcile a statement line in EUR with an invoice in USD, upon indicating the amount to reconcile in the currency invoice the results would be incorrect. 

This fix also includes an improvement on the message shown in the bank statement line on the effect of the reconciliation to the invoice.

cc @etobella @pedrobaeza

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -10,6 +10,7 @@ from odoo import Command, _, api, fields, models
 from odoo.exceptions import UserError
 from odoo.fields import first
 from odoo.tools import float_compare, float_is_zero
+from odoo.tools.misc import formatLang
 
 
 class AccountBankStatementLine(models.Model):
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Can you please indicate all the needed data and steps to reproduce the problem to be able to check?

JordiBForgeFlow:
@pedrobaeza I need to dig in some more how to approach the fix. 

The core issue has to do with the currency rates and the fields 'manual_amount_in_currency' and 'manual_amount'. If you indicate the manual_amount_in_currency, it updates the manual_amount and this one then changes again the manual_amount_in_currency in turn, generating small variations during the conversions that cause that in the end, the invoice cannot be fully reconciled with the payment.

JordiBForgeFlow:
I have enconded only the tests entered here that fail, prior to the fixes, in https://github.com/OCA/account-reconcile/pull/834

victoralmau (approved):
Code review OK, although perhaps @etobella (I think he knows more about it) can give her opinion on whether this is a good approach or if there is another more suitable solution.

PR #836, opened 2025-04-24
[IMP] account_statement_base: Added internal_index in tree view for sorting

feg-adhoc:
The internal_index field is added to correctly sort the statement lines, because if we sort by date or sequence, the running_balance field will be incorrect.

### DIFF
--- a/account_statement_base/views/account_bank_statement_line.xml
+++ b/account_statement_base/views/account_bank_statement_line.xml
@@ -59,6 +59,7 @@
         <field name="model">account.bank.statement.line</field>
         <field name="arch" type="xml">
             <tree editable="top" multi_edit="1" decoration-muted="is_reconciled">
+                <field name="internal_index" optional="hide" />
                 <field name="sequence" />
                 <field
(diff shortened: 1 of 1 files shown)


account_statement_base/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 
 {
     "name": "Bank Statement Base",
-    "version": "16.0.1.14.0",
+    "version": "16.0.1.15.0",
```
pedrobaeza: Please don't change manually the module version. It's done by the bot on merge.

pedrobaeza (changes requested):
I don't think adding a column to press on it to sort is the best option. If the default order is not correct, then switch that default order to internal_index, but I don't have clear why the order is not correct.

feg-adhoc:
@pedrobaeza Thanks for the review!

pedrobaeza:
The rest of the comments continue applying the same. And this is not `[ADD]`. In any case, `[IMP]`.

Issue #813, opened 2025-03-17
[17.0, 18.0]account_reconcile_oca: Can't fetch record error to reconcile


victoralmau:
Affected versions: 17.0 y 18.0

Steps to reproduce the error:
- Create a new customer.
- Create an invoice linked to the new customer.
- Create a payment linked to the new customer.
- Go to the Reconciliation menu (Invoicing > Accounting > Reconcile).
- Select the new customer in the left column.
- Select the 2 entries in the right column and click on “Reconcile”.
- An error is displayed because the account.account.reconcile record has been deleted.

![Image](https://github.com/user-attachments/assets/1d9ab4ef-8997-4f71-859a-75ab68a00715)

@Tecnativa TT55470

victoralmau:
Ping @etobella

amar989ms:
reconcile, clean, add all buttons not work. Error: on version 18
UncaughtPromiseError > TypeError
Uncaught Promise > Cannot read properties of undefined (reading 'querySelectorAll')

Occured on localhost:8018 on 2025-04-09 11:03:36 GMT

TypeError: Cannot read properties of undefined (reading 'querySelectorAll')
    at disableButtons (http://localhost:8018/web/assets/52c6210/web.assets_web.min.js:9631:624)
    at executeButtonCallback (http://localhost:8018/web/assets/52c6210/web.assets_web.min.js:9633:1)
    at Object.onClickViewButton (http://localhost:8018/web/assets/52c6210/web.assets_web.min.js:9643:487)
    at ViewButton.onClick (http://localhost:8018/web/assets/52c6210/web.assets_web.min.js:9622:10)
    at Object.mainEventHandler (http://localhost:8018/web/assets/52c6210/web.assets_web.min.js:1616:77)
    at HTMLButtonElement.listener (http://localhost:8018/web/assets/52c6210/web.assets_web.min.js:771:15)

amar989ms:
Add all button error solution:
export class ReconcileMoveLineController extends ListController {
    setup() {
        super.setup();
        this.orm = useService("orm");
    }

setup() function added then it worked. But Reconcile, Clean btn still not work

hussain:
I'm not that well in OWL, so I used Augment Code to fix it!

# Issue Summary and Resolution for account_reconcile_oca Module

## Issue Description
When using the account_reconcile_oca module in Odoo 18.0, clicking the "Reconcile" button successfully reconciled the entries but produced a JavaScript error in the console:

```
[...]

PR #840, opened 2025-05-08
[18.0][FIX] account_reconcile_oca: Can't fetch record error to reconcile

hussain:
Before this fix using the account_reconcile_oca module in Odoo 18.0, opening the widget cause a loading module warning.

In the reconcile_data_widget.esm.js file where the widget wasn't being loaded correctly due to improper props definition.
This fix properly defines the widget props using standardFieldProps and moves the template definition to the class declaration,
ensuring the widget is loaded correctly and preventing the FetchRecordError after reconciliation.

This seems to be partially related #813 in a way it does not toast error message of deleted IDs not found in all cases, which I'm not sure why.

I signed OCA CLA.

### DIFF
--- a/account_reconcile_oca/static/src/js/widgets/reconcile_data_widget.esm.js
+++ b/account_reconcile_oca/static/src/js/widgets/reconcile_data_widget.esm.js
@@ -3,10 +3,15 @@ import {formatDate, parseDate} from "@web/core/l10n/dates";
 import {formatMonetary} from "@web/views/fields/formatters";
 import {registry} from "@web/core/registry";
 import {useService} from "@web/core/utils/hooks";
+import {standardFieldProps} from "@web/views/fields/standard_field_props";
 
 const {Component} = owl;
 
 export class AccountReconcileDataWidget extends Component {
+    static props = {
+        ...standardFieldProps,
+    };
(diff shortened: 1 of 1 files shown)


victoralmau (changes requested):
Reviewing the error described at https://github.com/OCA/account-reconcile/issues/813 the error still occurs.

ejemplo

hussain:
I test it in runboat and actually it does work without error!


https://github.com/user-attachments/assets/717d83eb-87ef-49f3-9b93-222952d15871




but there is another issue: after reconcile/clear it does not allow selection to move to upper panel. Checking the issue.

victoralmau:
The error still occurs.

ejemplo

Important to select in the left panel the Test customer

ejemplo

hussain:
This is getting interesting!
There was initially a module loading issue which was fixed along same code change,I over sighted this and did not test before changing the other part!
Tested in runboat successfully for loading the widget, and for reconcile it seems less errors but there are cases  where it did toast the deleted IDs message but I cannot get the case identified and I'm not sure if it actually relate to this PR conceptually.

etobella:
I will invest today sometime in order to check why it happens.

Issue #830, opened 2025-04-13
After a successful reconciliation, a message is shown stating that "IDs cannot be found"


agrogers:
If i refresh the page, the reconciled transactions are no longer visible. Maybe what is happening is at the end of the reconcile process the form tries to access a record that has been removed via the successful reconcile.

![Image](https://github.com/user-attachments/assets/06408612-b7a5-4aff-9125-19bfb2c6ec9c)

PR #841, opened 2025-05-12
[18.0][FIX] account_reconcile_oca

etobella:
Includes #840 

It applies 2 fixes:
- When we press reset, the selector of the lines was not working anymore
- #830 

@pedrobaeza @hussain @victoralmau

### DIFF
--- a/account_reconcile_oca/models/account_account_reconcile.py
+++ b/account_reconcile_oca/models/account_account_reconcile.py
@@ -21,6 +21,7 @@ class AccountAccountReconcile(models.Model):
     account_id = fields.Many2one("account.account", readonly=True)
     name = fields.Char(readonly=True)
     is_reconciled = fields.Boolean(readonly=True)
+    active = fields.Boolean(default=True)
 
     @property
     def _table_query(self):
@@ -54,7 +55,19 @@ class AccountAccountReconcile(models.Model):
                 FALSE as is_reconciled,
                 aml.currency_id as currency_id,
                 am.company_id,
-                null as foreign_currency_id
+                null as foreign_currency_id,
+                (
+                    SUM(
+                        CASE WHEN aml.amount_residual > 0
+                        THEN aml.amount_residual
+                        ELSE 0 END
+                    ) > 0
+                    AND SUM(
+                        CASE WHEN aml.amount_residual < 0
+                        THEN -aml.amount_residual
+                        ELSE 0 END
+                    ) > 0
+                ) as active
         """
 
     def _from(self):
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):

victoralmau (commented):
The use case explained at https://github.com/OCA/account-reconcile/issues/813#issue-2924255383 is still happening.

ejemplo2

hussain:
> The use case explained at [#813 (comment)](https://github.com/OCA/account-reconcile/issues/813#issue-2924255383) is still happening.

Can you try in asset mode as well? [here](http://oca-account-reconcile-18-0-pr841-98967d829168.runboat.odoo-community.org/odoo/action-302?id=1&debug=assets)
js is really buggy in browsers, I had to regenerate bundle and sometimes even clear all cookies/cache to get what actually is running in code.

This might help as well + asset mode
image

I tested same and it is ok!

reconcile - ok

hussain (approved):

etobella:
Mmmh, it is true that it does not happen in debug assets mode, I will dig on it, but we are clearly closer than before :wink:

pedrobaeza:
If not happening in debug=assets mode, I will say it's something of dependency imports.

victoralmau:
Clarify that debug=assets does not work either.

ejemplo

ejemplo

ejemplo-1
ejemplo-2
ejemplo-3
ejemplo-4
ejemplo-5

etobella:
I found the real problem. The error seems to apear if we have more moves to do.

I need to find why this happens :disappointed:

etobella:
evenmore, only if we select the first item :dizzy_face:

victoralmau:
For your convenience, I have an idea to solve it (although I do not know if it is the most appropriate) because I think it is an error regarding core because the record (`account.account.reconcile`) no longer exists. Create an `active` field (compute) in the `account.account.reconcile` model and delete the existing where (that logic will be transferred to the active field); this way, the error will not occur and everything will continue to work the same. 
What do you think?

etobella:
I think that the problem is related to the account.account.reconcile model, and that is a view. As we removed the id, it tries to find it again and everything fails.

etobella:
@victoralmau I loved your solution. it worked!!

etobella:
BTW, I added you as co-author of the commit, the work wouldn't be possible with your help. Thanks!!!

victoralmau:
Thanks, it works properly.

It would be great if the left side panel would refresh (because if there is nothing pending anymore, it won't be necessary to show it), although it is true that it refreshes when clicking on another item.

ejemplo

victoralmau:
Can you apply this change also v17?

etobella:
And probably on 16, this issues are relevant on all versions :wink:

PR #843, opened 2025-05-15
[17.0][FIX] account_reconcile_oca: display foreign currency symbol in reconciliation widget

MaxyMoos:
When available, the "_Amount in currency_" column displayed in the reconciliation widget only displays the amount, without the currency symbol.

Seems like the call made to `formatMonetary` which should format this amount is currently incorrectly passed 3 args (probably because the method in previous Odoo versions used to be called `formatCurrency` and indeed used to have 3 arguments)

**Currently:**
before

**With this fix:**
after


etobella (approved):

pedrobaeza:
@MaxyMoos can you please fw-port it to 18?

PR #844, opened 2025-05-15
[18.0][IMP] account_reconcile_oca: refresh screen, keep the url id

norlinhenrik:
Menu: Accounting - Accounting - Reconcile

When I click on open items, they appear on the top, and I click on Reconcile. 
**Then**, when I click on open items, **nothing happens**.
When I refresh the screen, I always come to the **first** section to reconcile.

**Existing workaround**: On the left side, click on another section, then click on the section I want to work on.

**Preferred solution**: When I click on an open item, it should always appear on the top and be ready to Reconcile.

**Solution of this PR**: Refresh the screen, and keep the active section (keep the ID in the url).


norlinhenrik:
Now I don't have the initial problem anymore. When I click on open items, they always appear on the top. So it seems to me that this PR is not needed, but still useful.

PR #845, opened 2025-05-16
[18.0][IMP] account_statement_base: Add kanban view to bank statement line action

rov-adhoc:
This change adds the kanban view mode to the 'account_bank_statement_line_action' to allow users to visualize bank statement lines in a kanban format, in addition to list and form views.

### DIFF
--- a/account_statement_base/views/account_bank_statement_line.xml
+++ b/account_statement_base/views/account_bank_statement_line.xml
@@ -180,7 +180,7 @@
     <record id="account_bank_statement_line_action" model="ir.actions.act_window">
         <field name="name">Bank Statement Lines</field>
         <field name="res_model">account.bank.statement.line</field>
-        <field name="view_mode">list,form</field>
+        <field name="view_mode">list,form,kanban</field>
         <field name="context">{'account_bank_statement_line_main_view': True}</field>
     </record>
 </odoo>


etobella (approved):

pedrobaeza:
I don't see how can you use such view. Can you tell me the steps to go to that view?

rov-adhoc:
@pedrobaeza Thanks for your question. This kanban view is not accessed directly via a user-visible action. However, it is required when the module account_accountant is installed.

In that case, the list view view_bank_statement_line_tree_bank_rec_widget is used to display account.bank.statement.line unreconciled  records, especially within the bank reconciliation widget (Match).

From this view, depending on the UI context (screen size, user preferences, or internal JS logic), Odoo may try to switch to a kanban view. If none exists, it raises a ViewNotFoundError, like this:

ViewNotFoundError: No se encontró ninguna vista de tipo "kanban" en la acción actual.

Adding this minimal kanban view prevents the frontend crash and ensures smoother integration with the reconciliation widget.

Let me know if you'd like me to share a screen recording of the error.
image

pedrobaeza:
But shouldn't that something to be fixed in `account_accountant` enterprise module itself?

rov-adhoc:
@pedrobaeza 
The Enterprise module already (account_accountant) defines an action including a kanban view.

However, the issue comes from the fact that the OCA module (account_stament_base) creating her own action "account_bank_statement_line_action"

So when both are installed, and the OCA module of importation statement, the kanban view is missing in the OCA action— which leads to a ViewNotFoundError in some reconciliation flows (e.g. from the list view with the bank_rec_widget).

This PR doesn't introduce new functionality, but simply restores compatibility with Enterprise by adding a minimal kanban view, so that Odoo doesn’t crash when it tries to switch to it.

Regards

pedrobaeza (approved):
OK, thanks for the explanation. Let's merge it for the compatibility shake. 

/ocabot merge patch

PR #842, opened 2025-05-13
[17.0][FIX] account_reconcile_oca: Show the name of the move if reconciliation model apply

victoralmau:
Show the name of the move if reconciliation model apply

Consistent with existing: https://github.com/OCA/account-reconcile/blob/bb97b2a2d122c4f99b98e93724b5b2c2647e786d/account_reconcile_oca/models/account_bank_statement_line.py#L983

Example use case:
- Invoice reconciliation model without self-validation.
- Create an invoice to a customer with an amount
- Create a bank statement line with the corresponding amount
- Click on Reset reconciliation
- The name of the invoice against which it has been linked will be displayed

**Before**
antes

**After**
despues

Please @pedrobaeza and @etobella can you review it?

@Tecnativa TT56294 TT56157

PR #847, opened 2025-05-21
[18.0][FIX] account_reconcile_oca: Show the name of the move if reconciliation model apply

victoralmau:
FWP from 17.0: https://github.com/OCA/account-reconcile/pull/842

Show the name of the move if reconciliation model apply

Consistent with existing: https://github.com/OCA/account-reconcile/blob/dcfebe9feb7896e569d9f2a1371566e803616ea9/account_reconcile_oca/models/account_bank_statement_line.py#L987

Example use case:
- Invoice reconciliation model without self-validation.
- Create an invoice to a customer with an amount
- Create a bank statement line with the corresponding amount
- Click on Reset reconciliation
- The name of the invoice against which it has been linked will be displayed

**Before**
antes

**After**
despues

Please @pedrobaeza and @etobella can you review it?

@Tecnativa TT56294 TT56157

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -670,6 +670,7 @@ class AccountBankStatementLine(models.Model):
                         is_counterpart=True,
                         max_amount=amount,
                         reconcile_auxiliary_id=reconcile_auxiliary_id,
+                        move=True,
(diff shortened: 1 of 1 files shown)


etobella (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #850, opened 2025-05-29
[17.0][IMP] account_reconcile_oca: Add multi-company rule to account.account.reconcile

victoralmau:
Add multi-company rule to account.account.reconcile

In a multi-company context, when working with only one company and accessing the Reconcile menu, we want to see only the data of the selected companies, not all of them.

Please @pedrobaeza and  @etobella can you review it?

@Tecnativa TT56510

### DIFF
--- a/account_reconcile_oca/security/security.xml
+++ b/account_reconcile_oca/security/security.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <record id="rule_account_account_reconcile_multi_company" model="ir.rule">
+        <field name="name">account.account.reconcile multi-company</field>
+        <field name="model_id" ref="model_account_account_reconcile" />
+        <field name="domain_force">[('company_id','in',company_ids)]</field>
+    </record>
+</odoo>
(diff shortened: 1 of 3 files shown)


account_reconcile_oca/security/security.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <record id="rule_account_account_reconcile_multi_company" model="ir.rule">
+        <field name="name">account.account.reconcile multi-company</field>
+        <field name="model_id" ref="model_account_account_reconcile" />
+        <field name="global" eval="True" />
```
pedrobaeza: AFAIK, this is a computed one.

pedrobaeza (commented):
Shouldn't this be in `account_reconcile_model_oca`?

pedrobaeza:
Oh, I think I read incorrectly... I said reconcile model, but it was `model_account_reconcile_reconcile`, so I think my recommendation about moving to the other module was incorrect.

pedrobaeza (approved):

etobella (approved):

PR #853, opened 2025-05-30
[16.0][IMP] account_reconcile_oca: Add multi-company rule to account.account.reconcile

victoralmau:
Backport from 17.0: https://github.com/OCA/account-reconcile/pull/850

Add multi-company rule to account.account.reconcile

In a multi-company context, when working with only one company and accessing the Reconcile menu, we want to see only the data of the selected companies, not all of them.

Please @pedrobaeza and  @etobella can you review it?

@Tecnativa TT56510


pedrobaeza (approved):
Finishing the chain.

/ocabot merge patch

PR #856, opened 2025-06-09
[17.0][FIX] account_reconcile_oca: Define the appropriate name in the suggested counterparts of transactions (similar to v15)

victoralmau:
Define the appropriate name in the suggested counterparts of transactions (similar to v15)

Example use case:
- Create a vendor invoice + validate
- Create a bank statement line
- Add as a suggested reconciliation of the statement line the vendor invoice
- The label (name) to be displayed will be the name of the journal entry
- Make the reconciliation
- The journal entry of the generated entry will have the appropriate label

ejemplo

@Tecnativa TT56592

### DIFF
--- a/account_reconcile_oca/models/account_reconcile_abstract.py
+++ b/account_reconcile_oca/models/account_reconcile_abstract.py
@@ -95,7 +95,7 @@ class AccountReconcileAbstract(models.AbstractModel):
             if line.partner_id
             else False,
             "date": fields.Date.to_string(line.date),
-            "name": line.name,
+            "name": line.name or line.move_id.name,
             "debit": amount if amount > 0 else 0.0,
             "credit": -amount if amount < 0 else 0.0,
             "amount": amount,


etobella (approved):

account_reconcile_oca/models/account_reconcile_abstract.py
```diff
@@ -95,7 +95,7 @@ def _get_reconcile_line(
             if line.partner_id
             else False,
             "date": fields.Date.to_string(line.date),
-            "name": line.name,
+            "name": line.name or line.move_id.name,
```
pedrobaeza, line 98: But `move_id.name` is the journal entry number, not the ref, so I don't think it's the correct solution.

victoralmau: I think the best is an example of v15 and the reconciliation process. Even if the invoice name + reference is shown, only the invoice name will be set in the journal entry.

ejemplo-v15

pedrobaeza: But what we need is the label of the statement line, not that name.

victoralmau: IMO no, what is intended is to be able to identify in the label of the account.move.line that the “origin” has been generated, and the way to do it is this, the same as v15.

Defining there the label of the bank statement line I think would be a mistake and would generate more confusion than help, why do we want to indicate the same thing in all the account.move.line? If the label of the bank statement line is REF0001 it will be defined in the corresponding account.move.line, but if it is linked to for example 4 invoices, personally it would not help me to see in those 4 account.move.line REF0001, but it would be helpful to see the name of the corresponding invoices: INVOICE001, INVOICE002, etc.

In any case, which is then the right behavior?

pedrobaeza: Please check how it's done on 15- and on enterprise to compare. I'm not sure this should have the same label as the counterpart.

victoralmau: The above video is the behavior of **v15**. The behavior of enterprise in v16 is as follows (it does not define anything in the account.move.line).

ejemplo

victoralmau: v18 enterprise example.

ejemplo

pedrobaeza: The GIF is a bit difficult to follow, but If I'm not seeing wrong, it basically don't put any label on the generated counterpart. Is that? If so, I think it's correct as it is now without patches, or am I wrong?

victoralmau: In v16 enterprise NO label is defined, but in v18 it is, namely the name of the [counterpart](https://82068850-18-0-all.runbot225.odoo.com/odoo/accounting/6/reconciliation/account.move/393).

ejemplo

Anyway you can try it yourself to confirm it.

pedrobaeza: OK, seen online and we will put it this way.

pedrobaeza (approved):
/ocabot merge patch

PR #738, opened 2024-11-07
[18.0][MIG] account_reconcile_model_oca: Migration to 18.0

xaviedoanhduy:
### change in v18

- `get_sql` has removed in https://github.com/odoo/odoo/commit/5c397d56eb33dc37a0b0a8d1e7d4c43583c95950
- `account_journal_payment_credit_account_id` has removed in https://github.com/odoo/odoo/commit/01b87f1230beac0568f4e3b1b76e547909506892

### depend on

- https://github.com/OCA/account-reconcile/pull/740

PR #860, opened 2025-06-12
[18.0][FIX] account_reconcile_model_oca: partner matching - fix migration v18.0

jbaudoux:
Fix lost parenthesis in migration to v18.0 causing matching any aml when one partner is matched

Use case: I have some partners having `communication` in their name. When reconciling a bank statement line from a customer payment, the bank put as narration `Communication: xyz`. This causes those partners to match. As the query `OR` is applied on the `JOIN` between `res_partner.id` and `aml.partner_id`, this join is not respected anymore causing to return the first aml partner. 

@victoralmau @pedrobaeza @mmequignon @JordiBForgeFlow

### DIFF
--- a/account_reconcile_model_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_model_oca/models/account_bank_statement_line.py
@@ -88,14 +88,14 @@ class AccountBankStatementLine(models.Model):
                     aml.partner_id = partner.id
                     AND partner.name IS NOT NULL
                     AND partner.active
-                    AND (
+                    AND ((
             """)
             query_parts = SQL(") OR (").join(sub_queries)
             final_query = SQL(
                 """
                 %s
                     %s
-                )
+                ))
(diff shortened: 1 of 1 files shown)


pedrobaeza:
It's difficult to see the scope of the parenthesis and its effect reviewing the diff. Maybe you have a screenshot where you analyze such lost diff and can be seen easily?

jbaudoux:
> It's difficult to see the scope of the parenthesis and its effect reviewing the diff. Maybe you have a screenshot where you analyze such lost diff and can be seen easily?

@pedrobaeza see here https://github.com/OCA/account-reconcile/pull/738/commits/7d3576c02eb06a94dd8049d1dd4803ef42e680c3

pedrobaeza (approved):
OK, spotted!

imagen

I think it's clear and can be fast-tracked.

/ocabot merge patch

PR #861, opened 2025-06-16
[17.0][IMP] account_reconcile_model_oca: Add Unique match option to reconciliation models

victoralmau:
Add Unique match option to reconciliation models

@Tecnativa TT56737

### DIFF
--- a/account_reconcile_model_oca/README.rst
+++ b/account_reconcile_model_oca/README.rst
@@ -58,9 +58,13 @@ Authors
 Contributors
 ------------
 
-- Dixmit
+-  Dixmit
 
-  - Enric Tobella
+   -  Enric Tobella
+
+-  `Tecnativa <https://www.tecnativa.com>`__:
+
+   -  Víctor Martínez
 
 Maintainers
 -----------
--- a/account_reconcile_model_oca/views/account_reconcile_model_views.xml
+++ b/account_reconcile_model_oca/views/account_reconcile_model_views.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_account_reconcile_model_form" model="ir.ui.view">
+        <field name="name">account.reconcile.model.form</field>
+        <field name="model">account.reconcile.model</field>
+        <field name="inherit_id" ref="account.view_account_reconcile_model_form" />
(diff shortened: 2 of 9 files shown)


pedrobaeza:
Is this still draft?

victoralmau:
Customer feedback is still pending.

victoralmau:
Now it is ready for review.
Ping @pedrobaeza

account_reconcile_model_oca/README.rst
```diff
@@ -58,9 +58,9 @@ Authors
 Contributors
 ------------
 
-- Dixmit
+-  Dixmit
 
-  - Enric Tobella
+   -  Enric Tobella
```
pedrobaeza, line 63: Add yourself in contributors.

pedrobaeza (approved):

account_reconcile_model_oca/views/account_reconcile_model_views.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_account_reconcile_model_form" model="ir.ui.view">
+        <field name="name">account.reconcile.model.form</field>
+        <field name="model">account.reconcile.model</field>
+        <field name="inherit_id" ref="account.view_account_reconcile_model_form" />
+        <field name="arch" type="xml">
+            <field name="matching_order" position="after">
+                <field name="unique_matching" />
```
carlos-lopez-tecnativa: LGTM, just a doubt, should this field always be visible, even when the rule type is `Button to generate counterpart entry`, or is it independent of this setting?

image

victoralmau: The option that is added only applies to the “Rule to match invoices/bills” type.

carlos-lopez-tecnativa: > The option that is added only applies to the “Rule to match invoices/bills” type.

Ok, I suspected that. So, I think the new field should be displayed only when the type is `Rule to match invoices/bills`, and it should be set to `False` when it was previously `True` but the type is changed to another one, to prevent bad configurations or user confusion. WDYT?

carlos-lopez-tecnativa (approved):
LGTM

PR #863, opened 2025-06-19
[16.0][IMP] account_reconcile_oca: Defining the appropriate views in the fiscalyear_lock action

victoralmau:
Defining the appropriate views in the fiscalyear_lock action

Example use case:
- Install the account_lock_date_update module.
- Go to Invoicing > Accounting > Accounting > Actions > Update lock dates
- Create a bank statement line (unreconciled) with today's date
- Define "Lock Date for All Users with today's date

**Before (1 statement line)**
antes-1

**After (1 statement line)**
despues-1

**Before (multi statement line)**
antes-multi

**After (multi statement line)**
despues-multi

@Tecnativa TT56805

### DIFF
--- a/account_reconcile_oca/models/res_company.py
+++ b/account_reconcile_oca/models/res_company.py
@@ -1,4 +1,5 @@
 # Copyright 2024 Dixmit
+# Copyright 2025 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import fields, models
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
Isn't this something to be included in account_statement_base instead, or the problem is due to this module?

victoralmau:
I have thought the same initially, that it should be added in `account_statement_base` but...what views are used then? 
IMO the most appropriate ones are `account_reconcile_oca`, and for that reason I have added the code in this module.

pedrobaeza (approved):

etobella (approved):
I like the idea :smile:

PR #871, opened 2025-06-24
[16.0][IMP] account_statement_base: Add statement_id field to statement lines tree view

victoralmau:
Add `statement_id` field to statement lines tree view

ejemplo

@Tecnativa TT56879

### DIFF
--- a/account_statement_base/views/account_bank_statement_line.xml
+++ b/account_statement_base/views/account_bank_statement_line.xml
@@ -64,6 +64,11 @@
                     name="date"
                     attrs="{'readonly': [('is_reconciled', '=', True)]}"
                 />
+                <field
+                    name="statement_id"
+                    optional="hide"
+                    domain="[('journal_id', '=', journal_id)]"
+                />
                 <field name="move_id" optional="hide" required="0" />
                 <field
                     name="payment_ref"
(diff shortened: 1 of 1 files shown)


account_statement_base/views/account_bank_statement_line.xml
```diff
@@ -63,6 +63,12 @@
         <tree editable="top" multi_edit="1" decoration-muted="is_reconciled">
             <field name="sequence" />
             <field name="date" attrs="{'readonly': [('is_reconciled', '=', True)]}" />
+            <field
+                    name="statement_id"
+                    optional="hide"
+                    domain="[('journal_id', '=', journal_id)]"
+                    attrs="{'readonly': [('is_reconciled', '=', True)]}"
```
pedrobaeza: You can change the statement even in reconciled, so remove this.

account_statement_base/views/account_bank_statement_line.xml
```diff
@@ -63,6 +63,11 @@
         <tree editable="top" multi_edit="1" decoration-muted="is_reconciled">
             <field name="sequence" />
             <field name="date" attrs="{'readonly': [('is_reconciled', '=', True)]}" />
+            <field
+                    name="statement_id"
```
pedrobaeza, line 68: Incorrect indentation

victoralmau: I know that the identation is not right, I don't like it either, but pre-commit defines it that way.

pedrobaeza: OK, I see why, because the first 2 tags are at the same indentation. You can add a refactor commit to fix that.

pedrobaeza (approved):
Thank you!

marcelsavegnago (approved):
LGTM

PR #881, opened 2025-07-11
[16.][FIX] account_reconcile_oca: Dates of new move when using Reconcile Mode = 'Keep suspense accounts'

JordiBForgeFlow:
When using Reconcile Mode = 'Keep suspense accounts' the new move created after the reconciliation should by respect the lock dates, if set.

cc @etobella @JordiMForgeFlow


JordiBForgeFlow:
Replaced by https://github.com/OCA/account-reconcile/pull/887

PR #882, opened 2025-07-17
[16.0][FIX] account_reconcile_oca: Correct open action for Journal Items

antoniospneto:
When clicking "Journal Items" on a Bank journal in dashboard, the system incorrectly launched the OCA reconciliation view  instead of displaying the regular Journal Items. See:

https://github.com/user-attachments/assets/693a170e-00e6-44a3-83b9-e77443863380

This patch removes/adjusts the override that hijacked the action and re-establishes the expected account move line tree 

Note: Clicking the journal name continues to open the OCA Reconcile All view, per PR #763. This PR only restores the proper Journal Items (account.move.line) action and ensures no regressions in the remaining action views.

cc @chienandalu @pedrobaeza

### DIFF
--- a/account_reconcile_oca/models/account_journal.py
+++ b/account_reconcile_oca/models/account_journal.py
@@ -33,10 +33,13 @@ class AccountJournal(models.Model):
         return _("Well done! Everything has been reconciled")
 
     def open_action(self):
-        self.ensure_one()
-        if self.type not in ["bank", "cash"]:
-            return super().open_action()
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

felipemotter (approved):

ValentinVinagre (approved):

pedrobaeza:
To be fw-ported.

PR #884, opened 2025-07-18
[16.0][IMP] account_reconcile_oca: Fix update of external currency

etobella:
With multiple currencies, changing the value in currency had no effect.

With this it should work properly

@ValentinVinagre

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -80,6 +80,12 @@ class AccountBankStatementLine(models.Model):
         prefetch=False,
         currency_field="manual_in_currency_id",
     )
+    previous_manual_amount_in_currency = fields.Monetary(
+        store=False,
+        default=False,
+        prefetch=False,
+        currency_field="manual_in_currency_id",
+    )
     manual_exchange_counterpart = fields.Boolean(
         store=False,
     )
@@ -398,6 +404,7 @@ class AccountBankStatementLine(models.Model):
             "currency_id"
         ] != line.get("line_currency_id")
         self.manual_amount_in_currency = line.get("currency_amount")
+        self.previous_manual_amount_in_currency = line.get("currency_amount")
         self.manual_name = line["name"]
         self.manual_exchange_counterpart = line.get("is_exchange_counterpart", False)
         self.manual_partner_id = line.get("partner_id") and line["partner_id"][0]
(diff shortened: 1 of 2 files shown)


ValentinVinagre:
We will prioritize it and review it in the next few days.
Ty 😸

Jesarregui (approved):
@etobella Thank you very much for your effort and work.

Functional review: LGTM

Functional review completed successfully. The currency exchange now correctly affects the module when working with different currencies. The module works perfectly.

To carry out the relevant tests, the provided Runboat environment was used and the following configuration was applied:

A journal was created with a foreign currency different from EUR — in this case, USD.

An exchange rate was set from EUR to USD with double the base value.

image


Based on this setup, the following tests were performed:

Partial reconciliation of a transaction in USD entered in the USD journal via the manual tab. This partial reconciliation shows how each entry correctly reflects the currency value.

image

image


Partial reconciliation including a match between a foreign currency invoice and a manual accounting entry.

image

All tests produced satisfactory results.

I only identified a minor visual issue.

When validating the journal entry. For example, in one case we split an amount into $600 and $400. Upon validation, the foreign currency information changes and displays the total as $1,000. 

image

image

This is worth mentioning, although I do not consider it critical, since the accounting entry correctly reflects the exchange rate.

image

Jesarregui:
@ValentinVinagre Functional review completed. You may proceed with the technical review if you consider it appropriate. Thank you.

account_reconcile_oca/models/account_bank_statement_line.py
```diff
@@ -80,6 +80,12 @@ class AccountBankStatementLine(models.Model):
         prefetch=False,
         currency_field="manual_in_currency_id",
     )
+    previous_manual_amount_in_currency = fields.Monetary(
```
pedrobaeza, line 83: Can we avoid to add yet another field? What happen with data previous to this update?

etobella: It is not stored, it will be filled when we click in the line.

It was the only way I found to know that the specific field was changed.

Otherwise, JS was doing weird things and the execution of the function was done 2 times, making everything break :cry:

pedrobaeza (approved):
OK, let's go this way:

/ocabot merge patch

Please fw-port it

etobella:
Next week we will port it to 17 an 18

PR #887, opened 2025-07-21
[16.][FIX] account_reconcile_oca: Dates of new move when using Reconcile Mode = 'Keep suspense accounts' 

JordiBForgeFlow:
'

the new move created after the reconciliation should by respect the lock dates, if set.


JordiBForgeFlow:
@etobella Do you agree with this PR?

etobella (approved):

PR #890, opened 2025-07-23
[16.0][IMP] account_reconcile_oca: Fix the amount in currency once we reconcile. Add tests

etobella:
We found that before this change, the widget was showing the full currency amount.

Without the changes we have:

image

Now we have the right amount

[Grabación de pantalla desde 2025-07-23 06-23-44.webm](https://github.com/user-attachments/assets/c67ec8e6-25da-4ffe-9ebd-fc5f9538aebe)

@HaraldPanten @ValentinVinagre Thanks for finding this :wink:

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -752,7 +752,7 @@ class AccountBankStatementLine(models.Model):
                     data += lines
             else:
                 reconcile_auxiliary_id, lines = self._get_reconcile_line(
-                    line, "other", from_unreconcile=False
+                    line, "other", from_unreconcile=False, is_reconciled=True
                 )
                 data += lines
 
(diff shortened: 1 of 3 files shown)


account_reconcile_oca/models/account_bank_statement_line.py
```diff
@@ -1157,20 +1157,10 @@ def _get_reconcile_line(
         self,
         line,
         kind,
-        is_counterpart=False,
-        max_amount=False,
-        from_unreconcile=False,
         reconcile_auxiliary_id=False,
-        move=False,
+        **kwargs,
```
pedrobaeza: Changing this is not good for a stable version.

pedrobaeza (approved):

account_reconcile_oca/models/account_bank_statement_line.py
etobella: :+1:

JordiBForgeFlow (approved):

PR #893, opened 2025-07-24
[16.0][FIX] account_reconcile_oca: Don't use residual on already reconciled lines

etobella:
There was a small bug. The reason was using residual amounts.

@HaraldPanten


HaraldPanten (approved):
LGTM

Functional review in runboat and another test database with other transactions.

@pedrobaeza could you make a quick review?

THX

HaraldPanten (changes requested):
Wait. First I want to tell you one comment

JordiBForgeFlow (approved):

etobella:
The main issue that @HaraldPanten had was related to exchange rates, in that case, the right amount was not showed. I already fixed it and it should work as expected now. Added tests for that part too

HaraldPanten (approved):
LGTM.

PR #786, opened 2025-01-21
[IMP] account_reconcile_oca: Reconcile multiple lines

Daemo00:
Add a wizard to apply a reconciliation model to multiple statement lines.

PR #899, opened 2025-07-27
[18.0][IMP] account_reconcile_oca: Reconcile multiple lines

Daemo00:
Add a wizard to apply a reconciliation model to multiple statement lines.

Forward port of https://github.com/OCA/account-reconcile/pull/786.


Daemo00:
One year has passed 🕯️ time really flies!
@etobella as module maintainer, what do you think of this change?

account_reconcile_oca/wizards/reconcile_multiple_lines.py
```diff
@@ -0,0 +1,46 @@
...
+from odoo import _, fields, models
+from odoo.exceptions import UserError
+
+
+class ReconcileMultipleLines(models.TransientModel):
+    _name = "account_reconcile_oca.reconcile_multiple_lines"
+    _description = "Reconcile multiple lines with a reconciliation model"
+
+    manual_model_id = fields.Many2one(
+        comodel_name="account.reconcile.model",
```
etobella, line 13: Maybe you should filter which models do you allow to use.

etobella (commented):
Sorry for the delay, generally LGTM from a technical perspective, however, I think that it should apply only to manual models, not automatic (for example)

account_reconcile_oca/wizards/reconcile_multiple_lines.py
Daemo00: > I think that it should apply only to manual models, not automatic (for example)

I have added a `domain` using `auto_reconcile`, is that what you meant?

Daemo00:
> Sorry for the delay

No worries, I am not expecting any ETA of course, but at least a reply since the bot was so nice to mention you 😄

etobella (approved):
Technical Review

PR #831, opened 2025-04-15
[17.0][FIX] account_mass_reconcile: writeoff with currencies of opposite signs

astirpe:
It appeared to be a bug related to account_mass_reconcile on method `create_write_off()` when creating write off journal items with different currencies. An error is raised, caused by this line:
https://github.com/OCA/account-reconcile/blob/17.0/account_mass_reconcile/models/base_reconciliation.py#L163

which triggers this standard Odoo sql constraint:
https://github.com/odoo/odoo/blob/17.0/addons/account/models/account_move_line.py#L428

Here is the full error:

```
2025-04-15 08:13:39,661 249 INFO odoo odoo.addons.account_mass_reconcile.tests.test_scenario_reconcile: ====================================================================== 
2025-04-15 08:13:39,661 249 ERROR odoo odoo.addons.account_mass_reconcile.tests.test_scenario_reconcile: ERROR: TestScenarioReconcile.test_reconcile_with_writeoff_currency_opposite_signs
Traceback (most recent call last):
  File "/__w/account-reconcile/account-reconcile/account_mass_reconcile/models/mass_reconcile.py", line 201, in run_reconcile
    ml_rec_ids = self.with_env(new_env)._run_reconcile_method(method)
  File "/__w/account-reconcile/account-reconcile/account_mass_reconcile/models/mass_reconcile.py", line 150, in _run_reconcile_method
    return auto_rec_id.automatic_reconcile()
  File "/__w/account-reconcile/account-reconcile/account_mass_reconcile/models/base_reconciliation.py", line 30, in automatic_reconcile
    return self._action_rec()
  File "/__w/account-reconcile/account-reconcile/account_mass_reconcile/models/base_advanced_reconciliation.py", line 211, in _action_rec
    result = self._rec_auto_lines_advanced(credit_lines, debit_lines)
  File "/__w/account-reconcile/account-reconcile/account_mass_reconcile/models/base_advanced_reconciliation.py", line 310, in _rec_auto_lines_advanced
    reconciled_ids = self._rec_group(reconcile_groups, lines_by_id)
  File "/__w/account-reconcile/account-reconcile/account_mass_reconcile/models/base_advanced_reconciliation.py", line 232, in _rec_group
    reconciled, full = self._reconcile_lines(group_lines, allow_partial=True)
  File "/__w/account-reconcile/account-reconcile/account_mass_reconcile/models/base_reconciliation.py", line 220, in _reconcile_lines
    writeoff_line = self.create_write_off(
[...]

PR #901, opened 2025-07-28
[18.0][FIX] account_mass_reconcile: writeoff with currencies of opposite signs

dcatarino:
This PR applies a fix on the migration of https://github.com/OCA/account-reconcile/pull/817 (version 18.0 - still unmerged)

The issue is described on this PR: https://github.com/OCA/account-reconcile/pull/831 (version 17.0)


StefanRijnhart (changes requested):
You'd need to cherry pick the commits from the 17.0 branch onto your own branch in order to preserve ownership. However, because the target branch is actually the unmerged migration branch, you can review the migration PR, mention the bug fixing 17.0 PR on the migration branch, and if you want to be helpful, propose a PR onto the branch from the migration PR rather than a separate PR here. You should also review the bug fixing 17.0 PR.

PR #911, opened 2025-08-27
[16.0][FIX] account_reconcile_oca: calculate correct exchange gains with foreign currency statement

hbrunn:
when we have an invoice over USD 100 in an EUR journal that's valued at EUR 50 at that time, and import a bank statement where the USD 100 are EUR 60 according to the bank, we expect an exchange gain of EUR 10 proposed when the reconciliation widget finds the invoice via a reconciliation model.

This is already the case if the reconciliation model is set to auto validation, but the interactive reconciliation returns different results.


pedrobaeza (approved):

giarve (approved):

pedrobaeza:
/ocabot merge patch

Please fw-port it to upper versions.

PR #878, opened 2025-07-08
[18.0][FIX] account_reconcile_oca: Correct filtering in the manual_model_id field by the appropriate allowed journals

victoralmau:
FWP from 17.0: https://github.com/OCA/account-reconcile/pull/870

Correct filtering in the `manual_model_id` field by the appropriate allowed journals

**Before**
antes

**After**
despues

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT56841

PR #914, opened 2025-09-09
[18.0][FIX] account_reconcile_oca: auto reconcile journal

jbaudoux:
When importing a bank statement, use only auto reconciliation models for the related journal

cc @etobella @JordiBForgeFlow @victoralmau @hbrunn

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -1,4 +1,5 @@
 # Copyright 2023 Dixmit
+# Copyright 2025 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from collections import defaultdict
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Similar to #878

jbaudoux:
> Similar to #878

I'll adapt domain in the same way to include models without journal restriction

account_reconcile_oca/models/account_bank_statement_line.py
```diff
@@ -975,13 +976,22 @@ def create(self, mvals):
...
+            ("auto_reconcile", "=", True),
+        ]
+        # In case, for example, of bank statement import on a journal
+        journal_id = self.env.context.get("journal_id")
+        if journal_id:
+            domain += [
+                "|",
+                ("match_journal_ids", "=", False),
+                ("match_journal_ids", "in", journal_id),
             ]
```
victoralmau:
```suggestion
        domain += [
            "|",
            ("match_journal_ids", "=", False),
            ("match_journal_ids", "in", result.journal_id.ids),
        ]
```
Although using the context data should not be a problem, I think it is best to use the journal_id data from the created records.

victoralmau (changes requested):

lmarion-source (approved):

account_reconcile_oca/models/account_bank_statement_line.py
jbaudoux: Indeed. I did the change

victoralmau (approved):

PR #915, opened 2025-09-09
[18.0][FIX/IMP] account_reconcile_oca: auto reconcile in job queue & do not auto reconcile lines to not reconcile

jbaudoux:
1/ Do not auto-reconcile lines that do not need to be reconciled. They are already marked as reconciled.

2/ When importing a journal bank statement (lke for example a statement from a payment provider), the auto reconciliation will trigger on each line and it takes ages to finalize.
Allow to run the auto reconcilation asynchronously with queue jobs. Added hook in existing module and added a new module for this.
Added a company setting to enable this.

@pedrobaeza @etobella @victoralmau


account_reconcile_oca/models/account_bank_statement_line.py
```diff
@@ -983,45 +983,53 @@ def create(self, mvals):
...
-                    reconcile_auxiliary_id,
-                    self.manual_reference,
+            record._auto_reconcile(models)
+        return result
+
+    def _auto_reconcile(self, models):
+        # Install module account_reconcile_oca_queue to run this method in a job queue
+        self._auto_reconcile_async(models)
+
+    def _auto_reconcile_async(self, models):
```
pedrobaeza: This name should be fine-tuned.

pedrobaeza (commented):
OK about providing hooks. The method names should be fine-tuned, and this can lead to problems if you reconcile something before the job has been executed.

account_reconcile_oca/models/account_bank_statement_line.py
```diff
@@ -983,45 +983,58 @@ def create(self, mvals):
...
-                    data,
-                    reconcile_auxiliary_id,
-                    self.manual_reference,
+            if not record.is_reconciled:
+                record._enqueue_auto_reconcile(models)
+        return result
+
+    def _enqueue_auto_reconcile(self, models):
+        # Install module account_reconcile_oca_queue to run this method in a job queue
+        self._auto_reconcile(models)
```
etobella: With https://github.com/OCA/queue/blob/18.0/queue_job/models/base.py#L145 we don't need the two functions. can you review? This way, everything will be cleaner 😉 

We used that on edi-framework 

https://github.com/OCA/edi-framework/pull/192/commits/a2d277eae5bde8dd2dc4feccb0b315a128c1ee56

jbaudoux: Thanks. That's what I was looking for :)

jbaudoux: I dropped that indirection from base module and added in second one with `patch_method`. https://github.com/OCA/account-reconcile/pull/915/files#diff-4acf432b981146edced137de305496605a73d033ec2931f544a353375499761bR11

jbaudoux:
@pedrobaeza @etobella Can you update your review? I updated the method names and used `_patch_job_auto_delay` like suggested

account_reconcile_oca/models/account_bank_statement_line.py
```diff
@@ -976,57 +976,76 @@ def create(self, mvals):
...
-                        data, res["model"], reconcile_auxiliary_id
-                    ),
-                    self.manual_reference,
+        result._auto_reconcile()
+        return result
+
+    def _auto_reconcile(self):
+        """Try to auto reconcile records that are not yet reconciled"""
+        non_reconciled = self.filtered(lambda rec: not rec.is_reconciled)
+        if non_reconciled:
```
pedrobaeza: I think this `if` can go out for saving one indentation level.

account_reconcile_oca/tests/common.py
```diff
@@ -0,0 +1,85 @@
+from odoo import Command
+
+from odoo.addons.account_reconcile_model_oca.tests.common import (
+    TestAccountReconciliationCommon as TestAccountReconciliationModelCommon,
+)
+
+
+class TestAccountReconciliationCommon(TestAccountReconciliationModelCommon):
```
pedrobaeza: Don't move this to a new file, but keep it in the same one with 2 classes for reducing the diff.

jbaudoux: Won't it execute the tests when importing the Common class if it's in the same file?

jbaudoux: ok, apparently, it's fine. I merged it back in same file :)

pedrobaeza: Odoo only executes methods that start with `test_`.

pedrobaeza (approved):

etobella (approved):
/ocabot merge minor

PR #922, opened 2025-10-16
[18.0][MIG] account_bank_statement_reconcile_option: Migration to 18.0

nguyenminhchien:
This module was initially part of the [Foodcoop set of addons](https://github.com/AwesomeFoodCoops/odoo-production/tree/12.0). We're now migrating it to 18.0 and taking this occasion to contribute it to OCA.

## Summary
This module improves account reconciliation module. Once installed, accounting user can reduce reconciliation proposition made by Odoo, filtering only on the current accounts of the given journal.

## Configuration
Go to each journal you want to configure, then set 'Bank Reconcile Account Allowed' and 'Search Limit Days':
image


anothingguy (commented):
error when upgrading module sometimes

PR #923, opened 2025-10-21
[16.0][IMP] account_reconcile_oca: default filters on Reconcile tab

alexis-via:
New behavior in the bank statement reconcile interface: when the bank statement line has a partner, the "Reconcile" tab has 2 filters:
- filter "partner_id"
- filter "Payable or Receivable" (added)

That way, you do not see the VAT lines by default in the "Reconcile" tab when the bank statement line has a partner, but only the receivable/payable accounts.

When there is no partner on the bank statement line, the behavior is unchanged (there are no filters in the "Reconcile" tab)

BEFORE:
account_reconcile_oca-before

AFTER:
account_reconcile_oca-after

### DIFF
--- a/account_reconcile_oca/views/account_bank_statement_line.xml
+++ b/account_reconcile_oca/views/account_bank_statement_line.xml
@@ -238,7 +238,7 @@
                             name="add_account_move_line_id"
                             widget="account_reconcile_oca_match"
                             domain="[('parent_state', '=', 'posted'),('amount_residual', '!=', 0),('account_id.reconcile', '=', True), ('company_id', '=', company_id), ('statement_line_id', '!=', id)]"
-                            context="{'search_default_partner_id': partner_id, 'tree_view_ref': 'account_reconcile_oca.account_move_line_tree_reconcile_view', 'search_view_ref': 'account_reconcile_oca.account_move_line_search_reconcile_view'}"
+                            context="{'search_default_partner_id': partner_id, 'search_default_trade_payable': partner_id, 'search_default_trade_receivable': partner_id, 'search_default_payment_asset_current': partner_id, 'tree_view_ref': 'account_reconcile_oca.account_move_line_tree_reconcile_view', 'search_view_ref': 'account_reconcile_oca.account_move_line_search_reconcile_view'}"
                         />
                     </page>
                     <page name="manual" string="Manual operation">
--- a/account_reconcile_oca/views/account_move_line.xml
+++ b/account_reconcile_oca/views/account_move_line.xml
@@ -99,6 +99,11 @@
                     help="From Trade Receivable accounts"
                     name="trade_receivable"
                 />
+                    <filter
+                    string="Outstanding Payments/Receipts"
+                    name="payment_asset_current"
+                    domain="[('payment_id', '!=', False), ('account_id.account_type', '=', 'asset_current')]"
+                />
                 </search>
         </field>
     </record>


alexis-via:
This implementation only uses XML. If you think it's dirty to code:
```
'search_default_trade_payable': partner_id
```
then I have another implementation in JS:
```
diff --git a/account_reconcile_oca/static/src/js/widgets/reconcile_move_line_widget.esm.js b/account_reconcile_oca/static/src/js/widgets/reconcile_move_line_widget.esm.js
index eef08084..527ac78e 100644
--- a/account_reconcile_oca/static/src/js/widgets/reconcile_move_line_widget.esm.js
+++ b/account_reconcile_oca/static/src/js/widgets/reconcile_move_line_widget.esm.js
@@ -15,7 +15,7 @@ export class AccountReconcileMatchWidget extends Component {
         });
     }
     get listViewProperties() {
-        return {
+        const result = {
             type: "list",
             display: {
                 controlPanel: {
@@ -38,6 +38,11 @@ export class AccountReconcileMatchWidget extends Component {
             parentField: this.props.name,
             showButtons: false,
         };
+        if (this.props.record.data.partner_id) {
+            result.context.search_default_trade_payable = true;
+            result.context.search_default_trade_receivable = true;
+        }
+        return result;
     }
 }
 AccountReconcileMatchWidget.template = "account_reconcile_oca.ReconcileMatchWidget";
```

I have no preference between the pure-XML implementation and the JS implementation.

alexis-via:
I now have some first real-life feedback on this new default filter: it would be great to also include the payment outbound account and the payment inbound account.

alexis-via:
I implemented the improvement, following the real-life feedback that I received:

account_reconcile_oca_v2

When the bank statement line has a partner, the "Reconcile" tab has 2 filters:
    - filter "partner_id"
    - filter "Payable or Receivable or Outstanding Payments/Receipts" (added)

That way, you do not see the VAT lines by default in the "Reconcile" tab  when the bank statement line has a partner, but only the  receivable/payable accounts and asset_current accounts linked to a payment.

vdewulf:
Tested functionnaly on runboat, it works as described!
Thanks @alexis-via

bguillot (approved):
Used in production, it works well.
I'm fine with the XML implementation.

Thanks

pedrobaeza (changes requested):
This sounds weird, as this implies that your VAT accounts have the `allow_reconcile` flag, which is something abnormal IMO. On the other hand, when you reconcile payment orders, that are accounts that are not AR/AP, you are filtering out initially, so I don't think this is a good thing.

alexis-via:
I confirm that all our VAT accounts have reconcile=True. This is very common, at least in France, to reconcile VAT accounts. If you use l10n_fr_account_vat_return, having reconcile=True on autoliquidation VAT accounts is required, cf https://github.com/OCA/l10n-france/blob/16.0/l10n_fr_account_vat_return/models/l10n_fr_account_vat_return.py#L485

Of course, it is very important to support the payment orders/debit orders scenario in the default filters. The scenario with payment orders is covered by the default filters of this PR, thanks to the filter 'search_default_payment_asset_current' which has the domain [('payment_id', '!=', False), ('account_id.account_type', '=', 'asset_current')]. We use this PR in production at Akretion and we use payment orders.

pedrobaeza:
Well, that's not the normal thing, as in Spain, your VAT accounts (472 and 477 are regular ones), and when you do the regularization, then you move the balance to a reconcilable account, and that one should be visible. You reconcile one item, not hundred or thousands.

On contrary, payment/debit orders generates journal items to reconcilable accounts that are not AR/AP, and you want to see them even if the partner is set.

So I still think this is not a good idea at all.

alexis-via:
The scenario with payment/debit orders that generate journal items to reconcilable accounts that are not AR/AP **is supported by this PR**. Cf the filter  'search_default_payment_asset_current' which is enabled when the partner is set.

alexis-via:
Please refer to the last screenshot I posted in the comments, not to the screenshot of the first comment (which is outdated).

etobella (changes requested):
Sorry, but I do agree with @pedrobaeza in this topic, let me explain:

Using taxes as you did is crazy IMO. It has much more sense to have a second account that is reconciliable and port aggregated amounts to it at some point. Why?  Because your model will be unsustainable to assume if you have 1.000 invoices each day (as I have in some customers).

You are trying to modify this when it would be cleaner to make taxes non-reconciliable.

vdewulf:
Hello, for Belgium, we also set the VAT accounts as reconciliable. It's a common practice and I do that for all my customers with accounting on Odoo.
I am interested in the improvement proposed here by alexis for my own accounting company as it would save me a lot of time to find the right lines to reconcile, as right now I see many lines proposed by Odoo in this screen that comes fromt the open VAT lines that are not reconciled yet (we do the reconciliation on the VAT accounts each quarter for many customers).

etobella:
In case you want to do this kind of approach, I would recommend to set it in a different module.

I think it would be better to define tax.document generator that could solve this kind of issues. It would be cleaner for me. 

- All taxes are no reconciliable. For each tax account, we have a reconciliable account.
- At the tax closure, we define a new `tax.document` that makes the calculation (in Spain, we even generate the file to send to the tax agency).
- After approving the document, a new move is done, porting from this non reconciliable account to the reconciliable account to make the payment.

I think we can reuse Spanish work to make it.

alexis-via:
@etobella Which module do you use to generate the journal entry between the non-reconciliable VAT accounts and the reconciliable VAT accounts ?

etobella:
It is a set of modules. For example we have https://github.com/OCA/l10n-spain/tree/18.0/l10n_es_aeat_mod303 that moves customer VAT taxes  - supplier vat taxes to a final value. Also generates the document to send to the tax agency (In spain it is a weird XML file, for this reason we have this complex system)

pedrobaeza:
But the thing is that it's insane to have thousands of VAT entries to reconcile in one shot when you settle the VAT report. That's why Spanish accounts for VAT issued/received are not reconcilable (477/472 respectively), but they are balanced to 0 when the VAT report is done, and the debt is transferred to the 475000 account, which is reconcilable, and we want to see it in the reconcile interface (and it has a partner set, which is the Tributary Agency), so no extra filters are needed.

etobella:
I think that Odoo will not handle properly a reconciliation of the taxes for 100 invoices a day if we do the tax return each month (2.000 tax lines at least...).

For small companies, this approach might work, but once we increase numbers, the reconciliable taxes will not work.

alexis-via:
I agree with you that, for companies that have a high volume of sale/purchase, having reconcile=True on VAT accounts is not a good option. But, for companies with reasonable volume of sale/purchase, I think it's a good option (at Akretion, our VAT accounts have reconcile=True and it works very well).

To move forward on this, I see 3 options:
1) add a configuration parameter in account_reconcile_oca to have this filter or not
2) move the code of this PR to another module that depend on account_reconcile_oca.
3) change this additional filter: instead of filtering on Payable+Receivable+Outstanding Payments/Receipts, maybe we could imagine a filter that filters-out the VAT accounts configured on taxes.

pedrobaeza:
@alexis-via even if you don't have too many invoices, let's put 50 per month, doing a quarterly VAT report means to reconcile 150 entries against the VAT payment, which is a lot. I don't think you should continue with that approach, as it doesn't scale. And all the reconcilable changes are in your fork `l10n_fr_oca`, not in `l10n_fr`, isn't it?

@vdewulf does `l10n_be` also have the check "to reconcile" in the VAT accounts, or is it something that you have changed afterwards?

etobella:
@alexis-via as I mentioned, it doesn't escalate properly, so it shouldn't be in account_reconcile_oca.

vdewulf:
@pedrobaeza to make the VAT accounts reconciliable, it's not in the l10n_be, this is something I change for my customers on their chart of accounts.
The reconciliation is not an issue until now. For example one of my customers had 1606 entries to reconcile on one VAT account for one quarter in 2024, the reconciliation was ok in version 16 CE. It's true that in v9 where this customer was before, it was not always working but since v16 it's working good.

I like the option 3 suggested by @alexis-via. Actually, the technical solution is not super important to me, I just want to not see the VAT entries in the bank reconciliation screen if possible. That would help a lot.

Thanks!

pedrobaeza:
As this is something you are deforming on purpose, I don't think the solution should be in the standard module.

vdewulf:
@pedrobaeza well, it's also a point of view. Many configurable options in Odoo or usages in Odoo in general could be seen as "deforming" what Odoo intended when writing the code at first. That's what make Odoo powerfull. And I don't understand in what the proposal here will bring issues in the code of the module.

@alexis-via I am open to discuss with you an alternative to do what we need for our customers if it is not accepted by Pedro nor @etobella here.

etobella:
@vdewulf We are not saying that this shouldn't exist, we are asking for a secondary module. At the end, it is easy to add a new module :wink:

alexis-via:
Code moved to a separate module in this PR https://github.com/OCA/account-reconcile/pull/946

PR #926, opened 2025-10-31
[16.0][FIX] account_reconcile_analytic_tag: Apply the analytical tag to the account move

victoralmau:
Apply the analytical tag to the account move

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT59001


pedrobaeza (approved):
Fast-tracking this patch

/ocabot merge patch

PR #930, opened 2025-11-14
[18.0][MIG] account_reconciliation_widget_due_date: Migration 18.0 + rename to account_reconcile_due_date

eduezerouali-tecnativa:
cc @Tecnativa TT54100
ping @pedrobaeza @victoralmau 

We merge this module to add the field `due_date` to `account.bank.statement.line` so you can select a due date and once it conciliated is set in `date_maturity` in `account.move.line`.  I add some test too.


pedrobaeza:
I think it should continue separated.

eduezerouali-tecnativa:
Okay, but it is really necessary to keep a module just to add a due date. If it is keep empty it works as before. Anyways de field was added here
Selección_009
should we add it here as well? 
Selección_011

sergio-teruel:
@eduezerouali-tecnativa please make the PR with the original module.. 
Thanks

eduezerouali-tecnativa:
Now it has been separated again. The following changes apply during the migration:

- Renamed it, as it is no longer a widget
- Removed all .js files, as they are no longer needed
- Updated the tests to reflect the new behavior
 
ping @pedrobaeza @sergio-teruel @victoralmau

victoralmau (approved):
Functional review OK

eduezerouali-tecnativa:
ping @pedrobaeza

pedrobaeza (approved):
/ocabot migration account_reconciliation_widget_due_date

/ocabot merge nobump

PR #932, opened 2025-11-25
[18.0][IMP] Change sequence field to use handle widget

tarteo:


pedrobaeza:
As the order is by date and then by sequence, I don't think this is a good idea. In fact, the first field should be date, not sequence.

tarteo:
I see, I just tested it and it's not ideal.

PR #933, opened 2025-12-05
[18.0][FIX] account_reconcile_*_oca: Allow to use regex filtering

etobella:
I found that the standard odoo configuration was not working when searching by regex.

For example, demo data includes a mapping for Bank fees that should create but just one was created, with the new one, it works as expected and it gets the data from the label

image


luisDIXMIT (approved):
LGTM!

HaraldPanten (approved):
/ocabot merge patch

PR #934, opened 2025-12-08
[16.0][FIX] account_move_line_reconcile_manual: fix multi-currency full reconcile and reconcile with write-off

alexis-via:


florian-dacosta:
I have made this PR to propose a change : https://github.com/akretion/account-reconcile/pull/4
Indeed, we need the exchange move to happen for mono foreign currency reconciliation, case we implement here https://github.com/OCA/account-reconcile/pull/660

+ the reconciliation cases with multiple or foreign  currencies are complicated and multiple, I think we need to add a test to cover each case we manage by a unitest.

PR #937, opened 2025-12-12
[18.0][IMP] account_reconcile_oca: Multiple fixes

etobella:
We have detected that when the user goes to another record, like view move, when they return, they are not in the same reconciliation, making it harder to handle.

With this change, this is solved


luisDIXMIT (approved):
Awesome job! Tested on runboat and LGTM!

etobella:
I also found a race condition when there is several elements that could raise an error. The second commit fixes it.

It happens because there is too much elements on the mapping (too many possible lines to reconcile).

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
@etobella I have just seen that this has a side effect: when you are reconciling with the "Not reconciled" filter, and you click on "Validate", the record at the left list disappears, but the right part remains on the previous reconciled element.

etobella:
oh, don't worry. I will review this today

PR #940, opened 2025-12-16
[18.0][FIX] account_reconcile_oca: prevent duplicate aggregates in reconcile kanban renderer

JordiBForgeFlow:
Ensure aggregates are computed only once per aggregate_id by tracking processed IDs with a Set. This avoids repeating the same aggregate when records are not ordered or grouped by aggregate_id.

Previously, only the first aggregate was checked, causing duplicate entries when different aggregate_ids appeared later in the list.

@etobella


pedrobaeza:
@JordiBForgeFlow @etobella is this still relevant?

PR #666, opened 2024-06-07
[15.0][ADD] account_reconcile_sale_order

hbrunn:
This module allows a workflow where you don't invoice sale orders until
you've received a payment.

That's useful ie for webshops with non-instant payment like wire
transfer, where you might have a lot of customers not doing the payment
after all, which results in extra work for cancellation of the
invoices/orders involved.

Configuration
=============

To configure this module, you need to:

1. Go to Invoicing/Configuration/Reconciliation Models
2. Create a model of type *Rule to match sale orders*

Usage
=====

To use this module, you need to:

1. Have a payment on a bank statement matching the amount of an
   *invoicable* sale order
2. Enter the reconciliation screen
3. Observe that the sale order is offered as reconciliation counterpart

Note the reconciliation only works if fully invoicing the sale order
yields an invoice over the order's total amount. Usually this means that
all products in the sale order must have invoicing policy *Ordered
quantities*.

PR #941, opened 2025-12-29
[18.0][ADD] account_reconcile_sale_order

hbrunn:
This module allows a workflow where you don't invoice sale orders until
you've received a payment.

That's useful ie for webshops with non-instant payment like wire
transfer, where you might have a lot of customers not doing the payment
after all, which results in extra work for cancellation of the
invoices/orders involved.

Configuration
=============

To configure this module, you need to:

1. Go to Invoicing/Configuration/Reconciliation Models
2. Create a model of type *Rule to match sale orders*

Usage
=====

To use this module, you need to:

1. Have a payment on a bank statement matching the amount of an
   *invoicable* sale order
2. Enter the reconciliation screen
3. Observe that the sale order is offered as reconciliation counterpart
4. You can manually select and deselect sale orders in the *sale orders*
   tab of the reconciliation widget

Note the reconciliation only works if fully invoicing the sale order
yields an invoice over the order's total amount. Usually this means that
all products in the sale order must have invoicing policy *Ordered
quantities*.

v18 migration of https://github.com/OCA/account-reconcile/pull/666


pedrobaeza:
We are doing this in a PoC using `payment.transaction`, that creates itself a `account.payment` record, and when the invoice is created and validated, then the payment is auto-linked. I think it's the cleanest one.

hbrunn:
is your PoC available somewhere? Though I don't see how what you describe matches the requirement outlined above: No accounting items created for SOs until payment received

pedrobaeza:
Yes, I read bad the purpose of the module. The module name doesn't help either, although technically it makes sense.

We'll publish soon the other module. I'll ping you when done.

jguenat (approved):
Tested it and backported it to 16.0
LGTM
Thanks

account_reconcile_sale_order/readme/USAGE.md
```diff
@@ -0,0 +1,8 @@
+To use this module, you need to:
+
+1. Have a payment on a bank statement matching the amount of an *invoicable* sale order
```
jbaudoux, line 3:
```suggestion
1. Have a payment on a bank statement matching the amount of an *invoiceable* sale order
```

account_reconcile_sale_order/models/account_bank_statement_line.py
```diff
@@ -0,0 +1,129 @@
...
+
+        return super()._prepare_reconcile_line_data(lines)
+
+    def _prepare_reconcile_line_data_sale_order_invoice(self, order):
+        """
+        Invoice selected sale orders and post the invoices
+        """
+        if order.state in ("draft", "sent"):
+            order.action_confirm()
+        invoices = order._create_invoices()
```
jbaudoux, line 125: In case of web order, the payment transaction will remain pending.
What about relying on this module that I did to manually mark as paid a SO when it's paid:
* https://github.com/OCA/sale-workflow/pull/3779/changes#diff-f30e2537e164a1559e233c0888439b4a02876e92e27b81c537517b378edd0c18R11

When managing the transaction, your product can be invoiced on delivered quantity, odoo will still manage it properly. Odoo then also sends the email to the customer according to what you opt-in.

I'll see to add in my module the use case when it's a manual quotation without transaction.

jbaudoux (commented):
Thanks for the module. It's a nice addition.

Small english remark: can you rename `sale order` by `sales order` for the labels, comments...

jbaudoux (approved):
Approving the technical migration

PR #942, opened 2026-01-09
[18.0][FIX] account_reconcile_oca: Day aggregation was failing.

etobella:
We are pushing data, so we should check with the last element

@pedrobaeza 

Before:

image

After:

image


luisDIXMIT (approved):
Code review, LGTM!

pedrobaeza (approved):
/ocabot merge patch

PR #944, opened 2026-01-13
[18.0][IMP] account_reconcile_oca: Add @mute_logger("odoo.models.unlink") to tests to reduce logs

victoralmau:
Add `@mute_logger("odoo.models.unlink")` to tests to reduce logs

@Tecnativa

### DIFF
--- a/account_reconcile_oca/tests/test_bank_account_reconcile.py
+++ b/account_reconcile_oca/tests/test_bank_account_reconcile.py
@@ -2,6 +2,7 @@ import time
 
 from odoo import Command
 from odoo.tests import Form, tagged
+from odoo.tools import mute_logger
 
 from odoo.addons.account_reconcile_model_oca.tests.common import (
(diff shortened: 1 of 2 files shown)


account_reconcile_oca/tests/test_bank_account_reconcile.py
```diff
@@ -123,6 +124,7 @@ def test_reconcile_invoice_currency(self):
             self.assertFalse(f.add_account_move_line_id)
             self.assertTrue(f.can_reconcile)
 
+    @mute_logger("odoo.models.unlink")
```
pedrobaeza, line 127: Can you put this at class level instead?

victoralmau: It's not possible.

pedrobaeza (approved):
/ocabot merge patch

PR #889, opened 2025-07-22
[18.0][MIG] account_move_reconcile_helper

BhaveshHeliconia:

PR #953, opened 2026-01-30
[18.0][FIX] account_move_base_import: Use dedicated partner to fix test failure

BhaveshHeliconia:
This issue was also observed in: https://github.com/OCA/account-reconcile/pull/889

This PR fixes a test failure in account_move_base_import caused by a change in Odoo 18 demo data.

Issue: The test TestInvoice.test_all_completion_rules relied on the partner "Deco Addict" from the base demo data. In Odoo 18, this partner was renamed, causing the test to fail. Initially, we considered updating the name to "Acme Corporation", but relying on unstable demo data makes the test fragile to future changes.

Solution: Refactored tests/test_invoice.py to create a dedicated partner "TestPartner" in setUpClass. The test now uses this specific partner for the completion rule test, removing dependency on demo data and ensuring robustness.

### DIFF
--- a/account_move_base_import/tests/test_invoice.py
+++ b/account_move_base_import/tests/test_invoice.py
@@ -26,6 +26,13 @@ class TestInvoice(AccountTestInvoicingCommon):
         )
         cls.partner_agrolait_id = cls.partner_agrolait.id
 
+        cls.partner_autocomplete = cls.env["res.partner"].create(
+            {
+                "name": "TestPartner",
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please do the same as in 16.0

BhaveshHeliconia:
> Please do the same as in 16.0

Done. Thank you!

pedrobaeza (approved):
/ocabot merge patch

PR #954, opened 2026-01-30
[16.0][FIX] account_move_base_import: Use dedicated partner to fix test failure

BhaveshHeliconia:
This PR fixes a test failure in account_move_base_import caused by a change in Odoo 18 demo data, where "Deco Addict" was renamed to "Acme Corporation".

Issue: The test TestInvoice.test_all_completion_rules relied on the partner "Deco Addict" from the base demo data. In Odoo 16, this partner was renamed, causing the test to fail. Initially, we considered updating the name to "Acme Corporation", but relying on unstable demo data makes the test fragile to future changes.

Solution: Refactored tests/test_invoice.py to create a dedicated partner "TestPartner" in setUpClass. The test now uses this specific partner for the completion rule test, removing dependency on demo data and ensuring robustness.

### DIFF
--- a/account_move_base_import/tests/test_invoice.py
+++ b/account_move_base_import/tests/test_invoice.py
@@ -15,6 +15,12 @@ class TestInvoice(TestAccountReconciliationCommon):
         cls.account_move_line_obj = cls.env["account.move.line"]
         cls.journal = cls.company_data["default_journal_bank"]
         cls.account_id = cls.journal.default_account_id.id
+        cls.partner_autocomplete = cls.env["res.partner"].create(
+            {
+                "name": "TestPartner",
+                "is_company": True,
+            }
+        )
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Instead of this, it's better to create a partner and not depend anymore in demo data.

BhaveshHeliconia:
@pedrobaeza Thanks for the feedback. I have updated the test to create a dedicated partner instead of relying on the base demo data. I have also updated the PR description, commit message, and PR title accordingly.

Please let me know if everything is proper now. Also, let me know if we should refactor the entire module's test cases to remove reliance on demo data. Otherwise, I will proceed with applying this specific fix to 18.0.

pedrobaeza (approved):
Great, thanks!

Let's fast-track it to fix the CI.

/ocabot merge patch

PR #957, opened 2026-02-11
[16.0][IMP] account_reconcile_oca: Refactor stmt line creation, reconcile i…

aritzolea:
…n new function

It makes reconcile process inheritable in statement line creation


etobella (changes requested):
It is already inheritable. I don't understand why you need it 🤔 

You can inherit later on...

aritzolea:
> It is already inheritable. I don't understand why you need it 🤔
> 
> You can inherit later on...

You're right. I'm closing PR.
Thanks!

PR #962, opened 2026-03-04
[18.0][FIX] account_reconcile_oca: Account move smartbutton missing context

jguenat:
When opening the "Bank reconcile" on an account move, the reconciliation view is not loaded correctly.

Fixed by adding the reconcile form view in context.

smartbutton
reconcile_view


pedrobaeza (approved):

etobella (approved):
/ocabot merge patch

PR #969, opened 2026-03-11
[18][MIG] account_mass_reconcile

florian-dacosta:
Replace closed PR : https://github.com/OCA/account-reconcile/pull/817
 Contain a light refactore to avoid issue with multi company account thanks to @StefanRijnhart comment here https://github.com/OCA/account-reconcile/pull/817#pullrequestreview-3082950531
 I chose to have a mandatory company_id field, so it is not possible (for now) to share a mass_reconcile configuration between companies sharing a same account.
 I did this because it is simpler/less risky and before v18, it was not possible to share the configuration so we don't loose any feature.
 Multi company case and even more multi company case sharing the accounts are quite rare for now and it would complexify a module already not so simple.

This PR also contain a fix related to https://github.com/OCA/account-reconcile/pull/831(but with a different solution)

Also add the domain field in the view and remove the use of  safe_eval to process it.

Review welcome @lfreeke @dcatarino @StefanRijnhart @amkarthik @astirpe @CasVissers-360ERP @bguillot @sebastienbeau


DorianMAG (approved):
Code review.
Functional test OK.
LGTM

BhaveshHeliconia (approved):
LGTM!

StefanRijnhart (commented):
Thanks! There are some commits that do not follow the convention of mentioning the module name. Maybe some of these commits can be squashed into the migration commit? You can share credits using Co-authored-by.

florian-dacosta:
> Thanks! There are some commits that do not follow the convention of mentioning the module name. Maybe some of these commits can be squashed into the migration commit? You can share credits using Co-authored-by.

I did rename the commit to add the module names to be complient
A squash into the migration commit does not seem necessary to me because for each commit, either there is a refactore which is good to be in a dedicated commit (IMO) either a fix on something that was already broken before the migration.

PR #970, opened 2026-03-11
[16.0][IMP] account_reconcile_oca: replace Narration tab with Statement Details

StuartJMackintosh:
The Narration tab in the reconciliation view shows payment_ref and narration fields without labels, and omits transaction_type, ref, partner_name, account_number and unique_import_id.

Replace with a Statement Details tab showing all statement line fields with labels matching reconciliation model terminology: Label, Type, Reference, Note in the Transaction group, and Partner Name, Account Number, Import ID in the Source group.

### DIFF
--- a/account_reconcile_oca/views/account_bank_statement_line.xml
+++ b/account_reconcile_oca/views/account_bank_statement_line.xml
@@ -319,9 +319,43 @@
                             </group>
                         </group>
                     </page>
-                    <page name="narration" string="Narration">
-                        <field name="payment_ref" />
-                        <field name="narration" />
+                    <page name="narration" string="Statement Details">
+                        <group>
+                            <group string="Transaction">
+                                <field
+                                    name="payment_ref"
+                                    string="Label"
+                                    attrs="{'readonly': [('is_reconciled', '=', True)]}"
+                                />
+                                <field
+                                    name="transaction_type"
+                                    string="Type"
+                                    attrs="{'readonly': [('is_reconciled', '=', True)]}"
+                                />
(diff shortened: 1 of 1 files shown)


account_reconcile_oca/views/account_bank_statement_line.xml
```diff
@@ -319,9 +319,31 @@
                             </group>
                         </group>
                     </page>
-                    <page name="narration" string="Narration">
-                        <field name="payment_ref" />
-                        <field name="narration" />
+                    <page name="narration" string="Statement Details">
+                        <group>
+                            <group string="Transaction">
+                                <field name="payment_ref" string="Label" readonly="1" />
```
etobella: Why readonly? Maybe it should be if and only if the move is not reconciled. WDYT?

etobella (commented):
Can you review your commit name and follow up the guidelines of OCA 🙏

account_reconcile_oca/views/account_bank_statement_line.xml
StuartJMackintosh: good plan, done

etobella (approved):
Thanks! Worked like a charm 😄 

Will you forward port it to 17 and 18?

StuartJMackintosh:
> Will you forward port it to 17 and 18?

I will bed this in first on 16, and have some other changes I want to make as well, re linking to the reconciliation models to show what would match. Once all settled, then will look at porting.

pedrobaeza (approved):
/ocabot merge minor

PR #971, opened 2026-03-14
[18.0][IMP] account_reconcile_sale_order: payment transaction & downpayment

jbaudoux:
Properly mark SO as paid by managing the payment transaction for webshop SO or SO for which you sent a payment link.

This will also:
* send email for payment confirmation to the customer if enabled it in the odoo standard settings
* send the invoice  if enabled it in the odoo standard settings

For manually created quotation, a transaction will also be created.

In case of downpayment on the SO, the reconciliation counterpart is for the open amount on the SO.

Depends on:
* https://github.com/OCA/sale-workflow/pull/3779

cc @hbrunn @pedrobaeza

### DIFF
--- a/account_reconcile_sale_order/__manifest__.py
+++ b/account_reconcile_sale_order/__manifest__.py
@@ -9,10 +9,10 @@
     "category": "Accounting",
     "website": "https://github.com/OCA/account-reconcile",
     "author": "Hunki Enterprises BV, Odoo Community Association (OCA)",
-    "maintainers": ["hbrunn"],
+    "maintainers": ["hbrunn", "jbaudoux"],
     "license": "AGPL-3",
     "depends": [
-        "sale",
+        "sale_action_paid",  # OCA/sale-workflow
         "account_reconcile_oca",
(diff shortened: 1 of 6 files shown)


account_reconcile_sale_order/__manifest__.py
```diff
@@ -9,10 +9,10 @@
     "category": "Accounting",
     "website": "https://github.com/OCA/account-reconcile",
     "author": "Hunki Enterprises BV, Odoo Community Association (OCA)",
-    "maintainers": ["hbrunn"],
+    "maintainers": ["hbrunn", "jbaudoux"],
     "license": "AGPL-3",
     "depends": [
-        "sale",
+        "sale_action_paid",  # OCA/sale-workflow
```
etobella, line 15: why you add an extra dependancy?

jbaudoux: to decouple the reconciliation mecanism from the flow to mark a SO as paid (which is a feature on its own)

PR #978, opened 2026-04-08
[16.0][FIX] account_reconcile_oca: isolate searchModel state in match widget

giarve:
The `account_reconcile_oca_match` widget mounts an inner `View` for account.move.line inside the bank statement line reconcile form. Without isolation, the inner `WithSearch` registers its searchModel state on the parent action's `__getGlobalState__` / `__getLocalState__` / `__beforeLeave__` CallbackRecorders. When the user backtracks via the breadcrumb, action_service merges every recorded `getGlobalState()` under the same `searchModel` key (last write wins), so the inner account.move.line state overwrites the outer
account.bank.statement.line one. The outer kanban then restores with a domain that references account.move.line-only fields (e.g. `account_id.non_trade` from the **Trade Receivable / Trade Payable** filters of `account_move_line_search_reconcile_view`) and crashes `web_search_read` on `account.bank.statement.line`:

    ValueError: Invalid field account.bank.statement.line.account_id
    in leaf ('account_id.non_trade', '=', False)

Fix: override `__beforeLeave__`, `__getGlobalState__` and `__getLocalState__` to `null` in the widget's `useSubEnv`. The inner WithSearch then skips registering its callbacks on the parent action's recorders, so the outer kanban's searchModel state is preserved on breadcrumb back.

### DIFF
--- a/account_reconcile_oca/static/src/js/widgets/reconcile_move_line_widget.esm.js
+++ b/account_reconcile_oca/static/src/js/widgets/reconcile_move_line_widget.esm.js
@@ -2,6 +2,7 @@
 
 import {View} from "@web/views/view";
 import {registry} from "@web/core/registry";
+import {CallbackRecorder} from "@web/webclient/actions/action_hook";
 
 const {Component, useSubEnv} = owl;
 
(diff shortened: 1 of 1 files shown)


etobella (commented):
Interesting...
Silly question, doesn't have sense to use  new CallbackRecorder(); as it is done in definition of this parameters in https://github.com/odoo/odoo/blob/16.0/addons/web/static/src/webclient/actions/action_service.js#L632-L635 ?

giarve:
Yes you are right. I updated it with the new empty callback and it is working. And better than the null which could have caused another error if not properly caught.

Unfortunately, the correct way to proceed would be to save current state in the reconcile wizard but I have not looked into it too much and this quick fix will at least stop crashing the reconcile session.

etobella:
Before approving I tried to replicate the original issue, but I am unable to do it. Can you share how you made it?

giarve:
> Before approving I tried to replicate the original issue, but I am unable to do it. Can you share how you made it?

In the filters add all of them

image

then click on view move blue button to exit the wizard, do some steps (follow other buttons) and after that go back through the breakcumb to the reconcile "statement lines"

image


it must be the filters that have "**account_id**" field on them.


No need to merge for now if you dont need it. Maybe in the future someone else will find it if this was just happening to me

etobella:
Can you share a video recording using runbot? Still doesn't happening to me...

Let's be clear that I agree with the change, but I want to replicate your issue before approval to verify that everything is correct.

giarve:
> Can you share a video recording using runbot? Still doesn't happening to me...
> 
> Let's be clear that I agree with the change, but I want to replicate your issue before approval to verify that everything is correct.

I tried on open runboat instance, cannot replicate.

I will investigate further.

It is possible I added account_reconcile_oca_add_default_filters and they got saved as default filters, but it is unclear to me yet. Before the patch I was getting the error, now it solved it for me, seems to be database issue then.

PR #979, opened 2026-04-09
[ADD] account_reconcile_match_regex

etobella:
Use Cases / Context
===================

Sometimes, we have cases where the invoice is included in the text and
we need to find it somehow. This fixes the problem by passing a regex
and replacing the elements.

Usage
=====

1. Access Reconcile Models
2. On an invoice matching rule, add a regex match
3. From now on, the match will be used for searching and obtaining the
   information to find

For example, if we set "Fattura (.\*)", and we receive a settlment with
the following label "Fattura INV/21/12323", Odoo will transform it to
INV/21/12323 and search for this exact text.


luisDIXMIT (commented):
Can these fields be reused, or do they have another purpose?

Image

etobella:
This fields are intended for filtering, not for transforming the data

luisDIXMIT (approved):
Code review, LGTM!

giarve (approved):

mt-software-de (approved):
Thx. LGTM

PR #980, opened 2026-04-11
[18.0][IMP] account_reconcile_model_oca: Get write off line balance from note via regex

mt-software-de:
Previously, it was only possible to extract the amount of an automatic counterparty line from the payment reference. With this change, the regular expression will also be applied to the note.

@jbaudoux @etobella @victoralmau Could you review this?

### DIFF
--- a/account_reconcile_model_oca/models/account_reconcile_model.py
+++ b/account_reconcile_model_oca/models/account_reconcile_model.py
@@ -1,3 +1,7 @@
+# Copyright 2024 Dixmit
+# Copyright 2025 Victor M.M. Torres, Tecnativa SL
+# Copyright 2026 Jacques-Etienne Baudoux (BICM) <je@bcim.be>
+# Licence LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0).
 import re
 from collections import defaultdict
 
--- a/account_reconcile_model_oca/models/account_reconcile_model.py
+++ b/account_reconcile_model_oca/models/account_reconcile_model.py
@@ -138,7 +138,9 @@ class AccountReconcileModel(models.Model):
             return 0.0
         return self._str2float(m[0])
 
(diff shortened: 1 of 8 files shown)


etobella:
@mt-software-de In the description you asked for one thing, but there is 5 commits with different goals (most of them have sense)

mt-software-de:
> @mt-software-de In the description you asked for one thing, but there is 5 commits with different goals (most of them have sense)

I can remove/put the first commits into an own PR. 
But the others are serving the same purpose, IMO its more readable when the REF of `_get_write_off_amount_from_string` and REF for one model one file is in an own commit.
That makes it easier to read, what actually was changed. 
Let me know how i should change the commits to get this merged.

account_reconcile_model_oca/models/account_reconcile_model.py
```diff
@@ -150,24 +157,16 @@ def _get_write_off_move_lines_dict(self, residual_balance, partner_id, label=Non
 
         lines_vals_list = []
         for line in self.line_ids:
-            balance = 0
+            balance = 0.0
             if line.amount_type == "percentage":
-                balance = currency.round(residual_balance * (line.amount / 100.0))
+                balance = abs(residual_balance) * (line.amount / 100.0)
```
etobella, line 162: currency.round should be needed no matter what....

etobella:
No, no, don't worry. In general it seems ok, however, leaver it in 3 commits please, one for the license issue, one for all the changes in account_reconcile_model_oca and one for account_reconcile_oca. There is no need for so many commits.

account_reconcile_model_oca/models/account_reconcile_model.py
mt-software-de: It still is. It is just refactored and move to the bottom and removes the duplication of Code.

PR #983, opened 2026-04-17
[19.0][MIG] account_statement_reconcile_status: Migration to 19.0

BhaveshHeliconia:


account_statement_reconcile_status/models/account_bank_statement.py
```diff
@@ -0,0 +1,45 @@
...
+            reconciled = len(posted_lines.filtered("is_reconciled"))
+            if total and reconciled == total:
+                stmt.reconcile_state = "done"
+            elif reconciled > 0:
+                stmt.reconcile_state = "in_progress"
+            else:
+                stmt.reconcile_state = "not_started"
+
+    @api.constrains("reconcile_state")
+    def _set_date_first_fully_reconciled(self):
```
etobella: Wouldn't be cleaner to do this like a computed stored field?

BhaveshHeliconia: Thanks for the suggestion. I've refactored it to use a computed stored field. Please have another look when you get a chance.

etobella (approved):
Thanks @BhaveshHeliconia

PR #987, opened 2026-04-24
[17.0][FIX] account_reconcile_oca: incorrect amounts when reconciling with a model and a foreign currency

MaxyMoos:
When reconciling a bank statement line with a foreign currency amount on a journal (in company currency) using a reconciliation model, the resulting write-off line has incorrect/unexpected amounts:

## Before
Starting from this state:
BEFORE_1

Clicking on the reconciliation model button (which is defined to map 100% of the remaining balance) gives out this:
BEFORE_2

which is not what one would expect as it does not reconcile the full balance - a double conversion lands the bank statement line balance onto the `amount_currency` field and then reconverts this for the writeoff balance.

## After
This PR changes the result obtained after clicking on the reconciliation model button to this state of things:
AFTER

We also add a test case for this scenario.

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -595,14 +595,11 @@ class AccountBankStatementLine(models.Model):
                 (partner.id, partner.display_name) if partner else False
             )
             amount = line.get("balance")
-            if self.foreign_currency_id:
-                amount = self.foreign_currency_id._convert(
-                    amount,
-                    self.journal_id.currency_id or self.company_currency_id,
-                    self.company_id,
-                    self.date,
+            if self.foreign_currency_id and self.amount:
+                currency_amount = currency.round(
+                    amount * self.amount_currency / self.amount
                 )
-            if currency != self.company_id.currency_id:
+            elif currency != self.company_id.currency_id:
                 currency_amount = self.company_id.currency_id._convert(
                     amount,
                     currency,
--- a/account_reconcile_oca/tests/test_bank_account_reconcile.py
+++ b/account_reconcile_oca/tests/test_bank_account_reconcile.py
@@ -796,6 +796,82 @@ class TestReconciliationWidget(TestAccountReconciliationCommon):
         self.assertEqual(bank_stmt_line.move_id.line_ids[0].debit, expected_amount)
         self.assertEqual(bank_stmt_line.move_id.line_ids[1].credit, expected_amount)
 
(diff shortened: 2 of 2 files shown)


ryanduguid (commented):
Thanks for this, and for a test that pins the rate rather than the arithmetic.
Setting the implied rate to 2.0 against a system rate of 1.5289 is the right way to
write it, because it fails if the code falls back to `_convert`.

Three things, one of which I think is a blocker.

**This is being fixed twice.** #990 makes the same change on 16.0, opened two weeks
after yours, and neither PR references the other. It deletes the same conversion and
also propagates the statement line's implied rate, but it keeps the change nested
inside the existing `if currency != self.company_id.currency_id:` gate rather than
restructuring to `if/elif`. It would be worth the two of you agreeing one shape,
because 18.0 and 19.0 still carry the original logic at lines 601 and 600 and will
need whichever version wins.

**The restructure changes behaviour in one case.** Core only forbids
`foreign_currency_id == currency_id`, the journal currency, so the foreign currency
can equal the company currency when the journal is in a third one: a USD journal in a
EUR company with a EUR foreign amount. There `_get_reconcile_currency()` returns EUR,
so previously the `if currency != company_currency` gate was skipped, `currency_amount`
stayed `False`, and `currency_amount or amount` produced the correct company-currency
figure. With the `elif`, the first branch now runs and applies an implied rate to a
line that does not need one. #990's nesting does not have this path. Am I reading that
right?

**The rate's denominator may not be what you want.** `self.amount` is a `Monetary` on
`currency_id`, which core labels Journal Currency, while `balance` from
`_get_write_off_move_lines_dict` is in company currency. So
`balance * amount_currency / amount` is dimensionally right only when the journal
currency is the company currency, which is exactly the configuration your description
scopes the bug to and exactly what the test builds. For a CHF journal in a EUR company
the rate would be CHF per company applied to a EUR balance. Issue #993 calls that
case (a) and reports it as still broken. Would a guard on
`self.currency_id == self.company_currency_id` be worth adding, so the implied rate is
used only where `self.amount` really is a company-currency figure?

On #993 more generally: it puts the root cause in
`account_reconcile_abstract._get_reconcile_line`, where three `_convert` calls inside
`if max_amount:` use the date rate, and notes the same problem reaches the manual
counterpart path that neither this PR nor #990 touches. Fixing
`_reconcile_data_by_model` alone leaves that one. Not a reason to hold this, but worth
linking so the three threads stay together.

@OCA/accounting-maintainers could this and #990 get the no stale label while the
duplication is sorted out?

PR #989, opened 2026-05-05
[18.0][FIX] account_reconcile_model_oca: Use sudo to prevent access error to create res.partner.category

carlos-lopez-tecnativa:
When this test is executed with the `partner_category_security` module. This module disables access to create in `res.partner.category`, and without sudo, the test will fail with an access error. 
https://github.com/OCA/partner-contact/blob/418726c3539eda61ce444f02729494d9bc8cf706/partner_category_security/security/ir.model.access.csv#L4

Because this class inherits from `AccountTestInvoicingCommon`, the test is executed with the rights of a user who does not have access to res.partner.category. See https://github.com/odoo/odoo/blob/d1955028bb95eff8d33c1c2b1c211d8520bb33a2/addons/account/tests/common.py#L237

image

@Tecnativa @pedrobaeza could you please review this?

### DIFF
--- a/account_reconcile_model_oca/tests/test_reconciliation_match.py
+++ b/account_reconcile_model_oca/tests/test_reconciliation_match.py
@@ -717,12 +717,18 @@ class TestReconciliationMatchingRules(AccountTestInvoicingCommon):
 
     def test_matching_fields_match_partner_category_ids(self):
         self.rule_1.match_text_location_label = False
-        test_category = self.env["res.partner.category"].create(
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Yeah, no better solution to this conflict, but being totally limited to this test situation, it's OK to go with this one. If any other tests creates partner category records, we will face the same situation, but fortunately, this is not very frequent.

Fast-tracking for solving the integration issue.

/ocabot merge patch

PR #990, opened 2026-05-06
[16.0][FIX] account_reconcile_oca: Preserve bank line currency rate

aritzolea:
Termina de corregir https://github.com/OCA/account-reconcile/pull/792

En casos de que el extracto se cargue con otro tipo de conversión de monedas el proceso de conciliación no es correcto.

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -594,6 +594,10 @@ class AccountBankStatementLine(models.Model):
         partner = (
             reconcile_model._get_partner_from_mapping(self) or self._retrieve_partner()
         )
+        # Propagate the bank line implicit rate to keep the move consistent.
+        line_rate = None
+        if self.foreign_currency_id and self.amount and self.amount_currency:
+            line_rate = self.amount_currency / self.amount
         for line in reconcile_model._get_write_off_move_lines_dict(
             -liquidity_amount, partner.id
         ):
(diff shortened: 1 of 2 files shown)


ryanduguid (commented):
-

ryanduguid:
Following up on my own question about the forward ports: #987 already makes this
change on 17.0, opened two weeks before this one. I had not spotted it when I
commented. The two differ in shape. Yours keeps the existing
`if currency != company_currency` gate and guards `amount_currency` as well, while
#987 restructures to `if/elif` and has a test that deliberately sets an implied rate
away from the system rate. I have written up the comparison over there.

Rather than three more ports from here, it is probably worth the two of you settling
one version for 16.0 through 19.0.

One point from that comparison applies to this PR too. `self.amount` is a `Monetary`
on `currency_id`, which core labels Journal Currency, while the `balance` you multiply
is in company currency, so `line_rate` only carries the meaning you want when the
journal currency is the company currency. Issue #993 reports the other configuration,
a foreign-currency journal, as still broken.

pedrobaeza:
@ryanduguid please stop doing AI reviews too verbose and with dubious value.

ryanduguid:
@pedrobaeza wish granted, no more AI comments.

pedrobaeza:
You can use AI to assist you, but publish 10 reviews in one minute with very long paragraphs stating the obvious is not the right direction, supported by the AI Policy: https://github.com/pypa/pip/blob/main/AI_POLICY.md

PR #994, opened 2026-05-21
[18.0][IMP] account_reconcile_oca: defensive active_id in action contexts

dnplkndll:
## Problem

Four `ir.actions.act_window` records in `account_reconcile_oca` reference the bare name `active_id` in their `domain` and/or `context`:

- `action_bank_statement_line_reconcile`
- `action_bank_statement_line_reconcile_all`
- `action_bank_statement_line_reconcile_to_check`
- `action_bank_statement_line_move_view_reconcile`

When the Odoo 18 web client restores one of these via `WebClient.loadRouterState` (browser back/forward, bookmark, refresh on the reconcile route), the eval namespace does not include `active_id` — only the calling record provides it. The expression raises `NameError: name 'active_id' is not defined` and surfaces as an UncaughtPromiseError EvalError; the view fails to open.

```
EvalError: Can not evaluate python expression: ({'default_journal_id': active_id, ...})
Error: Name 'active_id' is not defined
  at evaluateExpr → makeContext → _preprocessAction → WebClient.loadRouterState
```

## Fix

Switch every reference to `context.get('active_id')` in both the `domain` and `context` attributes of the four affected actions. `context` is always in the Odoo expression eval namespace, and `.get()` returns `None` when the key is missing — `default_journal_id` becomes `None` (no default), the domain becomes `[('journal_id', '=', None)]`, and the kanban lands on its "Nothing to check" empty state until the user picks a journal.

## Test plan

- [x] Reproduced on a lab DB (`account_reconcile_oca@18.0.1.1.8`, ~115 bank journals): open Accounting → journal → *Reconcile*, then navigate to the action by URL without `active_id` → confirm modal pre-fix
- [x] Apply patch, `-u account_reconcile_oca` → repeat: confirm no modal, zero console errors, the empty state renders
- [x] Same flow against `action_bank_statement_line_reconcile_all` and `action_bank_statement_line_reconcile_to_check`
- [ ] Reviewer to confirm on their own deployment that the journal-record path (active_id from the calling record) still pre-fills correctly


account_reconcile_oca/__manifest__.py
```diff
@@ -5,7 +5,7 @@
     "name": "Account Reconcile Oca",
     "summary": """
         Reconcile addons for Odoo CE accounting""",
-    "version": "18.0.1.1.8",
+    "version": "18.0.1.1.9",
```
pedrobaeza: Please don't change manifest version, unless you provide migration scripts with the change (not talking about OpenUpgrade).

pedrobaeza (changes requested):
This is not fixing the underlying issue, as the context won't fail, but the link will be lost. Can you find a way to keep that information?

dnplkndll:
Thanks @pedrobaeza, both addressed:

**Version bump** — reverted to `18.0.1.1.8` (force-pushed). Apologies, that crept in from habit; OCA tooling handles version on merge.

**Preserving the journal link** — the `context.get('active_id')` pattern actually does preserve it in the from-record path. When the web client invokes an action from a record (button, kanban click, list row), the action evaluator builds the eval context with `active_id` placed in BOTH the top-level namespace AND the `context` dict. So `context.get('active_id')` returns the record ID in that path and `None` only when the action is restored via URL/history without a host record.

This is the same idiom Odoo core uses extensively for the analogous defensive pattern — e.g. `addons/sale/wizard/sale_order_discount.py:default=lambda self: self.env.context.get('active_id')`, `addons/base/models/ir_actions.py` (8 occurrences), `addons/base/wizard/base_language_install.py`, etc.

Happy to verify on a live instance if there's still doubt — the staging deploy of this branch shows: opening the action from a journal record's *Reconcile* button correctly defaults the kanban to that journal's lines; refreshing the URL (the previously-broken case) opens the kanban with no default and an empty journal selector that the user can populate manually. The link from-record is preserved.

If you'd prefer a different pattern (e.g. explicit `context.get('active_id') or context.get('default_journal_id')` to also support a `default_journal_id` URL param), happy to switch — let me know.

pedrobaeza:
Well, I don't think so. That context call will return empty, so for whatever `active_id` is used, no it won't work.

etobella:
On v18, tested on runboat, it is not happening....

Can you review it there?

dnplkndll:
this seems to be from using 16 style URLs I have generated that at not correct on 18. closing.

PR #1000, opened 2026-06-16
[18.0][FIX] account_reconcile_oca: keep candidates list page/scroll when selecting a counterpart

CarlosRoca13:
The candidates list is rendered inside the reconcile form. Selecting a counterpart updates the parent record, which re-renders the form and passes a new (but value-wise identical) domain to the list. RelationalModel resets the pagination offset to 0 whenever it is reloaded "from above with a domain", so the user was sent back to the first page of candidates (and the right panel scrolled away) on every click.

  - Add ReconcileMoveLineModel, which drops the domain from the reload params when its value hasn't changed, preserving the current page. A real search change still carries a different domain and resets the page as expected.
  - Wire the new model into ReconcileMoveLineView.
  - Preserve the scroll position of the reconcile panel in ReconcileMoveLineController.openRecord.

cc @Tecnativa TT63153

ping @sergio-teruel @carlosdauden

### DIFF
--- a/account_reconcile_oca/models/account_move_line.py
+++ b/account_reconcile_oca/models/account_move_line.py
@@ -3,10 +3,44 @@
 
 from odoo import _, models
 from odoo.exceptions import ValidationError
+from odoo.tools.sql import create_index
 
 
 class AccountMoveLine(models.Model):
     _inherit = "account.move.line"
+
+    def init(self):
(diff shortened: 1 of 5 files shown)


pedrobaeza (approved):
This reduces the number of calls as they are unneeded. The index is not acting always, but it's also welcomed.

etobella (approved):
It makes sense.

PR #1002, opened 2026-06-18
Draft PR [ADD] module account_reconcile_invoice_matching_extension

yweng8111:
This is a draft PR. Please Check this @mt-software-de 
With this extension we can just use the reconcile mode to do the reconciliation for skonto automatically.

Screenshot from 2026-07-22 13-54-24
Screenshot from 2026-07-22 11-08-06

### DIFF
--- a/account_reconcile_invoice_matching_extension/__init__.py
+++ b/account_reconcile_invoice_matching_extension/__init__.py
@@ -1,0 +1,2 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from . import models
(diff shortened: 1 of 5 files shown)


smaddlsoft:
@mt-software-de in case you find the time to pitch your alternative ideas, would be great.
@pedrobaeza maybe you have some thoughts as well as you may have some clients with Skonto/cash discount.

The PR here is not ideal as it requires manual creation of reconciliation models for each payment term, thats not ideal.
It is thus more meant as draft for discussion for a better solution.

PR #1004, opened 2026-06-22
[19.0][FIX] account_reconcile_oca: create partner bank with sudo in test

eduezerouali-tecnativa:
cc @Tecnativa

`account_payment_order` (in the bank-payment repo) overrides the base ACL by reusing the same external id.
Because it reuses `base.access_res_partner_bank_group_account_payment,` it replaces the base rule, so in any DB where `account_payment_order` is installed only `group_account_payment` can create `res.partner.bank`. The `account_reconcile_oca` test user has accounting/manager groups but not that payment group, so the nested `bank_ids` create fails

ping @pedrobaeza @etobella


pedrobaeza (approved):
This is for being compatible with `account_payment_order`, while we fix the underlying problem, which is to change the permissions of bank account creation

/ocabot merge patch

PR #1005, opened 2026-06-22
[19.0][FIX] account_reconcile_oca: test setup for res.partner.bank permissions

Andrii9090-tecnativa:
The test creates partner bank accounts, but the group allowed to do so can very depending on installed modules that extend or update the base ACL. Instead of hardcoding a specific group or bypassing access rights, the test now grants the user the group configured on the actual res.partner.bank access rule.
For example, when `account_payment_order` is installed, it reassigns the existing `base.access_res_partner_bank_group_partner_manager` ACL to its own “Accounting / Payments” group. 
@pedrobaeza ping

@Tecnativa

### DIFF
--- a/account_reconcile_oca/tests/test_bank_account_reconcile.py
+++ b/account_reconcile_oca/tests/test_bank_account_reconcile.py
@@ -17,6 +17,12 @@ class TestAccountReconciliationCommon(TestAccountReconciliationModelCommon):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        bank_account_access = cls.env.ref(
+            "base.access_res_partner_bank_group_partner_manager"
+        )
+        cls.env.user.write(
+            {"group_ids": [Command.link(bank_account_access.group_id.id)]}
+        )
         cls.env = cls.env(context=cls._setup_context())
         # Auto-disable reconciliation model created automatically with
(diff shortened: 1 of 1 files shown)


account_reconcile_oca/tests/test_bank_account_reconcile.py
```diff
@@ -17,6 +17,12 @@ def _setup_context(cls):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        bank_account_access = cls.env.ref(
```
pedrobaeza, line 20: Why putting the ref of an ACL instead of the group?

Andrii9090-tecnativa: Because, in the case of `account_payment_order`, if the group changes, I use the ACL and select the current group from the ACL. This guarantees that we always have the actual/current group.

Traceback:
```
odoo.exceptions.AccessError: You are not allowed to create 'Bank Accounts' (res.partner.bank) records.
This operation is allowed for the following groups:
	- Accounting / Payments
Contact your administrator to request access if necessary.

```

pedrobaeza:
I think we are going to use the approach of #1004 using `sudo`, and it has been proposed before.

PR #1007, opened 2026-06-29
[16.0][FIX] account_reconcile_oca: keep statement rate for manual counterparts

### LINKED ISSUE #993: [16.0][BUG] account_reconcile_oca: incorrect amount when manually adding a counterpart line in company currency from a foreign-currency statement (late rate import)
## Module
account_reconcile_oca

## Describe the bug
When a bank statement line is in a foreign-currency journal (e.g. a CHF cash journal in a EUR company) and the user manually adds a counterpart `account.move.line` that is in the company currency (e.g. an EUR supplier invoice), the proposed reconciled amount is computed via `_convert` at `self.date`, using whatever currency rate is **currently** in the database for that date. If a new rate for the same date is imported **after** the statement was recorded, this rate differs from the one used to actually book the statement, and the widget proposes an amount that no longer matches the booked statement amount, producing an artificial residual on the suspense account and preventing reconciliation.

This happens whenever the reconcile currency is foreign and the counterpart is in the company currency, in two configurations:
  - (a) the journal is in a foreign currency (e.g. a CHF cash journal in a EUR company);
  - (b) the journal is in the company currency but the line carries a foreign foreign_currency_id.

A fix must branch on the reconcile currency (_get_reconcile_currency() → foreign_currency_id or journal currency), not the journal currency, otherwise case (b) is missed.

  The root cause is in `account_reconcile_abstract._get_reconcile_line` (lines 70–85 in 16.0; same code in 17.0, 18.0, 19.0). Both `_convert` calls inside the `if max_amount: ...` block use the date-based rate, which is unreliable when rates are (re-)imported after the statement is booked. This pattern was partially addressed by commit d873775 for `_reconcile_data_by_model`, but the manual click path (`_add_account_move_line` → `_get_reconcile_line`) and several other call sites are not covered.

## To Reproduce
Verified on 16.0. The same code pattern appears in 17.0/18.0/19.0 (not independently tested there).

1. Set EUR as the company currency, activate CHF
2. Add a CHF rate: `1 CHF = 1.07 EUR`, dated 17/12/2025 (the only CHF rate available at that point)
3. Create a cash journal in CHF
4. Create a supplier invoice in EUR for 1,261.31
5. Create a cash withdrawal statement line on the CHF journal: 250 CHF, dated 13/01/2026 → booked at 267.50 EUR (= 250 × 1.07, using the only rate available at that date)
6. Later, add a CHF rate: `1 CHF = 1.09004 EUR`, dated 13/01/2026 (e.g. via daily ECB import or a manual correction)
7. Open the bank reconciliation widget on the cash withdrawal statement line
8. Click on the supplier invoice's payable line to add it as counterpart

Observed: the payable line is imputed at **272.51 EUR** (= 250 × 1.09004, the newly imported rate), leaving an artificial **5.01 EUR** residual on the suspense account. `can_reconcile` is `False`.

Image

Variant (b) (company-currency journal + foreign_currency_id): same outcome. A EUR-journal line booked at 267.50 EUR with amount_currency = 250 CHF is also imputed at 272.51 after a late rate import. Not caught by a journal-currency guard, only by a reconcile-currency one.

**Expected behavior**
The payable line should be imputed at **267.50 EUR**, the amount actually booked when the statement was recorded and the reconciliation should be balanced with no residual on the suspense account.

Image

The booked statement is the authoritative source for the rate: the cash account has already been credited by 267.50 EUR, so the counterpart line must be debited by the same amount for the move to remain balanced and consistent with the historical booking. Odoo's standard helper `_prepare_counterpart_amounts_using_st_line_rate` (defined on `account.bank.statement.line` in core Community 16) implements exactly this logic`.

**Additional context**
This bug surfaces in any reconciliation flow that pairs a foreign-currency bank statement line with a counterpart line in the company currency, when a rate update happens between the booking and the reconciliation. Concrete patterns I observed on an install (EUR company with a customer using a CHF cash journal):
  - Cash CHF outflow paying a supplier invoice in EUR (the scenario above)
  - Cash CHF inflow refunding/settling a customer invoice in EUR
  - Bank CHF transfer settling any EUR journal entry on a partner account

This can occur on companies that import rates daily because any statement booked at a date for which the rate is later updated is at risk.

Related code paths in `account_reconcile_oca` (all using `_get_reconcile_line(is_counterpart=True, max_amount=...)`) that i think exhibit the same root cause:
  - `account_bank_statement_line._add_account_move_line` (manual click -> the one reproduced above)
  - `account_bank_statement_line._reconcile_data_by_model` (reconcile model match -> partially fixed by d873775 only when `line.currency_id == reconcile_currency`)
  - account_bank_statement_line.create (auto-match at creation). It passes amount_currency as max_amount (already in reconcile currency), so it is covered by the abstract hook alone, without a dedicated max_amount conversion.
  - `account_bank_statement_line.button_manual_reference_full_paid` ("mark as fully paid" button)
  - `account_account_reconcile._recompute_data` (manual partner reconciliation -> additional issue: max_amount is accumulated in company currency, while _get_reconcile_line seems to expect it in reconcile currency

Related PRs that address adjacent currency-rate issues:
  - d873775 (merged on 16.0): pre-converts `max_amount` using the statement rate, but only when `line.currency_id == reconcile_currency`
  - #987 (open on 17.0): fix incorrect amounts when reconciling with a model and a foreign currency
  - #990 (open on 16.0): preserve the bank line currency rate on writeoff lines

I'm working on a PR for 16.0 that replaces the date-based _convert calls in the manual-add path (and related call sites) by _prepare_counterpart_amounts_using_st_line_rate, the Odoo Helper for the exact same purpose. This helper derives the rate from the booked statement move (balance / amount_currency of the liquidity line) instead of querying res.currency.rate at self.date, so it is immune to late rate imports. The default behavior of _get_reconcile_line is preserved (the abstract is only modified to delegate the rate-dependent calculation to an overridable hook, whose default implementation reproduces the current _convert based code). Separately, account_account_reconcile._recompute_data is fixed to accumulate the cap in currency_amount instead of company amount, which intentionally changes the multi-currency partner-reconciliation result.

jmtdev0:
Fixes #993.

## Description

When a bank statement line is created in a foreign-currency journal before a late currency rate update, the reconciliation widget must keep using the rate stored on the statement line.

Before this fix, manually adding a counterpart line could recompute the counterpart amount using the updated global currency rate instead of the original statement line rate.

This PR keeps the default conversion behavior in the abstract reconciliation model, but lets `account.bank.statement.line` convert limited counterpart amounts through Odoo's statement-line-rate helper.

## Changes

- Add `_convert_reconcile_counterpart_max_amount()` as a small hook in `account.reconcile.abstract`.
- Override that hook in `account.bank.statement.line`.
- Use `_prepare_counterpart_amounts_using_st_line_rate()` for company/reconcile currency amounts on bank statement lines.
- Keep the previous generic conversion behavior for other currency cases.
- Add regression tests for both payable and receivable manual counterpart flows.

## Chronology

1. Reproduced the issue with a CHF bank statement line created with an original rate of `1.07`.
2. Updated the same day's rate afterwards to `1.09004`, simulating a late rate import.
3. Confirmed the bug: when adding a payable counterpart manually, the widget proposed `272.51` instead of keeping the statement-line amount `267.5`.
4. Traced the recomputation to the generic `max_amount` conversion path in `account.reconcile.abstract`.
5. Added a hook to preserve the generic behavior by default.
6. Specialized the bank statement line implementation to reuse Odoo's existing statement-line-rate helper.
7. Added a symmetric receivable test to cover the positive-amount sign case as well.

## Testing

Before the fix, the payable regression test failed with:

```text
AssertionError: 272.51 != 267.5
```

After the fix:

```text
account_reconcile_oca: 40 tests
0 failed, 0 errors
```

Also run successfully:

```text
pre-commit run --files \
  account_reconcile_oca/models/account_reconcile_abstract.py \
  account_reconcile_oca/models/account_bank_statement_line.py \
  account_reconcile_oca/tests/test_bank_account_reconcile.py
```

All selected pre-commit hooks passed, including black, flake8, and pylint-odoo.

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -1184,6 +1184,19 @@ class AccountBankStatementLine(models.Model):
         self.ensure_one()
         self.move_id.to_check = False
 
+    def _convert_reconcile_counterpart_max_amount(self, max_amount, currency, date):
+        reconcile_currency = self._get_reconcile_currency()
+        if currency not in (self.company_id.currency_id, reconcile_currency):
+            return super()._convert_reconcile_counterpart_max_amount(
+                max_amount, currency, date
+            )
+        amounts = self._prepare_counterpart_amounts_using_st_line_rate(
+            reconcile_currency, 0.0, max_amount
+        )
+        if currency == self.company_id.currency_id:
+            return amounts["balance"]
+        return amounts["amount_currency"]
+
     def _get_reconcile_line(
         self,
         line,
--- a/account_reconcile_oca/models/account_reconcile_abstract.py
+++ b/account_reconcile_oca/models/account_reconcile_abstract.py
@@ -36,6 +36,11 @@ class AccountReconcileAbstract(models.AbstractModel):
     def _get_reconcile_currency(self):
         return self.currency_id or self.company_id._currency_id
 
+    def _convert_reconcile_counterpart_max_amount(self, max_amount, currency, date):
+        return self._get_reconcile_currency()._convert(
+            max_amount, currency, self.company_id, date
+        )
+
     def _get_reconcile_line(
         self,
         line,
@@ -71,8 +76,10 @@ class AccountReconcileAbstract(models.AbstractModel):
                     -real_currency_amount > max_amount > 0
                     or -real_currency_amount < max_amount < 0
                 ):
-                    currency_max_amount = self._get_reconcile_currency()._convert(
-                        max_amount, currency, self.company_id, date
+                    currency_max_amount = (
+                        self._convert_reconcile_counterpart_max_amount(
+                            max_amount, currency, date
+                        )
                     )
                     amount = currency_max_amount
                     net_amount = -max_amount
--- a/account_reconcile_oca/tests/test_bank_account_reconcile.py
+++ b/account_reconcile_oca/tests/test_bank_account_reconcile.py
@@ -1464,6 +1464,140 @@ class TestReconciliationWidget(TestAccountReconciliationCommon):
             self.assertTrue(f.can_reconcile)
             self.assertEqual(f.reconcile_data_info["data"][-1]["amount"], 3.63)
 
+    def test_manual_counterpart_uses_statement_rate_after_late_rate_update(self):
+        chf = self.env.ref("base.CHF")
+        chf.active = True
+        rate = self.env["res.currency.rate"].create(
+            {
+                "currency_id": chf.id,
+                "name": time.strftime("%Y-01-13"),
+                "inverse_company_rate": 1.07,
+            }
+        )
+        chf_journal = self.env["account.journal"].create(
+            {
+                "name": "Bank CHF",
+                "type": "bank",
+                "code": "BCHF",
+                "currency_id": chf.id,
+                "suspense_account_id": self.company.account_journal_suspense_account_id.id,
+            }
+        )
+        invoice = self._create_invoice(
+            move_type="in_invoice",
+            currency_id=self.currency_euro_id,
+            invoice_amount=1261.31,
+            date_invoice=time.strftime("%Y-02-01"),
+            auto_validate=True,
+        )
+        bank_stmt = self.acc_bank_stmt_model.create(
+            {
+                "journal_id": chf_journal.id,
+                "date": time.strftime("%Y-01-13"),
+                "name": "test",
+            }
+        )
+        bank_stmt_line = self.acc_bank_stmt_line_model.create(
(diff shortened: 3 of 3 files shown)


jmtdev0:
A broader alternative/follow-up PR has been opened in #1008 by the issue author. It appears to cover the same single-counterpart manual case fixed here plus additional call sites mentioned in #993. I am happy to keep this PR as the smaller focused option, or close it if maintainers prefer the broader approach.

Personally, after comparing both approaches, I think #1008 is stronger than #1007 because it addresses the broader set of call sites described in #993.

PR #1010, opened 2026-07-02
[18.0[FIX] account_reconcile_sale_order: don't subscribe sales order, picking

jans23:
Until now the user who reconciles subscribes the sale order (and subsequent delivery order). This is not desired for accountants responsible for reconciliation. This fix prevents such subscription.

### DIFF
--- a/account_reconcile_sale_order/models/account_bank_statement_line.py
+++ b/account_reconcile_sale_order/models/account_bank_statement_line.py
@@ -121,7 +121,18 @@ class AccountBankStatementLine(models.Model):
         Invoice selected sales orders and post the invoices
         """
         if order.state in ("draft", "sent"):
+            order = order.with_context(mail_create_nosubscribe=True)
+            has_pickings = "picking_ids" in order._fields
+            # Track existing pickings to only touch the ones we create
+            pickings_before = (
+                order.picking_ids if has_pickings else self.env["sale.order"].browse()
+            )
             order.action_confirm()
+            # Don't subscribe the reconciling user to the new delivery order(s)
+            if has_pickings:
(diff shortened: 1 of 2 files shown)


account_reconcile_sale_order/models/account_bank_statement_line.py
```diff
@@ -121,7 +121,7 @@ def _prepare_reconcile_line_data_sale_order_invoice(self, order):
         Invoice selected sales orders and post the invoices
         """
         if order.state in ("draft", "sent"):
-            order.action_confirm()
+            order.with_context(tracking_disable=True).action_confirm()
```
hbrunn: this will prevent tracking entirely, so you won't see which user did the confirmation, which is undesirable. set key `mail_create_nosubscribe` instead, this should do what you want and only that

hbrunn (changes requested):

account_reconcile_sale_order/models/account_bank_statement_line.py
jans23: Thank you. Like this?

hbrunn: exactly. Please remove the test part looking at pickings, that's dead code from the point of view of this repo because it doesn't install sale_stock

jans23:
Done.

hbrunn (approved):

jans23:
@hbrunn This works for sale orders but the user will still be subscribed to delivery orders. If you want to keep `mail_create_nosubscribe` I could add an explicit `message_unsubscribe` on pickings. Alternatively I would use `order.with_context(mail_notrack=True).action_confirm()` instead which seems to be simpler to me. What do you think?

PR #1012, opened 2026-07-06
[19.0][MIG] account_reconcile_model_oca: Migration to 19.0

mono1982:
## Migration to 19.0

Standard migration procedure was followed: module history transplanted from 18.0 via `git format-patch`/`git am`, followed by a single `[MIG]` adaptation commit.

### What changed in Odoo 19 and how this PR deals with it

Odoo 19.0 removed most of the reconciliation-model matching fields from the `account` core module (`rule_type`, `matching_order`, payment tolerance, text/note/reference matching options, partner mappings, `past_months_limit`, `decimal_separator`, ...). Since the whole purpose of this module is to carry the reconciliation logic that left Community, those fields are now reintroduced here (source: Odoo 18.0 Community, LGPL-3):

- Reintroduce removed fields, computes and constraints on `account.reconcile.model` and `account.reconcile.model.line`
- Reintroduce the `account.reconcile.model.partner.mapping` model incl. access rights
- Map `auto_reconcile` onto the new core `trigger` field (compute/inverse, single source of truth)
- Rebuild the form view extension: the 19.0 core form no longer contains the matching UI, so the module now provides it
- Adapt to the 19.0 ORM API: `_where_calc()` removed → `_search(..., bypass_access=True)` (keeps the pre-19 no-ir.rule behaviour), SQL clause tuple-unpacking → `.code`/`.params`, `self._cr` → `self.env.cr`
- Tests: fiscal position tax mapping adapted (`tax_src_id`/`tax_dest_id` removed in 19.0 → `original_tax_ids` on the destination tax)

All 35 module tests pass on 19.0.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pedrobaeza:
The module is now integrated into `account_reconcile_oca`. And please don't use blindly the AI.

PR #1017, opened 2026-08-04
[18.0][FIX] account_mass_reconcile: reconciles not being sorted

astirpe:
Porting of https://github.com/OCA/account-reconcile/pull/1016 to 18.0

Original Code:

New code:

### DIFF
--- a/account_mass_reconcile/models/mass_reconcile.py
+++ b/account_mass_reconcile/models/mass_reconcile.py
@@ -328,7 +328,7 @@ class AccountMassReconcile(models.Model):
         if run_all:
             reconciles.run_reconcile()
             return True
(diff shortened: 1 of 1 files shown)


ryanduguid (commented):
Thanks for finishing the port.

Same check as on #1016: #651 merged on 16.0 in January 2025 and that branch reads
`reconciles = reconciles.sorted(key=_get_date)` at line 304. 18.0 was still discarding
the sorted result at line 331, which is what this fixes, and 17.0 at line 302 is the
other half in #1016. `account_mass_reconcile` does not exist on 19.0, so between the
two PRs the port is complete.

This is not showing a CI result either. Both may need a rebase to trigger the build
before anyone can merge them.

PR #1020, opened 2026-08-17
[18.0][FIX] account_statement_base: Remove overrides of the actions

pedrobaeza:
Odoo already includes the form view in the actions:

https://github.com/odoo/odoo/commit/d5c5bffd360b1feab59d02d04e9da1dbe22d272c
https://github.com/odoo/odoo/commit/3e61fd28720e51fac467d1cc7a038933519ef295

so no need of them.

This conclusion comes after detecting the problem in the credit card journals, but seeing that it has been solved upstream just today, and no need of them.

@Tecnativa TT64224

### DIFF
--- a/account_statement_base/views/account_bank_statement.xml
+++ b/account_statement_base/views/account_bank_statement.xml
@@ -87,15 +87,6 @@
         </field>
     </record>
 
-    <!-- Add form to view mode. -->
-    <record id="account.action_bank_statement_tree" model="ir.actions.act_window">
-        <field name="res_model">account.bank.statement</field>
-        <field name="view_mode">list,form,pivot,graph</field>
-    </record>
-    <record id="account.action_view_bank_statement_tree" model="ir.actions.act_window">
-        <field name="view_mode">list,form,pivot,graph</field>
-    </record>
-
     <record id="view_bank_statement_tree" model="ir.ui.view">
         <field name="name">account.bank.statement.list</field>
         <field name="model">account.bank.statement</field>


christian-ramos-tecnativa (approved):
Functional and code review ok!

carlos-lopez-tecnativa (changes requested):
LGTM from the `Transactions` button in the dashboard. However, if you click `Cash Register` and create a new record, the default journal is different from the one you clicked on.

I think you should set the default journal here as well. What do you think?
Image

pedrobaeza:
I didn't check that flow, because it was not the reported one, but let's try to fix it as well.

victoralmau:
Maybe I haven't fully understood the problem, but right now, without this change, there's no problem accessing a journal from the “Transactions” smart button.

ejemplo

When creating a new bank statement and line items, the journal field that is set is the correct one.

pedrobaeza:
@victoralmau you can check the results of the before and the after of the patch on the still non patched route that Carlos commented (I'm investigating now that): clicking on the button works and put the correct journal, but clicking on the dropdown element not. Before the patch, you have the same problem in the button.

Peek 18-08-2026 11-30

pedrobaeza:
@carlos-lopez-tecnativa @victoralmau the dropdown action also respects the journal now.

carlos-lopez-tecnativa (approved):
LGTM

pedrobaeza:
At the end, the patch can be reduced to simply add the form on the CC action. The rest was contaminated to incorrect results while debugging.

carlos-lopez-tecnativa (commented):
You dropped everything related to the context. Are the changes correct?
Image

victoralmau:
Can you explain in detail (and include it as a comment in the code) why it's necessary to redefine `view_mode` for that act_window if `account` already has it defined? https://github.com/odoo/odoo/blob/014d58e3204d17db6dcba3c8ab7d8ad35003300e/addons/account/views/account_bank_statement_views.xml#L74

Additional comments (although I know you don't agree with them):
- The `journal_id` field from the statements should be displayed in the form view (this would eliminate the need to display and select the journal_id from the lines)
- The `journal_id` field in the statements should be a computed store readonly=False field; if there are no lines, the value can be set manually; otherwise, it will be the value of the lines. This is the expected behavior for statements when creating a new one (with no lines yet) from a form view, since that’s what happened in previous versions (v16, for example).

pedrobaeza:
@victoralmau that's because Odoo has just fixed it today:

https://github.com/odoo/odoo/commit/3e61fd28720e51fac467d1cc7a038933519ef295

In local, I was working with an outdated Odoo version, so that has confused me even more. My fault on that.

Then, there's no finally need for any of the overrides, so removing them. What a roller coaster!

Let's finally close this merging the removals. About showing the journal at header level and make it stored, I don't think it's correct as stated.

/ocabot merge patch

PR #1022, opened 2026-08-24
[18.0][FIX] account_reconcile_oca: do not spread the raw field context

CarlosRoca13:
`this.props.context` is the `context` attribute of the `add_account_move_line_id` field as written in the arch, unevaluated: a string.

Nothing is lost by dropping it. `getFieldContext()` reads that very same attribute from `record.activeFields[name].context` and returns it evaluated, merged with the context the parent record carries, so `list_view_ref`, `search_view_ref` and the `search_default_partner_id` pre-filter keep working.
See: https://github.com/odoo/odoo/blob/a77d1d69d0319d1fab7285187cf40936d6438946/addons/web/static/src/model/relational_model/utils.js#L296

The prop is unused afterwards, as the template only renders `listViewProperties`, so drop it from the component props and from `extractProps` too instead of leaving a dead prop around.

cc @Tecnativa

ping @pedrobaeza @victoralmau

### DIFF
--- a/account_reconcile_oca/static/src/js/widgets/reconcile_move_line_widget.esm.js
+++ b/account_reconcile_oca/static/src/js/widgets/reconcile_move_line_widget.esm.js
@@ -37,10 +37,7 @@ export class AccountReconcileMatchWidget extends Component {
             resModel: this.props.record.fields[this.props.name].relation,
             searchMenuTypes: ["filter"],
             domain: this.getDomain(),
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

etobella (approved):
It makes sense

pedrobaeza:
@CarlosRoca13 please fw-port it to 19

PR #1023, opened 2026-08-25
[18.0][IMP] account_reconcile_oca: avoid useless partner retrievals

CarlosRoca13:
_retrieve_partner() is expensive: it searches by bank account and by partner name, walks every reconcile model with its mappings and, as a last resort, runs a regexp_matches query over account_move_line. It was being called in situations where its result should not be used.

cc @Tecnativa TT64256

ping @pedrobaeza @victoralmau

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -660,19 +660,22 @@ class AccountBankStatementLine(models.Model):
             )
             data += lines
         if not from_unreconcile:
+            reconcile_models = self.env["account.reconcile.model"].search(
+                [
+                    (
+                        "rule_type",
+                        "in",
+                        ["invoice_matching", "writeoff_suggestion"],
+                    ),
+                    ("company_id", "=", self.company_id.id),
+                ]
+            )
+            # Retrieving the partner is expensive, so it is only done when
+            # there is at least one model that could use it
             res = (
-                self.env["account.reconcile.model"]
-                .search(
-                    [
-                        (
-                            "rule_type",
-                            "in",
-                            ["invoice_matching", "writeoff_suggestion"],
-                        ),
-                        ("company_id", "=", self.company_id.id),
-                    ]
-                )
-                ._apply_rules(self, self._retrieve_partner())
+                reconcile_models._apply_rules(self, self._retrieve_partner())
+                if reconcile_models
+                else {}
             )
             if res and res.get("status", "") == "write_off":
                 return self._recompute_suspense_line(
@@ -971,6 +974,18 @@ class AccountBankStatementLine(models.Model):
             vals["amount_currency"] = line["currency_amount"]
         return vals
 
+    def write(self, vals):
+        if "partner_id" in vals and not self.filtered(
+            lambda line: line.partner_id.id != (vals["partner_id"] or False)
+        ):
+            # The lines already have this partner. Keeping the key would make
+            # _synchronize_to_moves() recompute reconcile_data_info, which
+            # retrieves the partner again for nothing.
+            vals = {key: value for key, value in vals.items() if key != "partner_id"}
+            if not vals:
+                return True
+        return super().write(vals)
+
     @api.model_create_multi
     def create(self, mvals):
         result = super().create(mvals)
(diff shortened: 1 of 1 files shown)


etobella (approved):
Thanks!

pedrobaeza (commented):
Thanks for the improvements. I see that the partner is very coupled to the `_ apply_rules` method, so we can decouple the call with or without the partner according some things like `match_partner_ids` field existence, but there's one thing that is draining the performance: the partner is not written back to the bank statement line once retrieved, so each time you click on the element, all the algorithm is performed again. Can you write back the partner if found in the absl in `_retrieve_partner`? Or do you see any drawback, @etobella ?

etobella:
It makes sense to me.

CarlosRoca13:
Good catch on the recomputation cost, but I'd rather not write it into `partner_id` on the statement line.

Writing `partner_id` on `account.bank.statement.line` triggers `_synchronize_to_moves()` (https://github.com/odoo/odoo/blob/4f6f490b1a2d9d025364205da82bdbd152aa6937/addons/account/models/account_bank_statement_line.py#L810), which propagates the partner to `move_id` and its journal items, and on the way rewrites the liquidity/suspense lines and drops `other_lines` with (2, line.id) (https://github.com/odoo/odoo/blob/4f6f490b1a2d9d025364205da82bdbd152aa6937/addons/account/models/account_bank_statement_line.py#L830)

So we'd be persisting a heuristic guess (`ilike` on partner name, `regexp_matches` over `account_move_line`) as if the user had entered it, mutating the journal entry just by opening the view. And it would be frozen from then on: `_retrieve_partner` short-circuits on `if self.partner_id` (https://github.com/OCA/account-reconcile/blob/580d32911fcb4bb9a9ccab71e153502670a90475/account_reconcile_model_oca/models/account_bank_statement_line.py#L17-L18), so if a partner mapping or a bank account is configured later, that line will never be re-evaluated.

Would you be OK with caching the result in a dedicated technical field on the statement line (something like `retrieved_partner_id`, stored, not synced to the move), so `_retrieve_partner` reads from it instead of re-running the whole algorithm?

pedrobaeza:
Thanks for the analysis, @CarlosRoca13. I have some doubts about how this works before being able to judge: isn't at the end the goal of retrieving the partner to have that partner in the journal entry? Isn't it transferred in any moment? If so, how and if there's any shortcut for writing this data? If not transferred, I see a hole there.

CarlosRoca13:
I believe leaving `st_line.partner_id` untouched is deliberate rather than a hole, because once that field is set it becomes authoritative and changes behaviour beyond mere caching: `_recompute_suspense_line()` then stops using the suspense account and proposes the partner's receivable/payable one, setting `can_reconcile = True` (https://github.com/OCA/account-reconcile/blob/f06a6581a57f5d667123903b1ce14ccc58ccc063/account_reconcile_oca/models/account_bank_statement_line.py#L297-L306), a behaviour explicitly documented by the test at test_bank_account_reconcile.py https://github.com/OCA/account-reconcile/blob/46f971521618d8b8743eb51b329d7f57c1588b23/account_reconcile_oca/tests/test_bank_account_reconcile.py#L986. 

So persisting a guess coming from an `ilike` on the partner name or from a `regexp_matches` over `account_move_line` would silently turn unmatched lines into "ready to reconcile against this customer's receivable" without anybody reviewing it. On top of that, it would freeze the guess, since `_retrieve_partner()` returns early on `if self.partner_id` (https://github.com/OCA/account-reconcile/blob/580d32911fcb4bb9a9ccab71e153502670a90475/account_reconcile_model_oca/models/account_bank_statement_line.py#L17-L18), so configuring a partner mapping or a bank account later would never re-evaluate that line. That's why `partner_id` keeps meaning "what the user or the import stated", and it's only written when the user explicitly sets it on the liquidity line via `manual_partner_id` → `_update_move_partner()` (https://github.com/OCA/account-reconcile/blob/9e9c84779f04551031cb10c01c0f25d46108260c/account_reconcile_oca/models/account_bank_statement_line.py#L547).

pedrobaeza:
So the answer to my question is that is not written back that retrieved partner in any moment to the account entry even if reconciled, isn't it? I consider that behavior incorrect and should be written.

PR #1024, opened 2026-08-25
[19.0][FIX] account_reconcile_oca: do not spread the raw field context

CarlosRoca13:
`this.props.context` is the `context` attribute of the `add_account_move_line_id` field as written in the arch, unevaluated: a string.

Nothing is lost by dropping it. `getFieldContext()` reads that very same attribute from `record.activeFields[name].context` and returns it evaluated, merged with the context the parent record carries, so `list_view_ref`, `search_view_ref` and the `search_default_partner_id` pre-filter keep working.

The prop is unused afterwards, as the template only renders `listViewProperties`, so drop it from the component props and from `extractProps` too instead of leaving a dead prop around.

cc @Tecnativa

FW-PORT of https://github.com/OCA/account-reconcile/pull/1022

ping @pedrobaeza @adasatorres-tecnativa


pedrobaeza (approved):
Tested and it doesn't affect the normal operation.

adasatorres-tecnativa (approved):
LGTM, Tested on runboat.

PR #1027, opened 2026-08-27
[IMP] account_reconcile_oca: Make it more compatible with mobile

etobella:
The change applied make everything more easy to view on a mobile:

- [x] Buttons on header do not overlap later on.
- [x] On mobile, kanban view is hidden. We can open it with a chevron in the left, making it easier to handle
- [x] Make table more mobile prepared

image
image


etobella:
Table of reconcile has been improved too:

image

PR #1028, opened 2026-08-28
[19.0][FIX] account_reconcile_oca: keep candidates list page/scroll when selecting a counterpart

CarlosRoca13:
cc @Tecnativa TT63153

FWPORT of https://github.com/OCA/account-reconcile/pull/1000

ping @pedrobaeza @victoralmau

### DIFF
--- a/account_reconcile_oca/models/account_move_line.py
+++ b/account_reconcile_oca/models/account_move_line.py
@@ -3,10 +3,44 @@
 
 from odoo import models
 from odoo.exceptions import ValidationError
+from odoo.tools.sql import create_index
 
 
 class AccountMoveLine(models.Model):
     _inherit = "account.move.line"
+
+    def init(self):
+        """Index the candidate counterparts of the reconcile widget.
+
+        The `add_account_move_line_id` field of the reconcile view searches the
+        counterparts with the domain::
+
+            [("parent_state", "=", "posted"), ("amount_residual", "!=", 0),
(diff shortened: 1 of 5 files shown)


etobella (approved):

pedrobaeza:
The number of RPC calls have been reduced for the same result.

This is before:

imagen

And this is after:

imagen

However, there are two more extra calls that weren't in v18:

imagen

`get_views` is called twice each time the item at the left is changed, while I think it should be only called once when the reconciliation view is opened (as it should return all the times the same view).

This is not something of your PR, but of the migration of the module to 19. Can you please check and resolve it if possible? cc @etobella

CarlosRoca13:
Thanks for checking, @pedrobaeza. I think it's just HTTP: in 19.0 the `get_views` cache moved to the new `RPCCache`, which is only installed when `window.isSecureContext && session.browser_cache_secret` (https://github.com/odoo/odoo/blob/ec5ca8791de1945c3e3016aeddbf41420ded1688/addons/web/static/src/start.js#L35-L37) , in 18.0 `view_service.js` had its own in-memory cache that always worked.

Can you try it in localhost? it is a secure context, and it should not happen there

pedrobaeza (approved):
OK, @CarlosRoca13 we can test it also on runboat just putting https:// and accepting the warning, and indeed those calls are not done.

Merging.

/ocabot merge patch

PR #1032, opened 2026-09-01
[18.0][IMP] account_reconcile_oca: select the next line after reconciling

CarlosRoca13:
When a line was reconciled, the kanban was reloaded and the selection always fell back to the first pending line, forcing the user to look for where they were. Now the selection moves to the line right after the reconciled one, and to the last pending line when there is none after it.

cc @Tecnativa TT64256

ping @pedrobaeza @victoralmau


pedrobaeza (approved):
Tested on runboat and working flawlessly

victoralmau (approved):
Functional review OK

pedrobaeza:
@etobella, as I suppose you agree, I'm merging this:

/ocabot merge patch

PR #1033, opened 2026-09-02
[19.0] [IMP] account_reconcile_oca: Improve security

etobella:
I detected that we were not passing the security rules, so I added an option to do that.


etobella:
After a carefull review, there is a lot of reviews on the selects to avoid security issues, however, I think that an extra check is not wrong.

PR #1037, opened 2026-09-08
[19.0][FIX] account_reconcile_oca: Typo in invisible attribute

### LINKED ISSUE #1035: [17.0][18.0][19.0] account_reconcile_oca: "insible" typo hides nothing — the invoice hint is always visible
## Module

`account_reconcile_oca` (checked on 19.0 at 19.0.1.0.8, present on 17.0 and 18.0 as well; **16.0 is not affected**).

## Problem

In `views/account_bank_statement_line.xml`, two elements of the manual-operation panel use the attribute `insible` instead of `invisible`:

https://github.com/OCA/account-reconcile/blob/19.0/account_reconcile_oca/views/account_bank_statement_line.xml#L306
https://github.com/OCA/account-reconcile/blob/19.0/account_reconcile_oca/views/account_bank_statement_line.xml#L309

```xml
<label
    for="manual_move_id"
    string=""
    insible="manual_move_type not in ['in_invoice', 'in_refund', 'out_invoice', 'out_refund'] or manual_original_amount == 0 or is_reconciled"
/>
<div
    insible="manual_move_type not in ['in_invoice', 'in_refund', 'out_invoice', 'out_refund'] or manual_original_amount == 0 or is_reconciled"
>
    Invoice <field class="oe_inline" name="manual_move_id" />
    with an open amount ...
```

`insible` is not a known attribute, so it is silently ignored and the condition never applies. The `Invoice ... with an open amount ...` hint is therefore rendered for **every** manual counterpart — including lines that are not invoices, lines whose original amount is zero, and lines already reconciled — where the intent was clearly to hide it.

It fails silently: no warning at install time, nothing in the log.

## Steps to reproduce

1. Open the reconciliation view of a bank journal.
2. Select a statement line and go to the *Manual operation* tab.
3. Pick a counterpart account that is not an invoice (e.g. a bank-fee expense account).
4. The `Invoice … with an open amount …` block is shown, although `manual_move_type` is not an invoice type.

## Proposed fix

Rename the attribute in both places:

```diff
-                                    insible="manual_move_type not in [...]"
+                                    invisible="manual_move_type not in [...]"
```

Happy to open the PR if that helps.

## Context

Found while running `account_reconcile_oca` on Odoo 19 Community, where this module provides the reconciliation screen the Community edition does not ship. We are carrying the one-word fix as a local patch and would rather drop it.

pedrobaeza:
Closes #1035

@Tecnativa

### DIFF
--- a/account_reconcile_oca/views/account_bank_statement_line.xml
+++ b/account_reconcile_oca/views/account_bank_statement_line.xml
@@ -303,10 +303,10 @@
                                 <label
                                     for="manual_move_id"
                                     string=""
-                                    insible="manual_move_type not in ['in_invoice', 'in_refund', 'out_invoice', 'out_refund'] or manual_original_amount == 0 or is_reconciled"
+                                    invisible="manual_move_type not in ['in_invoice', 'in_refund', 'out_invoice', 'out_refund'] or manual_original_amount == 0 or is_reconciled"
                                 />
                                 <div
-                                    insible="manual_move_type not in ['in_invoice', 'in_refund', 'out_invoice', 'out_refund'] or manual_original_amount == 0 or is_reconciled"
+                                    invisible="manual_move_type not in ['in_invoice', 'in_refund', 'out_invoice', 'out_refund'] or manual_original_amount == 0 or is_reconciled"
                                 >
                                     Invoice <field
                                     class="oe_inline"


pedrobaeza:
Fast-tracking this as clear patch:

/ocabot merge patch

PR #1040, opened 2026-09-08
[19.0][FIX] account_reconcile_oca: use the line's company for partner accounts

dnplkndll:
The partner's receivable/payable is company-dependent and was read in the current user's company instead of the statement line's, so a line of another company got the wrong account and its move failed `_check_company`. Test fails without the fix; same lines exist on 18.0.

### DIFF
--- a/account_reconcile_oca/models/account_bank_statement_line.py
+++ b/account_reconcile_oca/models/account_bank_statement_line.py
@@ -293,12 +293,13 @@ class AccountBankStatementLine(models.Model):
                 )
             else:
                 account = self.journal_id.suspense_account_id
-                if self.partner_id and total_amount > 0:
+                partner = self.partner_id.with_company(self.company_id)
+                if partner and total_amount > 0:
                     can_reconcile = True
-                    account = self.partner_id.property_account_receivable_id
-                elif self.partner_id and total_amount < 0:
+                    account = partner.property_account_receivable_id
+                elif partner and total_amount < 0:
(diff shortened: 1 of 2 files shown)


ryanduguid (commented):
Thanks for the fix and for the test that pins it.

Code review only, no functional run.

The diagnosis matches core. In 19.0 both `property_account_receivable_id` and
`property_account_payable_id` are `company_dependent=True` with `check_company=True`,
so reading them off a bare `self.partner_id` resolves against `self.env.company` and
the resulting account then fails `_check_company` on a line belonging to another
company. `with_company` is the right lever, and it is already the idiom in this
module at `account_reconcile_abstract.py:49` and a few lines below yours at line 625.

I grepped the module for other reads of those two fields and yours are the only ones,
so the fix looks complete rather than one call site of several.

One request. 18.0 has the identical lines at 299 to 304 and I could not find an open
PR for it. Your description already says as much. Could you open the 18.0 one as
well, or say here that you will, so the older series does not keep the bug?

One adjacent observation, not for this PR. When there is no partner the account stays
`self.journal_id.suspense_account_id`, which carries the user's environment, and
`display_name` on `account.account` resolves the code through `code_store` against
`env.company.root_id`. So on a cross-company line with no partner the proposal would
still show the suspense account's code as read in the wrong company. Your change
fixes that incidentally for the partner paths, since the account now comes off a
`with_company` partner. Worth a follow-up if you agree it is real.

PR #1044, opened 2026-09-18
[19.0][WIP] account_reconcile_oca: Add auto_reconcile_invoices configuration

victoralmau:
Add auto_reconcile_invoices configuration

The goal is to be able to configure settings related to the old reconciliation model with `type=invoice_matching`

@Tecnativa TT64407


account_reconcile_oca/models/res_company.py
```diff
@@ -13,6 +13,21 @@ class ResCompany(models.Model):
         ._fields["reconcile_aggregate"]
         .selection
     )
+    # auto_reconcile_invoices fields conditions
+    auto_reconcile_invoices_past_months_limit = fields.Integer(default=18)
```
pedrobaeza: I think the names can be set with only `reconcile_` prefix, not `auto_reconcile` as anyway, it's not totally auto unless you check `auto_reconcile_invoices_auto_reconcile`, and you can see in the field name the redundancy. I also try to avoid plurals as much for avoiding mispellings when invoking the field, but that's not enforced.

account_reconcile_oca/models/res_company.py
```diff
@@ -13,6 +13,21 @@ class ResCompany(models.Model):
...
+    auto_reconcile_invoices_match_partner_ids = fields.Many2many(
+        string="Auto-reconcile Invoices Matching partners", comodel_name="res.partner"
+    )
+    auto_reconcile_invoices_match_partner_category_ids = fields.Many2many(
+        comodel_name="res.partner.category"
+    )
+    auto_reconcile_invoices_match_same_currency = fields.Boolean(default=True)
+    auto_reconcile_invoices_match_text_location_label = fields.Boolean()
+    auto_reconcile_invoices_match_text_location_note = fields.Boolean()
+    auto_reconcile_invoices_match_text_location_reference = fields.Boolean()
```
pedrobaeza: I think this option shouldn't be so unfolded, but maybe one (or a selectiong) saying to match by amount or by text and amount).

account_reconcile_oca/README.rst
```diff
@@ -85,8 +85,8 @@ Authors
 Contributors
 ------------
 
-- Enric Tobella
-- Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+-  Enric Tobella
+-  Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
```
pedrobaeza, line 89: You can add yourself as contributor here.

account_reconcile_oca/models/res_company.py
victoralmau: Okay, but those are different things; that is, searching by amount is something this PR doesn't add (or at least not yet, because e)
ejemplo

The fields you specify are used to search by different fields (you can choose which ones) regardless of the above.
ejemplo2

pedrobaeza: I think we should reduce options to the ones really used. If you search by text, you search on the available fields. I don't think we need such granularity. About amount, it's true that you may want to allow a tolerance, but it's also a rare case. Check if enterprise is allowing such tolerance, and we can match the feature.

victoralmau: I've combined the three fields into one.

As for the amount, I don't see how that's configured in Enterprise. Do you know exactly how to do it so I can check it?

